Issue #4, opened 2018-12-21
How to handle randomness in optimizer


se4u:
Hi, I have prevously worked¹ on a gradient-free optimization algorithm called SPSA²-³, and I have Matlab/mex code⁴ that I can port to python easily. I am interested in benchmarking SPSA against other zero-order optimization algorithms using nevergrad. 

I am following the instructions for benchmarking a new optimizer given in [adding-your-own-experiments-andor-optimizers-andor-function](https://github.com/facebookresearch/nevergrad#adding-your-own-experiments-andor-optimizers-andor-functions). My understanding is that I can just add a new `SPSA` class in `nevergrad/optimization/optimizerlib.py` and implement 

* __init
* _internal_ask
* _internal_provide_recommendation
* _internal_tell

functions and then add `SPSA` to the `optims` variable in the right experiment function in the `nevergrad/benchmark/experiments.py` module and then I should be able to generate graphs like 
[docs/resources/noise_r400s12_xpresults_namecigar,rotationTrue.png](https://github.com/facebookresearch/nevergrad/blob/master/docs/resources/noise_r400s12_xpresults_namecigar,rotationTrue.png)

However, SPSA itself uses an rng in the `_internal_ask` function. But the `optimizer` `base` class does not take any seed in the `__init__` function. What will be a good way to make the experiments reproducible in such situation? 

[1] Pushpendre Rastogi, Jingyi Zhu, James C. Spall (2016). Efficient implementation of Enhanced Adaptive Simultaneous Perturbation Algorithms. CISS 2016, pdf
[2] https://en.wikipedia.org/wiki/Simultaneous_perturbation_stochastic_approximation
[3] https://www.chessprogramming.org/SPSA
[4] https://github.com/se4u/FASPSA/blob/master/src/SPSA.m

tomjorquera:
@se4u not one of the authors here, but it seems the lib is relying on `random` and `numpy.random` for these matter. In both cases, you can set the seed globally by doing `random.seed(1234) or `np.random.seed(1234)` to make subsequent calls to rngs reproducibles.

You can see how the authors use this here https://github.com/facebookresearch/nevergrad/blob/7fb29affd443330d8b75936856f43182a2d3b233/nevergrad/benchmark/xpbase.py#L119-L124

So I guess if you are able to use base python or numpy random, you should be able to get reproducible experiments the same way.

jrapin:
Hi, we are happy to welcome new algorithms indeed!

[...]

PR #16, opened 2019-01-02
Adding SPSA

se4u:
## Motivation and Context

Adding SPSA algorithm.

https://github.com/facebookresearch/nevergrad/issues/4

## How Has This Been Tested

Currently the test `nosetests nevergrad/optimization/test_optimizerlib.py` fails. Hence the WIP tag.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

Changed `nevergrad/benchmark/experiments.py` and `nevergrad/optimization/optimizerlib.py`

## Checklist

- [ x] My code follows the code style of this project.
- [ x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.


jrapin:
I figured out the issue. Actually when you seed the random state, then it works. You still had an error because when you tried, you had already created a history for your algorithm from your previous attempts, which differed from the output you got with the seed. After deleting the file and rerunning, it worked.

But since we don't want to seed the random state, it does not solve the problem. I expected the RandomState to get its default seed from the default random generator, but it was a wrong assumption: f you create it without seed, we cannot seed from outside. I have modified the guidelines to state that one should initialize it this way then:
```python
self._rng = np.ranndom.RandomState(np.random.randint(2**32))
 ```
After this, it works!

Concerning the other error, I have updated the test to avoid using multiple workers on non-parallel optimizers. You're optimizer seems to be a bit slow to converge though. If it's normal (that's often what happens for optimizers dealing with noise), you can add it to the list of "SLOW" optimizers, in the test file:
https://github.com/facebookresearch/nevergrad/blob/0c29a99da53c484731dfb512df13744b39794656/nevergrad/optimization/test_optimizerlib.py#L47

se4u:
The tests now pass and I added a benchmark `spsa_benchmark` which gives results like the following:


Fig 1a. rotation_False | Fig 1b. rotation_true
:-------------------------:|:-------------------------:
fight_rotationfalse |  fight_rotationtrue

Fig 2a. Sphere | Fig 2b. Sphere4 | Fig 2c. Cigar
:--------------:|:-----------:|:-----------:
fight_namesphere |  fight_namesphere4 |  fight_namecigar

There are still [two comments about `todo` items](https://github.com/facebookresearch/nevergrad/pull/16/commits/c8ea286592b23aabb2a7b4cce93613cac4299ba5#diff-392c981c3de5c2d86ed5864cf03abbbaR469)  related to estimating the noise standard deviation and the gradient magnitude of the function in the beginning of the optimization, but they don't really change the overall picture shown in the benchmarks.

se4u:
Here are the plots 

1. cigar rotationfalse
2. cigar rotationtrue
3. sphere rotationfalse
4. sphere rotationtrue
5. sphere4 rotationfalse
6. sphere4 rotationtrue


xpresults_namecigar rotationfalse
xpresults_namecigar rotationtrue
xpresults_namesphere rotationfalse
xpresults_namesphere rotationtrue
xpresults_namesphere4 rotationfalse
xpresults_namesphere4 rotationtrue

se4u:
@teytaud Totally agree to your points.

@jrapin I think this PR is ready to merge now. I left the TODOs in, with the two hardcoded parameters.

jrapin:
Let's go then ;) Thank you!

PR #35, opened 2019-01-13
New `unregister` function

claforte:
## Motivation and Context

The current decorators allow registering a function/class but don't explicitly provide a way to deregister/unregister that function/class. This is fine when running the code once, but makes developing in Jupyter, where re-running cells is common, harder than it should be.

## How Has This Been Tested

I ran `nosetests nevergrad` and while there were errors, they appeared to be unrelated to this new feature and occurred without it.

I have also used this small feature with my own code without noticeable problem.

## Types of changes

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

### DIFF
--- a/nevergrad/common/decorators.py
+++ b/nevergrad/common/decorators.py
@@ -31,6 +31,15 @@ class Registry(dict):
             self._information[name] = info
         return obj
 
+    def unregister(self, name:str) -> None:
+        """Remove a previously-registered function or class, e.g. so you can 
+        re-register it in a Jupyter notebook.
+        """
+        if name in self:
+            obj = self[name]
+            if obj in self._registered: self._registered.remove(obj)
+            del self[name]
+
     def register_with_info(self, **info: Any) -> Callable:
         """Decorator for registering a function and information about it
         """


claforte:
I just noticed that while unregister and deregister are both acceptable verbs, most of the time, deregister is used instead of unregister. Feel free to correct it accordingly or I can do it myself in a later commit next time I fix typos. ;-)

jrapin:
I didn't know that "deregister" existed. I am fine with any one of them if both make sense (I am not the best judge for that).
Two comments:
- I prefer avoiding one liner "if" statements (pep8 says it is "discouraged")
- could you add a simple test in nevergrad/common/decorators.py? Just something simple to cover this bit of code. I consider it always helps to avoid breaking it later on. If you are not familiar with tests, I will do it myself though.

claforte:
Hi Jérémy, can you merge the code and apply the fixes you suggested. Next time I promise to run pep8. I will also look at how you write this new test to follow your conventions in future commits. I hope you don't mind.

jrapin:
Will do ;)

jrapin:
Was simpler to merge and create a new pull request for the tests. Here it is #41.
(I used this to remove a legacy attribute as well)

claforte:
Thanks Jérémy!

Christian

>

Issue #62, opened 2019-01-23
Uncommenting fails for custom file extensions


Dominik1123:
## Steps to reproduce

1. Create script with custom extensions (one that is not part of [the defaults](https://github.com/facebookresearch/nevergrad/blob/8d6381cbb750ce22e8f6ca3f288a2f37ebe4c1b8/nevergrad/instrumentation/folderfunction.py#L57)).
2. Place inside some line which should be interpolated, e.g. `// @nevergrad@ var = NG_G{0, 1};`.
3. Create a corresponding `FolderFunction`.
4. (By the way, it looks like the extension `.m` is missing the `.` [here](https://github.com/facebookresearch/nevergrad/blob/8d6381cbb750ce22e8f6ca3f288a2f37ebe4c1b8/nevergrad/instrumentation/folderfunction.py#L57).)

## Observed Results

The following exception is raised ([location in code](https://github.com/facebookresearch/nevergrad/blob/8d6381cbb750ce22e8f6ca3f288a2f37ebe4c1b8/nevergrad/instrumentation/instantiate.py#L41)):

```
Traceback (most recent call last):
  File "test.py", line 15, in <module>
    func = FolderFunction(folder, command, extensions=['.custom'])
  File ".../lib/python3.6/site-packages/nevergrad/instrumentation/folderfunction.py", line 61, in __init__
    self.instrumentized_folder = InstrumentizedFolder(folder, extensions=extensions, clean_copy=clean_copy)
  File ".../lib/python3.6/site-packages/nevergrad/instrumentation/instantiate.py", line 127, in __init__
    instru_f = InstrumentizedFile(fp)
  File ".../lib/python3.6/site-packages/nevergrad/instrumentation/instantiate.py", line 63, in __init__
    lines = [(l if LINETOKEN not in l else uncomment_line(l, ext)) for l in lines]
  File ".../lib/python3.6/site-packages/nevergrad/instrumentation/instantiate.py", line 63, in <listcomp>
    lines = [(l if LINETOKEN not in l else uncomment_line(l, ext)) for l in lines]
  File ".../lib/python3.6/site-packages/nevergrad/instrumentation/instantiate.py", line 42, in uncomment_line
    pattern += r'(?P<linetoken>' + comment_chars[extension] + r" *" + LINETOKEN + r" *)"
KeyError: '.custom'
```

## Expected Results

[...]

PR #65, opened 2019-01-24
62-custom-file-types

Dominik1123:
## Types of changes

Using custom file types for token instrumentation raised an exception as detailed in #62 . Using custom file types (extensions) was therefore not possible.

```
nosetests nevergrad
nosetests nevergrad.instrumentation.test_instantiate
```

## Checklist


jrapin:
Looks good to me, thank you!
2 remarks:
- I may add back the `extensions` parameters of `FolderFunction` eventually. Its goal was to help focus on one type of file and avoid parsing a full project. May have been over-optimization though, so I'll see if the need arises or not.
- any specific reason for using a `dict` with only a "comment" key: `".c": dict(comment="//")`? This seems unnecessary to me. I may remove it eventually if you do not see any future use.

Dominik1123:
No, I didn't have a strong reason for using another `dict`, but having a variable `FILE_TYPES` it seemed clearer to mark the comment chars explicitly as such by adding the "comment" key dict. But could be renamed to `COMMENT_CHARS` for example and then leave out the innermost `dict`.

Regarding the extensions parameter for filtering project files, maybe it would be more convenient to let the user specify a regex or wildcard pattern and only consider matching files? That way filtering by extensions would be included as a special case (e.g. `"*.c"`) while allowing for more complex filters as well.

jrapin:
`COMMENT_CHARS` seems good to me. Either you want to change this yourself, or I merge now and I will do it eventually (but I am in no hurry, it does not change the interface anyway).

regex seems a good idea indeed! Let's just remove the `extensions` parameter for now as you did, and when we need filtering we'll add regex ;)

Dominik1123:
Done :-)

jrapin:
Thanks a lot!

Issue #101, opened 2019-02-13
Make optimizers use less memory


jrapin:
Optimizer's archive is a dict of type `Dict[Tuple[float,...], Value]` where `Value` is a light-weighted class. It keeps track of all function evaluations.
This is quite resourceful in term of memory for several reasons:
 1. it's a dict.
 2. the keys are tuples of float of the function dimension size (which could be millions).
 3. it keeps track of all the estimations.

While I cannot think of a way to get rid of the dict, we can deal with the other 2 issues:
 2. use some kind of hashable array instead of tuple of floats. This can lead to a factor 4 improvement at least. See some examples here https://stackoverflow.com/questions/16589791/most-efficient-property-to-hash-for-numpy-array
 3. implement pruning on the archive, getting rid of the lowest values for instance.

PR #128, opened 2019-02-27
 Prepare a dict-like archive structure for arrays

jrapin:
## Types of changes

Preliminary work for #101 
This creates a dict-like object which can fakes arrays as keys, and is mostly compatible with the current code.

## Checklist

### DIFF
--- a/nevergrad/optimization/differentialevolution.py
+++ b/nevergrad/optimization/differentialevolution.py
@@ -25,7 +25,6 @@ class _DE(base.Optimizer):
     def __init__(self, dimension: int, budget: Optional[int] = None, num_workers: int = 1) -> None:
         super().__init__(dimension, budget=budget, num_workers=num_workers)
         self._parameters = DifferentialEvolution()
-        self.archive = base.utils.Archive()
         self._llambda: Optional[int] = None
         self.population: List[Optional[ArrayLike]] = []
         self.candidates: List[Optional[ArrayLike]] = []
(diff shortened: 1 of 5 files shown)


jrapin:
Manually testing with the snippet below:
- Archive: 3.6s and 0.8GB
- dict: 10.3s and 2.9GB

Conversion to tuple was slow, so `tobytes` does not seem an issue in term of performance and seems to provide a big win. Memory usage is much better as well. Assuming that access should be roughly equivalent as well.

```python
import time
import numpy as np
from nevergrad.optimization.utils import Archive

archive = Archive()
t0 = time.time()
for k in range(1000):
    x = np.random.normal(size=100000)
    archive[x] = 12
print("duration", time.time() - t0)

archive = dict()
t0 = time.time()
for k in range(1000):
    x = np.random.normal(size=100000)
    archive[tuple(x)] = 12
print("duration", time.time() - t0)
```

nevergrad/optimization/differentialevolution.py
```diff
@@ -25,6 +25,7 @@ class _DE(base.Optimizer):
     def __init__(self, dimension: int, budget: Optional[int] = None, num_workers: int = 1) -> None:
         super().__init__(dimension, budget=budget, num_workers=num_workers)
         self._parameters = DifferentialEvolution()
+        self.archive = base.utils.Archive()
```
jrapin: This plugs the new `Archive` into DE algorithms

jrapin: Unplugged now.

jrapin:
Following up to plug this into all algorithms.

PR #158, opened 2019-03-20
Make a generic type out of Registry

jrapin:
## Types of changes

Improves typing since so far registry were returning Any.

## Checklist

### DIFF
--- a/nevergrad/benchmark/experiments.py
+++ b/nevergrad/benchmark/experiments.py
@@ -257,11 +257,9 @@ def mldaas(seed: Optional[int] = None) -> Iterator[Experiment]:
                             yield xp
 
 
-# def arcoating(seed: Optional[int] = None) -> Iterator[Experiment]:
 @registry.register
-def arcoating() -> Iterator[Experiment]:
+def arcoating(seed: Optional[int] = None) -> Iterator[Experiment]:
     func = ARCoating()
-    seed = 12
     seedg = create_seed_generator(seed)
     algos = ["NaiveTBPSA", "Cobyla", "SQP", "Powell", "LargeScrHammersleySearch", "ScrHammersleySearch",
              "PSO", "OnePlusOne", "CMA", "TwoPointsDE", "QrDE", "LhsDE", "Zero", "StupidRandom"]
--- a/nevergrad/benchmark/experiments.py
+++ b/nevergrad/benchmark/experiments.py
@@ -257,9 +257,11 @@ def mldaas(seed: Optional[int] = None) -> Iterator[Experiment]:
                             yield xp
 
 
+# def arcoating(seed: Optional[int] = None) -> Iterator[Experiment]:
 @registry.register
-def arcoating(seed: Optional[int] = None) -> Iterator[Experiment]:
+def arcoating() -> Iterator[Experiment]:
     func = ARCoating()
+    seed = 12
     seedg = create_seed_generator(seed)
     algos = ["NaiveTBPSA", "Cobyla", "SQP", "Powell", "LargeScrHammersleySearch", "ScrHammersleySearch",
              "PSO", "OnePlusOne", "CMA", "TwoPointsDE", "QrDE", "LhsDE", "Zero", "StupidRandom"]
--- a/nevergrad/common/decorators.py
+++ b/nevergrad/common/decorators.py
@@ -3,11 +3,14 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Callable, Optional, Dict
+from typing import Any, Callable, Optional, Dict, TypeVar
 import functools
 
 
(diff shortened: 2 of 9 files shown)


nevergrad/common/decorators.py
```diff
@@ -3,28 +3,36 @@
...
 
-class Registry(dict):  # type: ignore
+X = TypeVar("X")
+
+
+# pylint does not understand Dict, and mypy needs Dict to be generic...
+# the following is (very) hacky but solves both issues for now
+# It can be safely reverted to Registry(dict) though...
+# pylint: disable=too-many-ancestors
+class Registry(Dict[str, X], UserDict):  # type: ignore
```
lerks: I don't think this does what you want. Dict means this is a *real* dict, whereas UserDict means that this has a dict as the `data` attribute. They define exactly the same methods, so the one that comes in the MRO overrides the second one. What this means here is that UserDict has no effect at all. If you want to keep UserDict make the first a MutableMapping.

nevergrad/common/decorators.py
```diff
@@ -3,28 +3,36 @@
...
         self._information: Dict[str, Dict[Any, Any]] = {}
 
-    def register(self, obj: Any, info: Optional[Dict[Any, Any]] = None) -> Any:
+    def register(self, obj: X, info: Optional[Dict[Any, Any]] = None) -> Any:
         """Decorator method for registering functions/classes
         The info variable can be filled up using the register_with_info
         decorator instead of this one.
         """
-        name = obj.__name__ if hasattr(obj, "__name__") else obj.__class__.__name__
+        name = obj.__name__ if hasattr(obj, "__name__") else obj.__class__.__name__  # type: ignore
```
lerks: This is `getattr(obj, "__name__", obj.__class__.__name__)`.

nevergrad/common/decorators.py
```diff
@@ -3,28 +3,36 @@
...
+class Registry(Dict[str, X], UserDict):  # type: ignore
     """Registers function or classes as a dict.
     """
 
     def __init__(self) -> None:
         super().__init__()
         self._information: Dict[str, Dict[Any, Any]] = {}
 
-    def register(self, obj: Any, info: Optional[Dict[Any, Any]] = None) -> Any:
+    def register(self, obj: X, info: Optional[Dict[Any, Any]] = None) -> Any:
```
lerks: `register` is now also *returning* an object of type `X`

nevergrad/benchmark/experiments.py
```diff
@@ -257,9 +257,11 @@ def mldaas(seed: Optional[int] = None) -> Iterator[Experiment]:
                             yield xp
 
 
+# def arcoating(seed: Optional[int] = None) -> Iterator[Experiment]:
 @registry.register
-def arcoating(seed: Optional[int] = None) -> Iterator[Experiment]:
+def arcoating() -> Iterator[Experiment]:
     func = ARCoating()
+    seed = 12
```
lerks: Why do you need this? Either this isn't used, or you're introducing a seemingly unrelated functional change.

nevergrad/common/decorators.py
jrapin: Good point

nevergrad/common/decorators.py
jrapin: Oh did not know there was a default!

nevergrad/common/decorators.py
jrapin: It actually *does* what I want but it is really hacky:
- the methods which are used are the ones in `Dict[str, X]`, which is convenient since they are typed.
- `UserDict` is basically only there so that pylint does not complain since it does not recognize Dict. The `data` field is unused... actually it does not even exist, not sure why :s
- using `Registry(UserDict, MutableMapping[str, X])` has untyped getitem... I would have to rewrite all dict methods :s

nevergrad/benchmark/experiments.py
jrapin: This was just a test that I removed probably a second after your comment

jrapin:
I can't make it work in a non-hacky way... is it worth it ? :s

nevergrad/common/decorators.py
lerks: You can have this extend just MutableMapping, remove all dict or UserDict superclass, then manually define a `data` attribute, which is a dict, and manually implement the 5 abstract methods of MutableMapping (see there [here](https://docs.python.org/3/library/collections.abc.html)). These methods will basically just be
```py
def foo(*args, *kwargs):
    return self.data.foo(*args, **kwargs)
```
Once you do so you get for free all other methods of the dictionary interface. So, in short, your class will be a regular class (not a dict) but will be dict-like.

jrapin: I was a bit afraid pylint would not recognize it either, but it worked perfectly! thank you!

Issue #179, opened 2019-04-13
RuntimeError("Only one call can be registered at a time") with RBO


robert1826:
## Steps to reproduce

  1. install nevergrad-0.2.0
  2. initialize a RBO optimizer with small budget (I used 10 in this case)
  3. run optimization
  
- note that this happens randomly so sometimes I run and get no errors

## Observed Results

```
Traceback (most recent call last):
  File ".\mnist.py", line 51, in <module>
    recommendation = optimizer.optimize(ifunc)
  File "C:\Users\robert\AppData\Local\Programs\Python\Python36\lib\site-packages\nevergrad\optimization\base.py", line 402, in optimize
    self.tell(x, job.result())
  File "C:\Users\robert\AppData\Local\Programs\Python\Python36\lib\site-packages\nevergrad\optimization\base.py", line 244, in tell
    self._internal_tell_candidate(candidate, value)
  File "C:\Users\robert\AppData\Local\Programs\Python\Python36\lib\site-packages\nevergrad\optimization\base.py", line 328, in _internal_tell_candidate
    self._internal_tell(candidate.data, value)
  File "C:\Users\robert\AppData\Local\Programs\Python\Python36\lib\site-packages\nevergrad\optimization\optimizerlib.py", line 1025, in _internal_tell
    self._fake_function.register(y, -value)  # minimizing
  File "C:\Users\robert\AppData\Local\Programs\Python\Python36\lib\site-packages\nevergrad\optimization\optimizerlib.py", line 967, in register
    raise RuntimeError("Only one call can be registered at a time")
```

## Expected Results

  There should be no errors running the optimization

## Relevant Code

  ```
from sklearn.datasets import load_digits
from sklearn.model_selection import train_test_split
from sklearn.svm import SVC

from nevergrad.optimization import optimizerlib
from nevergrad import instrumentation as instru
from nevergrad.optimization import base

X, y = load_digits(return_X_y=True)
Xtrain, XTest, yTrain, yTest = train_test_split(X, y, random_state=42)

print('training with {} examples, testing with {}'.format(len(Xtrain), len(XTest)))

def train_and_score(args):
    if args['C_in'] > 0:
        model = SVC(C=args['C_in'], kernel=args['kernel_in'], degree=args['degree_in'])
        model.fit(Xtrain, yTrain)
        res = model.score(XTest, yTest)
    else:
        res = 1000
    
    if res > 0.98 and res < 1:
        print('args: {}'.format(args), end=', ')
[...]

PR #185, opened 2019-04-19
Stop checking calls to fake function in BO

jrapin:
## Types of changes

#179 
For some unknown reason, BO package seems to ask twice for a point, but kept a memory of the value so does not call the "fake" function, which triggered an error. This test is therefore removed.

## Checklist

### DIFF
--- a/nevergrad/optimization/optimizerlib.py
+++ b/nevergrad/optimization/optimizerlib.py
@@ -1024,8 +1024,12 @@ class _BO(base.Optimizer):
         return candidate
 
     def _internal_tell_candidate(self, candidate: base.Candidate, value: float) -> None:
-        y = np.array(candidate._meta["x_probe"], copy=False)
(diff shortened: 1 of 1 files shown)


jrapin:
@teytaud does it make sense to use a `SoftmaxCategorical` variable with `BO`?
- if `deterministic=False` then the function is stochastic, which does not seem to be handled by `BO`.
- if `deterministic=True` then the function is discontinuous, which is probably not a good setting for `BO` right?
I am therefore wondering if I should ask a warning against this.

teytaud:
Softmax categorical could make sense for Bayesian Optimization in the noisy case, which does not seem to be handled by this version of Bayesian Optimization, so yes here this looks like a suboptimal choice.

Softmax categorical with deterministic=True is more reasonable, though you're right that discontinuity  might be complicated for BO - this is less a concern than the previous case, as far as I can tell.

PR #186, opened 2019-04-20
Fix optimizer when having started more jobs than workers

louismartin:
Handle the case when the optimizer started more jobs than num_workers, e.g. if you lower the value of num_worker during the optimization.

### DIFF
--- a/nevergrad/optimization/base.py
+++ b/nevergrad/optimization/base.py
@@ -411,7 +411,7 @@ class Optimizer:  # pylint: disable=too-many-instance-attributes
                 sleeper.sleep()
             # # # # # Start new jobs # # # # #
             if not batch_mode or not self._running_jobs:
-                new_sugg = min(0, remaining_budget, self.num_workers - len(self._running_jobs))
+                new_sugg = max(0, min(remaining_budget, self.num_workers - len(self._running_jobs)))
                 if verbosity and new_sugg:
                     print(f"Launching {new_sugg} jobs with new suggestions")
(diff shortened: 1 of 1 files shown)


jrapin:
What kind of error did it raise?

nevergrad/optimization/base.py
```diff
@@ -411,7 +411,7 @@ def optimize(self, objective_function: Callable[..., float],
                 sleeper.sleep()
             # # # # # Start new jobs # # # # #
             if not batch_mode or not self._running_jobs:
-                new_sugg = min(remaining_budget, self.num_workers - len(self._running_jobs))
```
jrapin: You could also use `np.clip`, but in this case you may need to cast to int afterwards, dunno if it is worth it

louismartin:
It did not actually raise any error but it kept saying something like "Launching -16 jobs with new suggestions" without starting any jobs (which it indeed shouldn't).

nevergrad/optimization/base.py
louismartin: Yes true but I think that:
`np.clip(min(remaining_budget, self.num_workers - len(self._running_jobs)), a_min=0, a_max=None)`
if not much clearer to justify using numpy instead of the builtins

jrapin: `np.clip(self.num_workers - len(self._running_jobs)), a_min=0, a_max=remaining_budget)` seems simpler though :p

PR #192, opened 2019-05-03
 Add a rounded method for Array

jrapin:
## Types of changes

Add a rounding capacity for `Array`

## Checklist


jrapin:
@louismartin is that what you wanted? A problem here is that is that the int is still cast to float if you use `asfloat`. So I am considering deleting this diff, and just adding a `asint` that rounds the output and cast to int. But then we would not be able to create an array of ints (dunno if an array of ints can be useful though).

PR #194, opened 2019-05-03
Rename asfloat to asscalar and allow int casting

jrapin:
## Types of changes

Allow instrumentation with ints

## Checklist

### DIFF
--- a/nevergrad/instrumentation/variables.py
+++ b/nevergrad/instrumentation/variables.py
@@ -2,7 +2,8 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import List, Optional, TypeVar, Union, Sequence, Any
+from typing import List, Optional, TypeVar, Union, Sequence, Any, Type
+import warnings
 import numpy as np
 from . import discretization
 from ..common.typetools import ArrayLike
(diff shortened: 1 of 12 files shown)


nevergrad/instrumentation/variables.py
```diff
@@ -186,13 +187,13 @@ def data_to_argument(self, data: ArrayLike, deterministic: bool = False) -> Y:
         array = np.array(data, copy=False)
         for transf in self.transforms:
             array = transf.forward(array)
-        if self._asfloat:
-            return float(array[0])
+        if self._scalar_type is not None:
+            return self._scalar_type(array[0] if self._scalar_type != int else round(array[0]))
```
louismartin: Wouldn't it better to *cast* it to an int `int(array[0])` as python and numpy do instead of rounding?

e.g. for numpy
```
> np.array([1.9]).astype(int)
array([1])
```

jrapin: I would expect `Array(1).affined(a=1, b=12).asscalar(int)` to be sampled with N(12, 1) and therefore have mean 12, but if we cast to int, it would have mean 11.5 or something like this. Either way, there is a problem I think, and rounding seems the more legitimate... what do you think?

louismartin: Hum that's a very good point, your call on this one, you have more insight

jrapin: I'll go for rounding then, since it does not give any special importance to 0 compared to other values while casting to int does

PR #204, opened 2019-05-09
Prevent bayesian optimization in discontinuous/noisy cases

jrapin:
## Types of changes

Instrumentation can be dis-continuous and noisy, which is not handled by the default parametrization of `BO`. This PR creates an error to make sure they are not used.

## Checklist


jrapin:
@robert1826 this is to prevent anyone using `BO` with noisy or discontinuous instrumentation without setting the `alpha` parameter, does that make sense to you? Or would you just use a warning?

jrapin:
I end up leaning in favor of a warning: the platform is used for research as well, one could want to compare different settings even if they are inefficient.

robert1826:
> I end up leaning in favor of a warning: the platform is used for research as well, one could want to compare different settings even if they are inefficient.

Actually I'd said as well ... for the exact same reason you've mentioned

PR #219, opened 2019-05-21
 Add a mechanism to suggest a given candidate

jrapin:
## Types of changes

This is a new flexible way to deal with telling a point which was not asked for.

## Checklist


nevergrad/optimization/base.py
```diff
@@ -246,6 +247,23 @@ def remove_all_callbacks(self) -> None:
         """
         self._callbacks = {}
 
+    def request(self, *args: Any, **kwargs: Any) -> None:
```
jrapin: Potential other names: `suggest` and `propose`.
`request` seems better to me because it highlights the fact that this will be actually *enforced* at the next `ask`.

nevergrad/optimization/base.py
```diff
@@ -246,6 +247,23 @@ def remove_all_callbacks(self) -> None:
...
+        """Requests a new point to ask.
+        It will be ask at the next call (last in first out).
```
jrapin: I think lifo is better because some algorithms may (will) use this feature for initialization. With lifo you make sure your point is asked first.

nevergrad/optimization/base.py
louismartin: Hum the problem is that request is less intuitive I think.
I would prefer suggest, you can still write in the doc that the optimizer WILL use the candidate

louismartin: Hum the problem is that request is less intuitive I think.
I would prefer suggest, you can still write in the doc that the optimizer WILL use the candidate

nevergrad/optimization/base.py
louismartin: Hum that's a good point. 
Although I think that intuitively I will think that its the other way around when using it "manually".
Could we add a kwarg like `prepend=False` that optimizers will change to True internally when calling the method if they need it? (e.g. to emulate a lifo only for those who need it but fifo by default)

nevergrad/optimization/base.py
jrapin: `suggest` used to be the old name of `ask`... so I am somehow uncomfortable with it :s

nevergrad/optimization/base.py
jrapin: problem with using a kwargs, is that it could interfere with the instrumentation kwargs, so not that nice (though it is not very likely). Do you think it is really important?

jrapin: Discussion summary: keep LIFO to be able to `request` and `ask` straight away

nevergrad/optimization/base.py
jrapin: Discussion summary: `suggest` is more intuitive

PR #230, opened 2019-05-29
Add Registry Import to Readme.md

kmader:
add line to import registry from benchmark to show the list of optimizers

## Types of changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ See the [instrumentation tutorial](docs/instrumentation.md) for more complex ins
 
 You can print the full list of optimizers with:
 ```python
-import nevergrad as ng
+from nevergrad.benchmark import registry
 print(sorted(registry.keys()))
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,7 +81,7 @@ See the [instrumentation tutorial](docs/instrumentation.md) for more complex ins
 
 You can print the full list of optimizers with:
 ```python
-import nevergrad as ng
+from nevergrad.benchmark import registry
```
jrapin: Thank you for spotting, I should have seen this :s
My aim with to access it through `ng` though, since now most things are defined there :
```python
import nevergrad as ng
print(sorted(ng.optimizers.registry.keys()))
```
Would that be fine for you ? It's more straightforward than going through `benchmark` (also I expect that most users do not install the full list of dependencies and may not be able to import `benchmark` anyway)

kmader: That looks good, I just wanted the readme to be runnable.

jrapin: It should have been indeed ;) thanks!
(btw, seems to me that `sorted` returns a `list` so it is probably useless to cast afterwards, but it works fine anyway so I'm merging)

kmader: some of those functions return iterables in python 3 and so I just had list to keep it clear, but @jrapin is right

PR #243, opened 2019-06-11
qo

teytaud:
## Types of changes

Very cool paper http://rahnamayan.ca/assets/documents/Center-Based%20Sampling%20for%20Population-Based%20Algorithms.pdf  about a sampling related to our "plus middle point".

Just CI.

## Checklist

### DIFF
--- a/nevergrad/optimization/oneshot.py
+++ b/nevergrad/optimization/oneshot.py
@@ -29,6 +29,35 @@ class _RandomSearch(OneShotOptimizer):
 
     def _internal_ask(self) -> ArrayLike:
         # pylint: disable=not-callable
+        if self._parameters.quasi_opposite == "quasi" and (self._num_ask % 2):
+            return -self.random_state.uniform(0., 1.) * self.last_guy
+        if self._parameters.quasi_opposite == "opposite" and (self._num_ask % 2):
+            return -self.last_guy
+        if self._parameters.middle_point and not self._num_ask:
+            self.last_guy = np.zeros(self.dimension)
+            return self.last_guy  # type: ignore
+        scale = self._parameters.scale
+        if isinstance(scale, str) and scale == "random":
+            scale = np.exp(self.random_state.normal(0., 1.) - 2.) / np.sqrt(self.dimension)
+        point = (self.random_state.standard_cauchy(self.dimension) if self._parameters.cauchy
+                 else self.random_state.normal(0, 1, self.dimension))
+        last_guy = scale * point
+        return last_guy  # type: ignore
+
+    def _internal_provide_recommendation(self) -> ArrayLike:
+        if self._parameters.stupid:
+            return self._internal_ask()
+        return super()._internal_provide_recommendation()
+
+
+class _RandomSearch(OneShotOptimizer):
+
+    def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optional[int] = None, num_workers: int = 1) -> None:
+        super().__init__(instrumentation, budget=budget, num_workers=num_workers)
+        self._parameters = RandomSearchMaker()  # updated by the parametrized family
+
+    def _internal_ask(self) -> ArrayLike:
+        # pylint: disable=not-callable
         if self._parameters.middle_point and not self._num_ask:
             return np.zeros(self.dimension)  # type: ignore
         scale = self._parameters.scale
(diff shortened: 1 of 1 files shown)


teytaud:
I failed to add fteytaud as a reviewer (winner of the related challenge at CEC).
Not sure, maybe only people who add a star or something can be added as a reviewer.

teytaud:
@jrapin : thinking carefully about it, I have a doubt on my own PR --- this looks correct to me, but maybe it would be better to do such changes in sequences.py, so that it can also be used in the initialization of non-one-shot optimization algorithms. On the other hand, all one-shot algorithms could be used as initializers for e.g. BO or DE or PSO; so maybe the proper solution is to have the possibility to use a oneshot optimizer as a sampling tool for the first population in DE / PSO / BO / etc. Your opinion ?

nevergrad/optimization/oneshot.py
```diff
@@ -29,14 +29,20 @@ def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optiona
 
     def _internal_ask(self) -> ArrayLike:
         # pylint: disable=not-callable
+        if self._parameters.quasi_opposite == "quasi" and self._num_ask % 2:
```
jrapin: I'll tend to restructure this:
```python
if self._num_ask % 2:
   if param.quasi_opposite == ...:
        ...
   elif ...:
        ...
```
I tend to find it slightly simpler to read.

Also, if I am correct, with `middle_point == True`, this can lead to a second middle point being asked, which is not optimal.

nevergrad/optimization/oneshot.py
```diff
@@ -53,6 +59,10 @@ class RandomSearchMaker(base.ParametrizedFamily):
         Provides a random recommendation instead of the best point so far (for baseline)
     middle_point: bool
         enforces that the first suggested point (ask) is zero.
+    quasi_opposite: str
```
jrapin: not sure about the mode. Maybe `opposition_mode` ? And it would be better to have it `opposition_mode: Optional[str] = None` (instead of default being the "none" string). This makes it clearer that it is optional

nevergrad/optimization/oneshot.py
```diff
@@ -65,10 +75,12 @@ class RandomSearchMaker(base.ParametrizedFamily):
 
     # pylint: disable=unused-argument
     def __init__(self, *, middle_point: bool = False, stupid: bool = False,
+                 quasi_opposite: str = "none",
                  cauchy: bool = False, scale: Union[float, str] = 1.) -> None:
         # keep all parameters and set initialize superclass for print
         assert isinstance(scale, (int, float)) or scale == "random"
         self.middle_point = middle_point
+        self.quasi_opposite = quasi_opposite
```
jrapin: add a check that it is `in [None, "quasi", "opposite"]`

jrapin:
In my opinion, keep the PR as it is (with the few comments I made), and we'll update the initializations in another PR

jrapin:
I've updated it to include recent changes and to solve the problems I mentioned.
If I am correct, you wan to do the same for the other one shot class `SamplingSearch`. This will make redundant code. I think it is the simpler for now anyway, but we should probably merge `RandomSearch` and `SamplingSearch` at some point.

nevergrad/optimization/oneshot.py
```diff
@@ -26,17 +26,25 @@ class _RandomSearch(OneShotOptimizer):
     def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optional[int] = None, num_workers: int = 1) -> None:
         super().__init__(instrumentation, budget=budget, num_workers=num_workers)
         self._parameters = RandomSearchMaker()  # updated by the parametrized family
+        self._last_ask: np.ndarray = np.zeros((self.dimension,))
```
teytaud: thanks for your implementation @jrapin, cleaner than mine :-)

nevergrad/optimization/oneshot.py
```diff
@@ -26,17 +26,25 @@ class _RandomSearch(OneShotOptimizer):
...
-        if self._parameters.middle_point and not self._num_ask:
-            return np.zeros(self.dimension)  # type: ignore
-        scale = self._parameters.scale
+        params = self._parameters
+        if params.opposition_mode is not None and self.num_ask % 2 and self.num_ask > params.middle_point:
+            if params.opposition_mode == "quasi":
+                return -self._rng.uniform(0., 1.) * self._last_ask  # type: ignore
+            elif params.opposition_mode == "opposite":
+                return -self._last_ask
+        if params.middle_point and not self._num_ask:
```
teytaud: thanks for the fix, your "not self._num_ask" is a significant improvement.

teytaud:
Maybe we merge that one, and when we find time for it we will discuss the best way of adding these things into the initialization of DE / PSO / BO ?

PR #245, opened 2019-06-17
Bug: Updating word "optimizer" to "optimization"

vishalshar:
typing mistake optimizer to optimization

## Types of changes

It fixes the problem of not able to run the optimizer

I tested it with available algorithms 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ import nevergrad as ng
 def square(x):
     return sum((x - .5)**2)
 
-optimizer = ng.optimizers.OnePlusOne(instrumentation=2, budget=100)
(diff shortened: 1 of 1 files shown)


jrapin:
Actually, `optimizer = ng.optimizers.OnePlusOne(instrumentation=2, budget=100)` does work, but only on master. It's a pretty recent change, I'll try to push a new version soon, so that the doc is not out of sync too long.

Also, it is preferable to use `OnePlusOne` instead of `_OnePlusOne`. They will act the same way, but one is an instance and the other a class (this is a work in progress for having parametrizable optimizers).

PR #255, opened 2019-07-15
added PBIL and cGA

LaurentMnr95:
## Types of changes

Added cGA and PBIL algorithm

Tested according to pytest test


nevergrad/optimization/optimizerlib.py
```diff
@@ -1109,4 +1164,74 @@ def __call__(self, instrumentation: Union[int, Instrumentation],
...
+    def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optional[int] = None, num_workers: int = 1) -> None:
+        super().__init__(instrumentation, budget=budget, num_workers=num_workers)
+        num_categories = 2
+        self.p: np.ndarray = np.ones((1, self.dimension)) / num_categories
+        self.llambda = 2 * (self.budget if self.budget is not None else max(num_workers, 40))
+        self._value_candidate: Optional[Tuple[float, np.ndarray]] = None
+
+    def _internal_ask_candidate(self) -> base.Candidate:
+        unif = self._rng.uniform(size=self.dimension)
+        data = (unif > self.p[0]).astype(float)
```
teytaud: Ok p[0][i] is the probability of generating 0.
But below it looks like we increase p[0][i] if the winner has a 1. Did I miss something ?

nevergrad/optimization/optimizerlib.py
```diff
@@ -1109,4 +1164,74 @@ def __call__(self, instrumentation: Union[int, Instrumentation],
...
+        num_categories = 2
+        self.p: np.ndarray = np.ones((1, self.dimension)) / num_categories
+        self.alpha = 0.3
+        self.llambda = max(100, num_workers)  # size of the population
+        self.mu = self.llambda // 2  # number of selected candidates
+        self._population: List[Tuple[float, np.ndarray]] = []
+
+    def _internal_ask_candidate(self) -> base.Candidate:
+        unif = self._rng.uniform(size=self.dimension)
+        data = (unif > self.p[0]).astype(float)
```
teytaud: Maybe we should check the compatibility with the instrumentation ?
Another possibility is to extend the algorithm to more instrumentations, but maybe later.

PR #276, opened 2019-09-19
Add confidence lines

jrapin:
## Types of changes

So far there was no confidence information on benchmark plots. This adds confidence area in transparency to prevent overloading plots (too much...).

## Checklist


nevergrad/benchmark/plotting.py
```diff
@@ -243,6 +248,16 @@ def __init__(
...
         # self._fig.tight_layout()
 
+    @staticmethod
+    def _get_confidence_arrays(vals: Dict[str, np.ndarray], log: bool = False) -> Tuple[np.ndarray, np.ndarray]:
+        loss = vals["loss"]
+        conf = vals["loss_std"] / np.sqrt(vals["num_eval"] - 1)
+        if not log:
+            return loss - conf, loss + conf
+        lloss = np.log10(loss)
+        lstd = 0.434 * conf / loss
```
jrapin: This should be crosschecked

jrapin: For the record: https://faculty.washington.edu/stuve/log_error.pdf

teytaud: Why not simply << return loss - conf, loss + conf >> so that we plot << x, log(y+-std) >> ?
The goal is to plot something inspired from << x, log(y) +- std(log(y)) >> rather than << x, log(y +- std(y)) >> ?

jrapin: From https://faculty.washington.edu/stuve/log_error.pdf it seems it's not a good practice to use log(y+-std).
From what I have observed, it does look better as well.

PR #277, opened 2019-09-24
Adding a "real world" experiment in Nevergrad

teytaud:
## Types of changes

Cool to have an aggregation of real-world experiments.
Could be added:
- OT's works on games (to be moved from dfoptim, after cleaning)
- LM's work on adversarial examples
- BR's work on GANs

## Checklist

### DIFF
--- a/nevergrad/benchmark/experiments.py
+++ b/nevergrad/benchmark/experiments.py
@@ -270,7 +270,22 @@ def realworld(seed: Optional[int] = None) -> Iterator[Experiment]:
     # - The 007 game
     # - TODO FIXME DO NOT SUBMIT
     
-    # Needed for 007.
+    # MLDA stuff, except the Perceptron.
+    funcs: List[InstrumentedFunction] = [
+        _mlda.Clustering.from_mlda(name, num, rescale) for name, num in [("Ruspini", 5), ("German towns", 10)] for rescale in [True, False]
+    ]
+    funcs += [
+        _mlda.SammonMapping.from_mlda("Virus", rescale=False),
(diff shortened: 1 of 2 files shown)


nevergrad/benchmark/experiments.py
```diff
@@ -263,6 +263,56 @@ def spsa_benchmark(seed: Optional[int] = None) -> Iterator[Experiment]:
...
+        _mlda.SammonMapping.from_mlda("Employees"),
+    ]
+    # funcs += [_mlda.Perceptron.from_mlda(name) for name in ["quadratic", "sine", "abs", "heaviside"]]
+    funcs += [_mlda.Landscape(transform) for transform in [None, "square", "gaussian"]]
+
+    # Adding ARCoating.
+    funcs += [ARCoating()]
+
+    # 007 with 100 repetitions, both mono and multi architectures.
+    base_env = rl.envs.DoubleOSeven(verbose=False)
```
teytaud: @jrapin    will DoubleOSeven as instantiated here be a deterministic optimization problem ?

teytaud: (for this xp, not centered on noise, I'd like to get rid of noise issues at least for the moment --- a second version of "realworld" should be stochastic, with different problems)

jrapin: It will be noisy:
- random agent is not deterministic
- the other agents are set to `deterministic=False`

The noise should be slightly smoothed out by `num_repetitions=100`, but that's still noisy I would think

jrapin: At some point I'll probably update this part, because it's a bit complicated so there's probably a better way to configure this... I have no idea how though :s

PR #297, opened 2019-10-23
adding multiobjective optimization in nevergrad

teytaud:
The doc of the main function says it all: this provides a simple tool for converting a multiobjective problem into a single objective one.

def multiobjective_minimization(functions, bad_values) -> InstrumentedFunction:
    """Given several functions, and threshold on their values (above which solutions are pointless),
    this function returns a single-objective function, correctly instrumented, the minimization of which
    yields a solution to the original multiobjective problem.
    
    functions: objective functions, to be minimized, of the original multiobjective problem.
    bad_values: bad_values[i] is a threshold above which x is pointless if functions[i](x) > bad_values[i].
    
    Returns an objective function to be minimized (it is a single objective function).
    Warning: this function is not stationary.
    The minimum value obtained for this objective function is -h,
    where h is the hypervolume of the Pareto front obtained, given bad_values as a reference point.
    When called with no argument, this monoobjective function returns an approximate Pareto front,
    reaching the current hypervolume.
    """

## Types of changes

People at Dagstuhl's seminar on the Theory of Randomized Search Heuristics told me that this is a missing feature so I do it.

I did not even check that what I do makes any sense.
It's very likely I am destroying the type testing.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,12 +92,17 @@ The [optimization documentation](docs/optimization.md) contains more information
 Let us minimize x1 and x2 (two objective functions) assuming that values above 2.5 are of no interest.
 ```python
 import nevergrad as ng
+import numpy as np
 
-# Let us work in dimension 7.
-# The instrumentation parameter can be an Instrumentation (describing variables and their types) or just the dimension in the
-# continuous case; it is not mandatory if functions provided as arguments are equipped with an instrumentation.
-f = ng.optimizers.multiobjective_minimization([lambda x: x[0], lambda x: x[1]], bad_values=[2.5, 2.5], 7)
-optimizer = ng.optimizers.CMA(instrumentation=f.instrumentation, budget=100)  # 2 is the dimension, 100 is the budget.
+f = ng.optimizers.multiobjective_minimization([lambda x: x[0], lambda x: x[1]], bad_values=[2.5, 2.5])
+
+# Let us work in dimension 3.
+# Let us check that it works.
+print(f(np.array([1,2,3])))
+
+# The parameter ``instrumentation`` can be replaced by a more sophisticated instrumentation (docs/instrumentation.md).
+# We here just use the dimension as we assume simple continuous rescaled variables (mean 0, std 1).
+optimizer = ng.optimizers.CMA(instrumentation=3, budget=100)  # 2 is the dimension, 100 is the budget.
 recommendation = optimizer.optimize(f)
 
 
--- a/README.md
+++ b/README.md
@@ -93,7 +93,10 @@ Let us minimize x1 and x2 (two objective functions) assuming that values above 2
 ```python
 import nevergrad as ng
 
-f = ng.functions.multiobjective_minimization([lambda x: x[0], lambda x: x[1]], bad_values=[2.5, 2.5])
+# Let us work in dimension 7.
+# The instrumentation parameter can be an Instrumentation (describing variables and their types) or just the dimension in the
+# continuous case; it is not mandatory if functions provided as arguments are equipped with an instrumentation.
+f = ng.optimizers.multiobjective_minimization([lambda x: x[0], lambda x: x[1]], bad_values=[2.5, 2.5], 7)
 optimizer = ng.optimizers.CMA(instrumentation=f.instrumentation, budget=100)  # 2 is the dimension, 100 is the budget.
 recommendation = optimizer.optimize(f)
 
--- a/README.md
+++ b/README.md
@@ -87,6 +87,20 @@ print(list(sorted(ng.optimizers.registry.keys())))
 
 The [optimization documentation](docs/optimization.md) contains more information on how to use several workers, take full control of the optimization through the `ask` and `tell` interface and some pieces of advice on how to choose the proper optimizer for your problem.
 
+## Multiobjective minimization with Nevergrad
+
+```python
+import nevergrad as ng
+
+f = ng.functions.multiobjective_minimization([lambda x: x[0], lambda x: x[1]])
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -87,6 +87,21 @@ print(list(sorted(ng.optimizers.registry.keys())))
 
 The [optimization documentation](docs/optimization.md) contains more information on how to use several workers, take full control of the optimization through the `ask` and `tell` interface and some pieces of advice on how to choose the proper optimizer for your problem.
 
+## Multiobjective minimization with Nevergrad
+
+Let us minimize x1 and x2 (two objective functions) assuming that values above 2.5 are of no interest.
+```python
+import nevergrad as ng
+
+f = ng.functions.multiobjective_minimization([lambda x: x[0], lambda x: x[1]], bad_values=[2.5, 2.5])
```
jrapin: if `f` is properly instrumented you should do `ng.optimizers.CMA(instrumentation=f.instrumentation, budget=100)`

nevergrad/functions/multiobjective.py
```diff
@@ -0,0 +1,395 @@
...
+        """
+        for i in range(index):
+            node.prev[i].next[i] = node
+            node.next[i].prev[i] = node
+            if bounds[i] > node.cargo[i]:
+                bounds[i] = node.cargo[i]
+            
+__all__ = ["hypervolume_kmax", "hypervolume"]
+
+if __name__ == "__main__":
```
jrapin: this is not used, right?

nevergrad/functions/multiobjective.py
```diff
@@ -0,0 +1,395 @@
...
+        nodes of the node that is reinserted are in the list.
+
+        """
+        for i in range(index):
+            node.prev[i].next[i] = node
+            node.next[i].prev[i] = node
+            if bounds[i] > node.cargo[i]:
+                bounds[i] = node.cargo[i]
+            
+__all__ = ["hypervolume_kmax", "hypervolume"]
```
jrapin: not used

nevergrad/functions/multiobjective.py
```diff
@@ -0,0 +1,395 @@
...
+    reference point *ref*.
+    """
+    if ref is None:
+        ref = numpy.array([3000000] * len(pointset[0]))
+    warnings.warn("Falling back to the python version of hypervolume "
+        "module. Expect this to be very slow.", RuntimeWarning)
+    hv = _HyperVolume(ref)
+    return hv.compute(pointset)
+
+def multiobjective_minimization(functions, bad_values) --> InstrumentedFunction:
```
jrapin: I don't understand what are `bad_values`
When I understand all this better I can restructure and make it more compatible with the rest of the code.

nevergrad/functions/multiobjective.py
teytaud: it's just a test --- I remove.

README.md
teytaud: Ok, done!

nevergrad/functions/multiobjective.py
teytaud: bad_values[i] is the threshold at which the objective function functions[i] becomes so bad that we don't want to keep any point, even if they are good for other objective function values.

nevergrad/benchmark/test_experiments.py
```diff
@@ -21,7 +21,7 @@ def test_experiments_registry(name: str, maker: Callable[[], Iterator[experiment
     with patch("shutil.which", return_value="here"):  # do not check for missing packages
         with datasets.mocked_data():  # mock mlda data that should be downloaded
             check_maker(maker)  # this is to extract the function for reuse if other external packages need it
-        if "mlda" not in name:
+        if "mlda" not in name and "moo" not in name:
```
teytaud: I have no idea what this test is doing and why moo does not pass it.

README.md
```diff
@@ -87,6 +87,21 @@ print(list(sorted(ng.optimizers.registry.keys())))
...
+optimizer = ng.optimizers.CMA(instrumentation=f.instrumentation, budget=100)  # 2 is the dimension, 100 is the budget.
+recommendation = optimizer.optimize(f)
+
+
+# The function embeds its Pareto-front:
+print("My Pareto front:", f())
```
jrapin: given that the calls to `f` can be performed in another process by an executor, this actually won't always work :s we'll need to discuss this :s

nevergrad/benchmark/test_experiments.py
jrapin: This test checks that running twice the same experiments with the same seed provide the same results. We'll investigate later why this occurs if it still does.

jrapin:
Thinking about it, I think the simplest path is to add wrapper function:

```python
class AggregateFunction:

      def __self__(multi_objective_function: Callable[..., Tuple[float, ...]], higher_bounds: Tuple[float, ...]) -> None:
          self.multi_objective_function = multi_objective_function
          ...

      def compute_aggregate_loss(losses: Tuple[float, ...], *args, **kwargs, candidate: Candidate) -> None:
          ...

      def __call__(self, *args, **kwargs):
          losses = self.compute_aggregate_loss(*args, **kwargs)
          return self.compute_aggregate_loss(losses, *args, **kwargs)

     @property
     def pareto_front(self) -> List[Tuple[Tuple[Any, ...], Dict[str, Any]]]:
          return ...  # list of args and kwargs?
```

This way, we could add some simple hack in `minimize` to only call `multi_objective_function` remotely and compute the aggregate locally. It's not very beautiful, but should be straightforward and effective. Wdyt?

PR #298, opened 2019-10-23
Add cheap constraints management

teytaud:
When constraints are super fast and finding satisfiable points is not complicated, a simple solution is to regenerate solutions as many times as required for having constraints satisfied, without spending the budget in terms of objective function evaluations.
It is important to not do this by fitness penalization, so that there is no premature convergence.

## Types of changes

People at Dagstuhl told me that this would be cool.
This is only for cheap constraints that can be repeatedly sampled without problem.
We just repeat mutation operators until we satisfy constraints.
As discussed with Patrick Spettel (thx) this is better than rejecting the mutation a posteriori, as it avoids premature convergence.

I've not even run my code a single time.

## Checklist


teytaud:
Hmmm, a problem with this PR is that the constraint is not part of the Experiment, but an argument of the optimizer.
@jrapin  should I add list_of_cheap_positivity_constraints as a data of Experiments ?

nevergrad/optimization/base.py
```diff
@@ -527,7 +528,13 @@ def minimize(
                 if verbosity and new_sugg:
                     print(f"Launching {new_sugg} jobs with new suggestions")
                 for _ in range(new_sugg):
+                    
                     args = self.ask()
+                    num_tries = 0
+                    while cheap_constraints_checker is not None and cheap_constraints_checker(c) and num_tries < 1000:
+                        self._num_ask -= 1
+                        args = self.ask()
+                        num_tries += 1
```
jrapin: Better loop:
```python
if cheap_constraints_checker is not None:
    for _ in range(1000):
         args = self.ask()
         if cheap_constraints_checker(*args.args, **args.kwargs)
              break
```
Other remarks:
- I don't think you should decrease num `ask`, you **did** ask so it does not really make sense.
- You should also do a `tell`, because many algorithms (I'm mostly thinking of `DE`) will break if you do `ask` without `tell`.

teytaud: Interesting.
It is on purpose that I decrease num_ask. The principle is that constraints are cheap - so we should not pay one budget evaluation when we redraw. No idea of a better design :-(

teytaud: Discussed in chat and solved, thx Jeremy.

PR #300, opened 2019-10-26
[PR on PR] constraints management

jrapin:


docs/optimization.md
```diff
@@ -67,6 +67,28 @@ recommendation = optimizer.provide_recommendation()
...
+We basically repeat mutations until we get a satisfiable point.
+Let us say that we want to minimize `(x[0]-.5)**2 + (x[1]-.5)**2` under the constraint `x[0] >= 1`.
+```python
+import nevergrad as ng
+
+def square(x):
+    return sum((x - .5)**2)
+
+optimizer = ng.optimizers.OnePlusOne(instrumentation=2, budget=100)
+optimizer.instrumentation.set_cheap_constraint_checker(lambda x: x[0] >= 1)
```
jrapin: that's the only additional line required and then everything should work

nevergrad/optimization/base.py
```diff
@@ -391,33 +391,36 @@ def ask(self, cheap_constraints_checker: Any = None) -> Candidate:
...
-        while True:  # Until we satisfy cheap constraints.
+        for callback in self._callbacks.get("ask", []):
+            callback(self)
+        MAX_TENTATIVES = 1000
+        current_num_ask = self.num_ask
+        for k in range(MAX_TENTATIVES):
             # call callbacks for logging etc...
-            for callback in self._callbacks.get("ask", []):
-                callback(self)
+            is_suggestion = False
```
jrapin: this all method is veeery sensitive...  some things are a bit tricky (like the fact that suggestions are  not registered will others are)

nevergrad/optimization/base.py
```diff
@@ -391,33 +391,36 @@ def ask(self, cheap_constraints_checker: Any = None) -> Candidate:
...
                 candidate = self._internal_ask_candidate()
                 # only register actual asked points
-                if candidate.uid in self._asked:
-                    raise RuntimeError(
-                        "Cannot submit the same candidate twice: please recreate a new candidate from data.\n"
-                        "This is to make sure that stochastic instrumentations are resampled."
-                    )
-                self._asked.add(candidate.uid)
-            assert candidate is not None, f"{self.__class__.__name__}._internal_ask method returned None instead of a point."
-            if not cheap_constraints_checker or cheap_constraints_checker(candidate.data):
```
jrapin: better not check the constraint multiple times even if cheap. Also, it must use args and kwargs, not data which is internal

nevergrad/optimization/differentialevolution.py
```diff
@@ -71,7 +71,7 @@ class _DE(base.Optimizer):
 
     def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optional[int] = None, num_workers: int = 1) -> None:
         super().__init__(instrumentation, budget=budget, num_workers=num_workers)
-        self._penalize_cheap_violations = False
+        self._penalize_cheap_violations = True
```
jrapin: isnt it the only algorithm which absolutely requires that the violations are penalized??

nevergrad/optimization/optimizerlib.py
```diff
@@ -564,7 +564,6 @@ class NoisyBandit(base.Optimizer):
     """
 
     def _internal_ask(self) -> ArrayLike:
-        self._penalize_cheap_violations = False
```
jrapin: not needed, since it's the default

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -321,3 +321,12 @@ def test_instrumentation_optimizer_reproducibility() -> None:
     optimizer = optimizerlib.RandomSearch(instrumentation, budget=10)
     recom = optimizer.minimize(_square)
     np.testing.assert_equal(recom.kwargs["y"], 67)
+
+
+def test_constrained_optimization() -> None:
+    instrumentation = inst.Instrumentation(x=inst.var.Array(1), y=inst.var.Scalar())
+    optimizer = optimizerlib.OnePlusOne(instrumentation, budget=100)
```
jrapin: We're up to maaaaany more problems eventually with this constraints I think :D let's solve them when we find them.
First one: OnePlusOne always ask zero at the first iteration, so because num_ask did not change it would just always ask for zero with never satisfies the constraint. I made some hack to avoid this (temporarily updating num_ask) but it may not be very robust)

teytaud: I agree for one plus one, but I had solve this in a previous PR (by trying only a limited number of times), maybe just git rebase.

nevergrad/optimization/differentialevolution.py
teytaud: argh my god sorry

docs/optimization.md
teytaud: Cool!

nevergrad/optimization/test_optimizerlib.py
jrapin: For now I am sending an error when I cant bypass the error, I'll change to avoid this

PR #302, opened 2019-10-26
[PR on PR] Chaining algorithms

jrapin:
Update to use Parametrized family and add tests


nevergrad/optimization/optimizerlib.py
```diff
@@ -1203,68 +1203,74 @@ def _internal_tell_candidate(self, candidate: base.Candidate, value: float) -> N
...
-                for b in self._list_of_budgets:
-                    if self._num_ask >= b:
-                        algo_index += 1
-                # Ok, that algorithm chooses the next point to evaluate.
-                return self._list_of_algorithms[algo_index].ask()
-        
-            def _internal_tell_candidate(self, candidate: base.Candidate, value: float) -> None:
-                # Let us inform all algorithms
-                for l in self._list_of_algorithms:
-                    l.tell(candidate, value)
```
jrapin: past algorithms did not require telling

nevergrad/optimization/optimizerlib.py
```diff
@@ -1203,68 +1203,74 @@ def _internal_tell_candidate(self, candidate: base.Candidate, value: float) -> N
...
-                    if self._list_of_budgets[i] == -1:
-                        self._list_of_budgets[i] = num_workers
-                    if self._list_of_budgets[i] == -2:
-                        self._list_of_budgets[i] = self.dimension
-
-            def _internal_ask_candidate(self) -> base.Candidate:
-                # Which algorithm are we playing with ?
-                algo_index = 0
-                for b in self._list_of_budgets:
-                    if self._num_ask >= b:
```
jrapin: this should probably have been a cumulative budget, right?

nevergrad/optimization/optimizerlib.py
```diff
@@ -1203,68 +1203,74 @@ def _internal_tell_candidate(self, candidate: base.Candidate, value: float) -> N
...
-        list_of_algorithms = self._list_of_algorithms
-        list_of_budgets = self._list_of_budgets
-
-        # We create the class we want to use.
-        class chain(base.Optimizer):
-
-            def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optional[int] = None, num_workers: int = 1) -> None:
-                super().__init__(instrumentation, budget=budget, num_workers=num_workers)
-                self._list_of_budgets = list_of_budgets
-                self._list_of_algorithms = [l(instrumentation, budget, num_workers) for l in list_of_algorithms]
```
jrapin: each optimizer should be initialized with its own budget: initializing LHS with budget 100 if it will be asked only 5 times is suboptimal, since only 5 of the points will be asked and there would not be any warranty on their distribution

nevergrad/optimization/optimizerlib.py
```diff
@@ -1203,68 +1203,74 @@ def _internal_tell_candidate(self, candidate: base.Candidate, value: float) -> N
...
+    """
+    _optimizer_class = _Chain
+
+    def __init__(self, optimizers: Sequence[Union[base.OptimizerFamily, Type[base.Optimizer]]],
+                 budgets: Sequence[Union[str, int]]) -> None:
+        # Either we have the budget for each algorithm, or the last algorithm uses the rest of the budget, so:
+        self.budgets = tuple(budgets)
+        self.optimizers = tuple(optimizers)
+        assert len(self.optimizers) == len(self.budgets) + 1
+        assert all(x in ("dimension", "num_workers") or x > 0 for x in self.budgets)  # type: ignore
```
jrapin: i've changed to string, it's easier to read

nevergrad/optimization/optimizerlib.py
```diff
@@ -1203,68 +1203,74 @@ def _internal_tell_candidate(self, candidate: base.Candidate, value: float) -> N
...
-                        self._list_of_budgets[i] = self.dimension
-
-            def _internal_ask_candidate(self) -> base.Candidate:
-                # Which algorithm are we playing with ?
-                algo_index = 0
-                for b in self._list_of_budgets:
-                    if self._num_ask >= b:
-                        algo_index += 1
-                # Ok, that algorithm chooses the next point to evaluate.
-                return self._list_of_algorithms[algo_index].ask()
```
jrapin: Here is one of the reasons I preferred avoiding propagating cheap constraints through `ask`: the constraints would have needed to be propagated to subalgorithms as well. With the constraints within the instrumentation, this is automatic

nevergrad/optimization/optimizerlib.py
teytaud: Ok cleaner than my jurassic-style version :-)

nevergrad/optimization/optimizerlib.py
teytaud: Argh good catch

nevergrad/optimization/optimizerlib.py
teytaud: Right.

PR #303, opened 2019-11-03
Archive optimizer jobs in optimizer._archived_jobs

louismartin:
## Types of changes

One recurring problem that I have is to access jobs after they are being told. For instance when the optimization is finished, sometime I need to retrieve all jobs that were started from the beginning.

I ran multiple optimizations with this change

## Checklist

### DIFF
--- a/nevergrad/optimization/base.py
+++ b/nevergrad/optimization/base.py
@@ -230,6 +230,7 @@ class Optimizer:  # pylint: disable=too-many-instance-attributes
         # to make optimize function stoppable halway through
         self._running_jobs: List[Tuple[Candidate, JobLike[float]]] = []
         self._finished_jobs: Deque[Tuple[Candidate, JobLike[float]]] = deque()
+        self._archived_jobs: List[Tuple[Candidate, JobLike[float]]] = []
 
     @property
     def _rng(self) -> np.random.RandomState:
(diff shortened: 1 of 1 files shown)


nevergrad/optimization/base.py
```diff
@@ -512,7 +513,7 @@ def minimize(
                 while self._finished_jobs:
                     x, job = self._finished_jobs[0]
                     self.tell(x, job.result())
-                    self._finished_jobs.popleft()  # remove it after the tell to make sure it was indeed "told" (in case of interruption)
+                    self._archived_jobs.append(self._finished_jobs.popleft())  # remove it after the tell to make sure it was indeed "told" (in case of interruption)
```
jrapin: what is the aim here?
I'm afraid this way will cause the memory to explode in some cases (very long local trainings for instance) :s for the point archive I've needed to implement a pruning every now and then (which does not satisfy me either :s)

louismartin: Then `_archived_jobs` could be a deque with fixed length to prevent that, what do you think?

The goal is to retrieve the jobs after training, for instance using slurm jobs, I want to be able to retrieve all their ids so that I can do other stuff with it (e.g. with their stdout).

jrapin: my prefered way would be to wrap the executor, I don't think that would generalize well to other type of executor than the slurm one

louismartin: Ok then I'll close the PR

Issue #211, opened 2019-05-17
Strange behaviour relating to bounds.


mpdunne:
I am observing some strange behaviour relating to **bounds**.

Let's find the minimum of a standard simple function:

    def ackley( x, a=20, b=0.2, c=2*pi ):
        x = np.asarray_chkfinite(x)  # ValueError if any NaN or Inf
        n = len(x)
        s1 = sum( x**2 )
        s2 = sum( cos( c * x ))
        return -a*exp( -b*sqrt( s1 / n )) - exp( s2 / n ) + a + exp(1)

I'd like to use Nevergrad to optimise this. The true minimum value is at

    minimum = np.zeros(n)

which gives a calculated value of

    4.440892098500626e-16.

I'm using Nevergrad as follows:

    from nevergrad import instrumentation as inst
    from nevergrad.optimization import optimizerlib

    n = 5
    
    # Define the variable ranges...
    instrum = inst.Instrumentation(*[inst.var.Array(1).asfloat().bounded(-10, 10) for _ in range(5)])
    
    # Optimise the function...
    optimizer = optimizerlib.TwoPointsDE(instrumentation=instrum, budget=10000)
    recommendation = optimizer.optimize(lambda *args: ackley(np.array(args)))
    print(recommendation)

This rightly returns:

    >>> Candidate(args=(-1.4223567512956928e-11, 1.161667735276156e-11, 1.5909965473446878e-11, 3.142964723880178e-11, 2.1667036390793597e-12), kwargs={})

I.e., roughly zero. However, if I change the bounds to `[-10, k]`, with `k<0` I get:
 
    k=-2
    >>> Candidate(args=(-2.97922395034131, -2.979223956777669, -2.9792239575180597, -2.979223949490683, -2.979223956813635), kwargs={})

    k=-4
    >>> Candidate(args=(-4.986180458441733, -4.9861804545723105, -4.986180452383419, -4.986180455508544, -4.986180459693069), kwargs={})

...the minimum value is always claimed to be `(k-1)*np.ones(n)`, however in all cases it should be at `k*np.ones(n)`.

My question is: what's going on here? Why is Nevergrad getting these answers so consistently wrong? Have I done something wrong with the bounds? Any help would be appreciated!

Thanks!

jrapin:
One quick comment first, you could simply use:
```python
instrum = inst.Instrumentation(inst.var.Array(n).bounded(-10, 10))  # makes an np.array of shape (n,)
[...]

PR #304, opened 2019-11-04
Add a Log variable

jrapin:
## Types of changes

In ML, many parameters are log-distributed, while properly instrumenting it with proper initial value (if we consider the center as a good default) and mutation rate is actually quite difficult. There should be more coming along those lines since this is related to #211 for instance

## Checklist


nevergrad/instrumentation/variables.py
```diff
@@ -334,3 +336,46 @@ class Scalar(Array):
     def __init__(self, dtype: Type[Union[float, int]] = float):
         super().__init__(1)
         self.asscalar(dtype=dtype)
+
+
+class Log(Variable):
```
jrapin: @teytaud instrumenting is actually still very difficult with my way of handling it. I want to simplify it quite a bit (and make it more flexible).
This instrumentation makes (at least) 3 choices I would like you to validate, for a variable like Log(0.001, 0.1):
1. we would tend to sample in the middle to start (i.e. 0.01)
2. we would like mutations to act the same at the center and around the boundaries.
3. we should be able to mutate to one extremity in the order of 10 iterations

This leads to:
1. we linearize [0.001, 0.1] to [-x, x] with x a given choice, 0 maps to 0.01
2. to keep the boundary, we use clipping because arctan makes it actually very difficult to reach the boundary (it maps it to infinity, so you never reach it), the farther you are from the center, the more the mutations are skewed
3. by default I map to [-10, 10] so that reaching the boundaries can be done in the order of magnitude of 10 mutations (can't remember the random walk stuff :D). Not sure it's the best default.

Any thoughts?

jrapin:
approved by @teytaud

PR #306, opened 2019-11-04
Add a simple way to log parameters

jrapin:
## Types of changes

## Checklist


nevergrad/optimization/callbacks.py
```diff
@@ -0,0 +1,80 @@
...
+from pathlib import Path
+import numpy as np
+from . import base
+
+
+class OptimizationPrinter:
+    """Printer to register as callback in an optimizer, for printing
+    best point regularly.
+
+    Parameters
```
jrapin: TODO: Update description + explain arrays

nevergrad/optimization/callbacks.py
```diff
@@ -0,0 +1,88 @@
...
+    ----------
+    num_eval: int
+        max number of evaluation before performing another print
+    num_sec: float
+        max number of seconds before performing another print
+    """
+
+    def __init__(self, num_eval: int = 0, num_sec: float = 60) -> None:
+        self._num_eval = max(0, int(num_eval))
```
louismartin: Why the  max here?
In case that someone puts a negative number? If that's a case, I think that raising an error would be better than silently changing the value.

nevergrad/optimization/callbacks.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Parameters
+    ----------
+    num_eval: int
+        max number of evaluation before performing another print
+    num_sec: float
+        max number of seconds before performing another print
+    """
+
+    def __init__(self, num_eval: int = 0, num_sec: float = 60) -> None:
```
louismartin: I would use more explicit names for num_eval and num_sec, but that might be a personal preference.

nevergrad/optimization/callbacks.py
```diff
@@ -0,0 +1,88 @@
...
+        self._num_eval = max(0, int(num_eval))
+        self._last_time: Optional[float] = None
+        self._num_sec = num_sec
+
+    def __call__(self, optimizer: base.Optimizer, *args: Any, **kwargs: Any) -> None:
+        if self._last_time is None:
+            self._last_time = time.time()
```
louismartin: Maybe I did not understand the logic, but from what I get, `self._last_time` is never reset to `time.time()` after printing the recommandation

nevergrad/optimization/callbacks.py
```diff
@@ -0,0 +1,88 @@
...
+        if (time.time() - self._last_time) > self._num_sec or (self._num_eval and not optimizer.num_tell % self._num_eval):
```
louismartin: I know we already had this debate but I think that `self._num_eval > 0` is way more explicit than `self._num_eval`.

And I don't get what `not optimizer.num_tell % self._num_eval` is supposed to do, for me we would need `(optimizer.num_tell - self._last_eval) > self._num_eval`

nevergrad/optimization/callbacks.py
```diff
@@ -0,0 +1,88 @@
...
+    def __call__(self, optimizer: base.Optimizer, candidate: base.Candidate, value: float) -> None:
+        data = {"#instrumentation": optimizer.instrumentation.name,
+                "#session": self._session,
+                "#num-ask": optimizer.num_ask,
+                "#num-tell": optimizer.num_tell,
+                "#num-tell-not-asked": optimizer.num_tell_not_asked,
+                "#loss": value}
+        params = dict(candidate.kwargs)
+        params.update({f"arg{k}": arg for k, arg in enumerate(candidate.args)})
+        data.update({x: y for x, y in params.items() if not isinstance(y, np.ndarray)})
```
louismartin: I would use k, v instead of x, y :p

nevergrad/optimization/callbacks.py
```diff
@@ -0,0 +1,88 @@
...
+        data.update({x + "_0": y.ravel()[0] for x, y in params.items() if isinstance(y, np.ndarray)})
```
louismartin: What's the purpose here?
Only log the first item of numpy arrays ?
Maybe it should be added as a comment

nevergrad/optimization/callbacks.py
```diff
@@ -0,0 +1,88 @@
...
+        try:  # avoid bugging as much as possible
+            with self._filepath.open("a") as f:
+                f.write(json.dumps(data) + "\n")
+        except:
```
louismartin: Bare except 😨

jrapin: now converted to broad except... I still want to catch all so as to avoid callbacks from making optimization explode for some weird corner case

nevergrad/optimization/callbacks.py
jrapin: `num_tell_period`  and `time_period_s`?

nevergrad/optimization/callbacks.py
louismartin: Ok I guess it depends on what you need to do.
If the last print was triggered by self._num_sec, then I would expect it to wait for at least self._num_eval or self._num_sec before printing again, whereas here it will print once with self._num_sec and maybe once directly after with self._num_eval.

Issue #314, opened 2019-11-07
Uncentered clipping bounds


jrapin:
Clipping does not make centered data, which can lead to the variable being ignored

jrapin:
Solved by #391

PR #323, opened 2019-11-14
Implement the backbone for future parametrization

jrapin:
## Types of changes

### What
This provide the backbone for a brand new way to perform instrumentation (but leaves nevergrad unchanged for now, the new methods are not plugged in this PR). Currently the instrumentation is just a stateless bridge between the value space/candidates (the space of parameters feeded to the functions) and the optimization space (where the optimizers perform). The new `Parameter`  class in this PR now provides a stateful `Instrumentation` (i.e. an `Instrumentation` with a value), and therefore merges both `Instrumentation`  and `Candidate` together.
To simplify understanding of the code, be aware that they are 3 main types of parameters:
- **data**: those are the actual base parameters leaf values. Everything is based on a generic `Array` parameters providing an interface for parametrizing np.ndarray. This is then subclassed into `Scalar` and `Log` (a shortcut for a log-distributed scalar).
- **containers**: those are parameters that contain other parameters like `Dict` and `Tuple`. This also includes the former `Instrumentation` which is basically a tuple of a tuple (args) and a dict (kwargs).
- **choice**: those are conditional parameters, which select one parameter among several choices. This includes `Choice` which is our former `SoftmaxCategorical`, and `TransitionChoice` which offers an equivalent to `OrderedDiscrete`, where mutations are defined as transitions between current element and the previous and next ones.

### Why
This new way of doing things will hopefully make our lives simpler:
- **more explicit**: the value serves as an initial value, and updating it is very easy, while it is currently tricky and not very intuitive (cf #314 for an example of how to provide an initial value in the current and the proposed method)
- **enables mutation and recombination**: along the conversion to and from the standardized optimization space (`get_std_data`, `set_std_data`) the new API defines `mutate` and `recombine` methods. This will enable a whole new range of optimizers which would relie on these methods for their implementations (eg: currently crossovers in `DE` are predefined, now they could just call `recombine` which could be cleverly chosen/implemented by the user)
- **simpler to experiment**: given that each `Parameter` has a `value` field, and `mutate` method, it is now easy to make sure that mutations behave as expected (correct standard deviation, logarithmic mutation etc). Eg. call `param.value`, then `param.mutate()` then recheck the value and check how it has been modified. When this new parametrization is plugged into the current framework, `Optimizer.minimize` function will preferably take only one argument which will be the output of the `value` attribute. A specific class of the new `Parameter` (called `Instrumentation` as before for compatibility) will however also allow to keep the current behavior with `args` and `kwargs` for those who prefer.
- **simplifies the code**: eventually this very `Parameter` class will replace `Variable`, `Candidate` and `CandidateMaker`  altogether.
- **enables endogeneous parameters**: evolution strategies sometimes define endogeneous parameters, i.e. internal parameters which can also mutate. This will also be possible with the new parametrization by having variable mutating using a standard deviation (for instance) which is itself a `Parameter`. This way, we will be able to define complex algorithms in a very straightforward way.
- **enables conditional variables**: `SoftmaxCategorical` variable is now called `Choice` and allows arguments which are themselves `Parameter`, meaning they mutate/recombine as well. This way, it will for eg be possible to define choices between different deep neural network optimizers for instance, each taking parameters with different names and/or scales which can be tuned with `nevergrad` as well.

### Advice

For reviewing, you may want to start by having a loop at the tests (`test_parameters.py`) with `core.py` alongside to check the docstrings (if a non-private method docstring is missing/not clear let me know). You can then check `container.py`, `data.py` and `choice.py` for implementation "details".

### Next steps

1. Plug this new framework into the current framework, without breaking anything.
2. Let new and old frameworks live along for a couple of months with deprecation warnings.
3. Kill the old framework.

## Checklist


jrapin:
TODO list:
- [ ] array bounds
- [ ] array distribution
- [ ] instrumentation
- [ ] log parameter
- [ ] sigma parameter?
- [ ] ordered choice
- [ ] unordered choice?

jrapin:
Possible future improvement (not to be included here):
- stochastic int casting for `Array` (instead of rounding)

jrapin:
Now ready to review :)

nevergrad/parametrization/core.py
```diff
@@ -0,0 +1,497 @@
...
+
+    def __init__(self, **subparameters: t.Any) -> None:
+        self.uid = uuid.uuid4().hex
+        self.parents_uids: t.List[str] = []
+        self._subparameters = None if not subparameters else Dict(**subparameters)
+        self._dimension: t.Optional[int] = None
+
+    @property
+    def descriptors(self) -> Descriptors:
+        return Descriptors(deterministic=True, continuous=True)
```
danthe3rd: As `deterministic` and `continuous` are not related, why not have 2 `property` in the class directly?

nevergrad/parametrization/core.py
```diff
@@ -0,0 +1,497 @@
...
+P = t.TypeVar("P", bound="Parameter")
+D = t.TypeVar("D", bound="Dict")
+
+
+class Descriptors(t.NamedTuple):
+    """Provides access to a set of descriptors for the parametrization
+    This can be used within optimizers.
+    """
+    deterministic: bool = True
+    continuous: bool = True
```
danthe3rd: I'm not sure if I understand what a continuous parameter is

nevergrad/parametrization/core.py
```diff
@@ -0,0 +1,497 @@
...
+    def _internal_set_std_data(self: P, data: np.ndarray, instance: P, deterministic: bool = False) -> P:
+        if data.size:
+            raise ValueError(f"Constant dimension should be 0 (got data: {data})")
+        return instance
+
+    def spawn_child(self: P) -> P:
+        return self  # no need to create another instance for a constant
+
+
+def as_parameter(param: t.Any) -> Parameter:
```
danthe3rd: Nit: Maybe easier to put that in the `Parameter` class directly? `Parameter.from(x)`

nevergrad/parametrization/core.py
```diff
@@ -0,0 +1,497 @@
...
+    def value(self, value: t.Dict[str, t.Any]) -> None:
+        if set(value) != set(self._parameters):
+            raise ValueError(f"Got input keys {set(value)} but expected {set(self._parameters)}")
+        for key, val in value.items():
+            as_parameter(self._parameters[key]).value = val
+
+    def get_value_hash(self) -> t.Hashable:
+        return tuple(sorted((x, y.get_value_hash()) for x, y in self._parameters.items() if isinstance(y, Parameter)))
+
+    def _internal_get_std_data(self: D, instance: D) -> np.ndarray:
```
danthe3rd: nit: do you need to type `self`?

nevergrad/parametrization/core.py
```diff
@@ -0,0 +1,497 @@
...
+                self._dimension = 0
+        return self._dimension
+
+    def mutate(self) -> None:
+        """Mutate subparameters of the instance, and then its value
+        """
+        self.subparameters.mutate()
+        data = self.get_std_data()  # pylint: disable=assignment-from-no-return
+        # let's assume the random state is already there (next class)
+        self.set_std_data(data + self.random_state.normal(size=data.shape), deterministic=False)  # type: ignore
```
danthe3rd: Dont' we want to change the `uid` in `mutate`?

nevergrad/parametrization/core.py
```diff
@@ -0,0 +1,497 @@
...
+    @random_state.setter
+    def random_state(self, random_state: np.random.RandomState) -> None:
+        self._set_random_state(random_state)
+
+    def _set_random_state(self, random_state: np.random.RandomState) -> None:
+        self._random_state = random_state
+        if self._subparameters is not None:
+            self.subparameters._set_random_state(random_state)
+
+    def spawn_child(self: P) -> P:
```
danthe3rd: Shouldn't this be in `BaseParameter`?

nevergrad/parametrization/core.py
jrapin: yes, because `instance` must be of the same type as `self`

nevergrad/parametrization/core.py
jrapin: In current instrumentation we have 2 different properties, but little by little (like as of today) more different properties like this one are creeping in (like "noisy_function", "discrete" etc). I'd rather we keep such "descriptors" contained somewhere to ease the interface, but I'm not 100% sure either

nevergrad/parametrization/core.py
jrapin: basically, the parametrization acts as a function between the optimizer and the actual function. Knowing whether the whole instrumentation+function is continuous can be useful for the optimizer.
A few examples:
- instrumentation of a scalar is continuous
- instrumentation of an integer is not continuous
- instrumentation of Ordered discrete is not continiuous
- instrumentation of Softmax categorical is continuous (!) but noisy (we sample based on softmax values, that evolve continuously, but the sampling acts like a noise)
- instrumentation of Softmax categorical in deterministic mode (just sample the one with max probability) is not continuous but not noisy

danthe3rd: So having 2 parameters (1 scalar + 1 integer) will not behave the same as a single Dict parameter which contains 1 scalar and 1 integer?

nevergrad/parametrization/core.py
jrapin: I thought of something like this but if it's in `Parameter` then all parameters would inherit from it, while this is not to be used outside of the implementations (it would not make sense). I don't like much the current way of doing it either though.
Another option was to convert **everything** as parameter, but I thought that would be messier. Not sure though...

nevergrad/parametrization/core.py
jrapin: Well... "BaseParameter" does not really exit actually, it's just Parameter split in two for readability... I need to make it more explicit in the comments I think

nevergrad/parametrization/core.py
jrapin: Nop, otherwise we'd need to update `uid` in `recombine` as well, then if we perform 1 mutation and 1 recombination, we lose track of the initial parent :s

jrapin: Basically the process is: spawn a child, then do whatever you want with it

jrapin:
> always wrap the `Constant` in the class constructors?

I'll try

nevergrad/parametrization/core.py
jrapin: what do you mean by "2 parameters"?

nevergrad/parametrization/core.py
jrapin: If the whole world is against me, then so be it lol, it's now merged

nevergrad/parametrization/core.py
```diff
@@ -0,0 +1,504 @@
...
+    @property
+    def value(self) -> t.Any:
+        raise NotImplementedError
+
+    @value.setter
+    def value(self, value: t.Any) -> t.Any:
+        raise NotImplementedError
+
+    def spawn_child(self: BP) -> BP:
+        """Creates a new which only shares the same random generator than its parent
```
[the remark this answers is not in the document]
jrapin: I'm updating it to:
Creates a new instance which shares the same random generator than its parent,
is sampled from the same data, and mutates independently from the parent.

Is it better?

nevergrad/parametrization/core.py
```diff
@@ -0,0 +1,504 @@
...
+        other instances.
+
+        Parameters
+        ----------
+        *others: Parameter
+            other instances of the same type than this instance.
+        """
+        raise NotSupportedError(f"Recombination is not implemented for {self.name}")  # type: ignore
+
+    def get_std_data(self: BP, instance: t.Optional[BP] = None) -> np.ndarray:
```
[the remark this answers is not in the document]
jrapin: I'll go with standardized if no other option (cant find any :s )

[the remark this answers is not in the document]
jrapin: get_opt(imization)_data? get_lin(earized)_data?

nevergrad/parametrization/choice.py
```diff
@@ -0,0 +1,245 @@
...
+        index = super()._find_and_set_value(value)
+        self._index = index
+        # force new probabilities
+        out = discretization.inverse_softmax_discretization(self.index, len(self))
+        self.weights.set_std_data(out, deterministic=True)
+        return index
+
+    def _draw(self, deterministic: bool = True) -> None:
+        weights = self.weights.value
+        random = False if deterministic or self._deterministic else self.random_state
```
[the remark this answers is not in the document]
jrapin: it's the other way round, sometimes I want to sample determininistically, but usually not. I dont have any better option in mind to do this :s

nevergrad/parametrization/choice.py
```diff
@@ -0,0 +1,245 @@
...
+
+    @property
+    def transitions(self) -> Array:
+        """The weights used to draw the value
+        """
+        return self["transitions"]  # type: ignore
+
+    @property
+    def position(self) -> Scalar:
+        """The weights used to draw the value
```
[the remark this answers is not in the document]
jrapin: well spotted

nevergrad/parametrization/core.py
danthe3rd: I'm just wondering how this will be used in practice I guess, and how it makes sense to aggregate those to the higher level

jrapin: oh, knowing that the whole (function+parametrization) is continuous or not, deterministic or not,  can be used inside the optimizers to choose a better strategy. Olivier is currently working on competence maps which basically does exactly that: depending on the case, it selects different algorithms

danthe3rd: Right makes sense thanks!

nevergrad/parametrization/data.py
```diff
@@ -0,0 +1,353 @@
...
+        """Bounds all real values into [a_min, a_max] using a provided method
+
+        Parameters
+        ----------
+        a_min: float or None
+            minimum value
+        a_max: float or None
+            maximum value
+        method: str
+            "clipping", "constraint", "tanh" or "arctan"
```
danthe3rd: Setting bounds with "clipping" is quite clear to me, but I'm not sure if I understand what a clipping "constraint" or "tanh" is... Maybe adding an example can help?
Also the parameter `full_range_sampling` can be a little confusing. If I want to do uniform sampling, it seems more intuitive to have a function `param.uniform_sampling(minimum, maximum[, transform])`

nevergrad/parametrization/data.py
```diff
@@ -0,0 +1,353 @@
...
+        Returns
+        -------
+        self
+        """
+        if sigma is not None:
+            # just replace if an actual Parameter is provided as sigma, else update value (parametrized or not)
+            if isinstance(sigma, core.Parameter) or isinstance(self.subparameters._parameters["sigma"], core.Constant):
+                self.subparameters._parameters["sigma"] = core.as_parameter(sigma)
+            else:
+                self.sigma.value = sigma  # type: ignore
```
danthe3rd: nit: Not sure if relevant here, but I would prefer to rather `assert isinstance(sigma, float)` if that can avoid a type ignore - and silent error fail

nevergrad/parametrization/data.py
```diff
@@ -0,0 +1,353 @@
...
+        checker = BoundChecker(*bounds)
+        if not checker(self.value):
+            raise ValueError("Current value is not within bounds, please update it first")
+        if not (a_min is None or a_max is None):
+            if (bounds[0] >= bounds[1]).any():  # type: ignore
+                raise ValueError(f"Lower bounds {a_min} should be strictly smaller than upper bounds {a_max}")
+        # update instance
+        transforms = dict(clipping=trans.Clipping, arctan=trans.ArctanBound, tanh=trans.TanhBound)
+        if method in transforms:
+            if self.exponent is not None and method != "clipping":
```
danthe3rd: nit: replace the hardcoded strings with constants, same for `self._parameters["sigma"]` (maybe when you are done with experimenting :) )

nevergrad/parametrization/data.py
```diff
@@ -0,0 +1,353 @@
...
+            if shape is not None:
+                raise ValueError(err_msg)
+            self._value = np.array(init, copy=False)
+        elif shape is not None:
+            assert isinstance(shape, tuple) and all(isinstance(n, int) for n in shape)
+            self._value = np.zeros(shape)
+        else:
+            raise ValueError(err_msg)
+        self.integer = False
+        self.exponent: t.Optional[float] = None
```
danthe3rd: Maybe it's easier to set a fixed exponent to say `e` and then change the sigma/initial value?
`(exponent=2 * e, sigma=1, init=(2 * e)^1)` should be equivalent to `(exponent=e, sigma=1 + ln(2), init=e^(1 + ln(2)))` no?
```
(2*e)^(x) -> (2*e)^(x + sigma) = (2 * e) ^ (x + 1)
e^(x.(1 + ln(2))) -> e^(x.(1 + ln(2)) + sigma) = e^(x.(1 + ln(2)) + (1 + ln(2))) = e^((x+1).(1 + ln(2))) = (2 * e) ^ (x + 1)
```

nevergrad/parametrization/choice.py
```diff
@@ -0,0 +1,245 @@
...
+        nums = sorted(int(k) for k in self.choices._parameters)
+        for k in nums:
+            choice = core.as_parameter(self.choices[k])
+            try:
+                choice.value = value
+            except Exception:  # pylint: disable=broad-except
+                pass
+            else:
+                index = int(k)
+                break
```
danthe3rd: nit: you can put that inside the `try`? Also the `int` cast is not needed

nevergrad/parametrization/data.py
jrapin: I've updated the description of `constraint` and `full_range_sampling`.
However having a method `uniform_sampling`  would not solve it: the aim is to have the algorithms use the `sample` when they need to create new instances from scratch. So this method must change behavior when we know that a new point should be created using a uniform sampling (so that the algorithm does not have to know anything about it + it does not make sense for any "non-data" parameter). Is it clearer? I tried to make this clearer in the description as well.
On the other hand, I don't like the name `full_range_sampling`, but I'm not sure using `uniform_sampling` would be better, since it's potentially **log-**uniform sampling sometimes as well. If you have a better name let me know :s

nevergrad/parametrization/data.py
jrapin: It can actually be a `np.ndarray` or a `float`, which makes typing a bit complicated :s

nevergrad/parametrization/data.py
jrapin: What do you mean? I don't understand it :s

nevergrad/parametrization/data.py
jrapin: I wanted to avoid the equations altogether :D 
I like it the current because that one can define mutations changes in one space or another
I don't like it the current way because both ways are equivalent so it's redundant...
For now I'm thinking less equations will hopefully make it less buggy, but I need to think about it today then.

nevergrad/parametrization/choice.py
jrapin: there must be an absolutely minimum number of stuff in the try, so that we don't hide any weird errors (like not being able to cast k to int)
Then again as you say, k is supposedly already an int so that should not be a problem indeed^^
So, fair enough, this is simplified

nevergrad/parametrization/data.py
danthe3rd:
```suggestion
            if self.exponent is not None and method != type(self).CLIPPING:
```

And also `self._parameters["sigma"]` -> `self._parameters[type(self).SIGMA]`

A bit longer to write, but allows to catch errors / typos

nevergrad/parametrization/data.py
```diff
@@ -0,0 +1,364 @@
...
+        self
+        """
+        if sigma is not None:
+            # just replace if an actual Parameter is provided as sigma, else update value (parametrized or not)
+            if isinstance(sigma, core.Parameter) or isinstance(self.subparameters._parameters["sigma"], core.Constant):
+                self.subparameters._parameters["sigma"] = core.as_parameter(sigma)
+            else:
+                self.sigma.value = sigma  # type: ignore
+        if exponent is not None:
+            if self.bound_transform is not None and not self.bound_transform.name.startswith("Cl"):
```
danthe3rd: nit

jrapin:
Merging this for now, dicussions can move on to #391 
This second step will aim at plugging the new pipeline inside the old one, and have both compatibility

PR #336, opened 2019-11-19
Optimizer splitting the problem into several problems

teytaud:
## Types of changes

In evolutionary computation, people sometimes split the objective function into blocks.
This is in particular the case in multiobjective optimization when working directly with a list of points for optimizing their hypervolume.

## Checklist

- [ ] I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).


teytaud:
@brockho

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
         return self.avg
 
 
+@registry.register
+class Splitter(base.Optimizer):
```
brozi: I'd prefer to call this class SplitOptimizer since it isn't obvious that something called Splitter is an optimizer without looking at the source (I would expect a class called Splitter to have a split() method returning an optimizer or something else).

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+
+    Be careful! The variables refer to the deep representation used by optimizers.
+    For example, a categorical variable with 5 possible values becomes 5 continuous variables.
+    """
+
+    def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optional[int] = None, num_workers: int = 1, num_optims: Optional[int] = None, num_vars: Optional[List[Any]] = None) -> None:
+        super().__init__(instrumentation, budget=budget, num_workers=num_workers)
+        if num_vars:
+            if num_optims:
+                assert num_optims == len(num_vars)
```
brozi: Add a comment, eg
assert num_optims == len(num_vars), f"The number of optimizers {num_optims} should be equal to the length of the variables vector {len(num_vars)}"

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+            else:
+                num_optims = len(num_vars)
+            assert sum(num_vars) == self.dimension
```
brozi: Add a comment f"The total number of variables ({sum(num_vars)}) is different from the dimension ({self.dimension})"

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+        else:
+            if not num_optims:  # if no num_vars and no num_optims, just assume 3.
```
brozi: The comment says assume 3 but you assume 2. I think assuming 2 is slightly more intuitive.

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+                num_optims = 2
+            # num_vars not given: we will distribute variables equally.
+        if num_optims > self.dimension:
+            num_optims = self.dimension
+            if num_vars:
+                num_vars = num_vars[:num_optims]
```
brozi: With the asserts above, this modifies num_vars only if the number of optimizers is greater than the dimension and some elements of num_vars are 0 or negative. I think it would be better to remove this and assert that every number in num_vars is strictly positive (or positive or 0, in which case you want to keep all the non zero num_vars and not the first dimension ones).

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+        self.num_optims = num_optims
+        self.optims: List[Any] = []
+        self.num_vars: List[Any] = num_vars if num_vars else []
+        self.instrumentations: List[Any] = []
+        self.num_yoyo = 0
+        for i in range(self.num_optims):
+            if not self.num_vars or len(self.num_vars) < i+1:
+                self.num_vars += [(self.dimension // self.num_optims) + (self.dimension % self.num_optims > i)]
+            
+            assert len(self.num_vars) > 0
```
brozi: It would be cleaner to check that at the beginning of the init with something like all(var_dim > 0 for var_dim in num_vars)

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+            assert self.num_vars[-1] >= 0
```
brozi: I'm not sure how informative this assert is.
It would feel cleaner to check that with something like all(var_dim > 0 for var_dim in num_vars) at the end of that loop.

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+            self.instrumentations += [Instrumentation(inst.variables.Array(self.num_vars[i]).affined(1, 0))]
+            assert len(self.optims) == i
+            if self.num_vars[i] > 1:
+                self.optims += [CMA(self.instrumentations[i], budget, num_workers)]  # noqa: F405
```
brozi: Shouldn't we extract CMA and RandomSearch at least as Constants in the class and maybe as parameters with these default values ?

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+        self.num_vars: List[Any] = num_vars if num_vars else []
+        self.instrumentations: List[Any] = []
+        self.num_yoyo = 0
+        for i in range(self.num_optims):
+            if not self.num_vars or len(self.num_vars) < i+1:
+                self.num_vars += [(self.dimension // self.num_optims) + (self.dimension % self.num_optims > i)]
+            
+            assert len(self.num_vars) > 0
+            assert self.num_vars[-1] >= 0
+            self.instrumentations += [Instrumentation(inst.variables.Array(self.num_vars[i]).affined(1, 0))]
```
brozi: I'm not sure why we always want to define the variable this way but it's probably because I lack context.

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+
+    def _internal_ask_candidate(self) -> base.Candidate:
+        data: List[Any] = []
+        #np.random.seed(1)
+        for i in range(self.num_optims):
+            opt = self.optims[i]
+            #print("working with optimizer ", opt, " in dim ", self.num_vars[i])
+            data += list(opt.ask().data)
+        #print(data)
+        #print("ask almost over", self.num_yoyo)
```
brozi: Remove the comments before pushing

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,94 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+        self.num_yoyo += 1
```
brozi: call it number_asks or remove it since it doesn't seem to be used.

nevergrad/optimization/optimizerlib.py
teytaud: added a check for all i, thx

nevergrad/optimization/optimizerlib.py
teytaud: you're right, i now assert that the user provides a meaningful num_vars with everything fine.

nevergrad/optimization/optimizerlib.py
teytaud: I go for 2, thx.

nevergrad/optimization/optimizerlib.py
teytaud: done

nevergrad/optimization/optimizerlib.py

nevergrad/optimization/optimizerlib.py
teytaud: oh my god thx, forgotten from debug print

teytaud:
thx @brozi for useful reviews

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,84 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+    Given that all optimizers have the same number of variables, we can also do:
+    opt = SplitOptimizer(instrumentation=10, num_workers=3, num_optims=5)
+
+    This is 5 parallel (by num_workers = 5).
+
+    Be careful! The variables refer to the deep representation used by optimizers.
+    For example, a categorical variable with 5 possible values becomes 5 continuous variables.
+    """
+
+    def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optional[int] = None, num_workers: int = 1, num_optims: Optional[int] = None, num_vars: Optional[List[Any]] = None, default_multivariate_optimizer: Any = CMA, default_monovariate_optimizer: Any = RandomSearch) -> None:
```
brozi: Call the optimizers without the default_ prefix. You only use the default if you don't set it.
I'm still a noob at forcing parameter types in python but can't you use base.Optimizer instead of Any for their types ?

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,84 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+            num_optims = self.dimension
+        self.num_optims = num_optims
+        self.optims: List[Any] = []
+        self.num_vars: List[Any] = num_vars if num_vars else []
+        self.instrumentations: List[Any] = []
+        for i in range(self.num_optims):
+            if not self.num_vars or len(self.num_vars) < i+1:
+                self.num_vars += [(self.dimension // self.num_optims) + (self.dimension % self.num_optims > i)]
+            
+            assert self.num_vars[i] >= 0, f"At least one variable per optimizer."
```
brozi: At least one variable per optimizer is self.num_vars[i] > 0

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,84 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+
+    Be careful! The variables refer to the deep representation used by optimizers.
+    For example, a categorical variable with 5 possible values becomes 5 continuous variables.
+    """
+
+    def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optional[int] = None, num_workers: int = 1, num_optims: Optional[int] = None, num_vars: Optional[List[Any]] = None, default_multivariate_optimizer: Any = CMA, default_monovariate_optimizer: Any = RandomSearch) -> None:
+        super().__init__(instrumentation, budget=budget, num_workers=num_workers)
+        if num_vars:
+            if num_optims:
+                assert num_optims == len(num_vars), f"The number of optimizers should match len(num_vars)."
```
brozi: You don't need the f at the beginning of your string if you don't add any formatting. 
I would add {len(num_vars)} and {optimizers} as in my previous comment since I think it can help debugging.

nevergrad/optimization/optimizerlib.py
```diff
@@ -762,6 +763,84 @@ def _internal_provide_recommendation(self) -> ArrayLike:
...
+            else:
+                num_optims = len(num_vars)
+            assert sum(num_vars) == self.dimension, f"sum(num_vars) should be equal to the dimension."
```
brozi: Same as above.

nevergrad/optimization/optimizerlib.py

nevergrad/optimization/optimizerlib.py
teytaud: The variables are defined this way only for the "internal" instrumentations, viewed by the different optimizers. I think it's hard to forward the original instrumentation to the internal algorithms as the splitting might split several continuous variables born from a single variables. SplitOptimizers only see the purely continuous problem.

nevergrad/optimization/optimizerlib.py
teytaud: Right! Done.

nevergrad/optimization/optimizerlib.py
teytaud: yes! right.

nevergrad/optimization/optimizerlib.py
teytaud: Thx, done.

PR #337, opened 2019-11-20
distributed collaborative optimization: multiagent power systems

teytaud:
## Types of changes

Work in progress with LRI / Inria / Cnrs / TAU / Paris-Sud.
Part of discussions with colleagues in Diego Suarez.

## Checklist


teytaud:
(Morgane added as a reviewer for her interest in combinations AI/OR)

applications/infra.py
```diff
@@ -0,0 +1,241 @@
...
+
+class Agent():
+    """An agent has an input size, an output size, a number of layers, a width of its internal layers 
+    (a.k.a number of neurons per hidden layer)."""
+    def __init__(self, input_size, output_size, layers=2, layer_width=20):
+        assert(layers >= 2)
+        self.input_size = input_size
+        self.output_size = output_size
+        self.layers = []
+        self.layers += [np.random.rand(int(input_size), int(layer_width))]
```
brozi: I'm a bit surprised that you're coding your neural network in numpy instead of using pytorch or another framework. 
It may be good to try to initialize your hidden layers with Xavier instead of a uniform distribution and a neural net framework would do it for you.

applications/infra.py
```diff
@@ -0,0 +1,241 @@
...
+year_to_day_ratio = 1.  # Ratio between variation of consumption in the year and variation of consumption in the day
+constant_to_year_ratio = 1.
+dam_output_factor = 1.0
+back_to_normal = 0.5  # How much of the gap with normal is cancelled at each iteration.
+consumption_noise = 0.1
+num_thermal_plants = 7
+number_of_years = 1
+
+average_consumption = constant_to_year_ratio * year_to_day_ratio
+thermal_power_capacity = [c * average_consumption for c in list(np.random.rand(num_thermal_plants))]
```
brozi: So that probably won't make much difference here because this is called only once but np.random.rand(num_thermal_plants) is already iterable so you don't need the list() thing. Multiplying by c would be more efficient and I think you can keep it as a numpy array:
` thermal_power_capacity = average_consumption * np.random.rand(num_thermal_plants)`
If you really want a list in the end you can write it `list(average_consumption * np.random.rand(num_thermal_plants))`

applications/infra.py
```diff
@@ -0,0 +1,241 @@
...
+thermal_power_prices = [c for c in list(np.random.rand(num_thermal_plants))]
```
brozi: This line can be written more simply
`thermal_power_prices = list(np.random.rand(num_thermal_plants)`

applications/infra.py
```diff
@@ -0,0 +1,241 @@
...
+
+for i in range(N):
+    dam_managers += [Agent(10 + N + 2*num_thermal_plants, 1)]
+
+def simulate_power_system(input_x):
+    x = [r for r in input_x]
```
brozi: `x = list(input_x)`
Do you need that ?

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,154 @@
...
+            base_x = [np.cos(2*pi*t/24.), np.sin(2*pi*t/24.), np.cos(2*pi*t/(365*24)), np.sin(2*pi*t/(365*24)), needed, self.average_consumption, self.year_to_day_ratio, self.constant_to_year_ratio, self.back_to_normal, self.consumption_noise]
+            x = list(base_x + self.thermal_power_capacity + self.thermal_power_prices + stocks)
+    
+            # Prices as a decomposition tool!
+            price: List[float] = [a.GetOutput(np.array(x))[0][0] for a in dam_managers]
+            volume: List[float] = [s for s in stocks]
+            dam_index: List[int] = list(range(len(price)))
+            price += self.thermal_power_prices
+            volume += self.thermal_power_capacity
+            dam_index += [-1] * len(price)
```
brozi: Why do you add a len(price) times -1 at the end of your indices list ? I think you only use this list in the for loop after and you never use indices after len(price) (which are the ones defined with the range on line 108).

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,154 @@
...
+            if t == 0:
+                consumption = base_consumption
+            else:
+                consumption = max(0., consumption + self.consumption_noise*(np.random.rand()-.5) + self.back_to_normal * (base_consumption - consumption))
+            consumption_per_time_step += [consumption]
+            # Water usage.
+            needed = consumption
+    
+            # Setting inputs for all agents.
+            base_x = [np.cos(2*pi*t/24.), np.sin(2*pi*t/24.), np.cos(2*pi*t/(365*24)), np.sin(2*pi*t/(365*24)), needed, self.average_consumption, self.year_to_day_ratio, self.constant_to_year_ratio, self.back_to_normal, self.consumption_noise]
```
brozi: Kind of a detail but math.cos is much faster than np.cos on scalars.

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,154 @@
...
+                if dam_index[i] >= 0:
+                    hydro_prod[dam_index[i]] += production
+                    stocks[dam_index[i]] -= production
+                    assert(stocks[dam_index[i]] >= -1e-7)
+                else:
+                    cost += production * price[i]
+                needed -= production
+            # Cost in case of failures -- this is
+            # harming industries and hospitals.
+            cost += 500. * needed
```
brozi: Extract the cost factor as a Constant

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,154 @@
...
+            for i in range(len(price)):
+                if needed == 0:
+                    break
+                production = min(volume[i], needed)
+                if dam_index[i] >= 0:
+                    hydro_prod[dam_index[i]] += production
+                    stocks[dam_index[i]] -= production
+                    assert(stocks[dam_index[i]] >= -1e-7)
+                else:
+                    cost += production * price[i]
```
brozi: That never happens.

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,154 @@
...
+            volume: List[float] = [s for s in stocks]
+            dam_index: List[int] = list(range(len(price)))
+            price += self.thermal_power_prices
+            volume += self.thermal_power_capacity
+            dam_index += [-1] * len(price)
+            
+            assert(len(price) == N + self.num_thermal_plants)
+            hydro_prod: List[float] = [0.] * N
+            for i in range(len(price)):
+                if needed == 0:
```
brozi: So I don't know if you are interested in making your code faster and you would need to profile it to know where you should optimize your code but this loop could be done much more efficiently with numpy arrays. It would also be more readable in my opinion.

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,154 @@
...
+                else:
+                    cost += production * price[i]
+                needed -= production
+            # Cost in case of failures -- this is
+            # harming industries and hospitals.
+            cost += 500. * needed
+            hydro_prod_per_time_step += hydro_prod
+        return cost  # Other data of interest: , hydro_prod, hydro_prod_per_time_step, consumption_per_time_step
+
+    def __init__(self, num_stocks: int = 13, depth: int = 3, width: int = 3) -> None:
```
brozi: Move __init__ to be the top method in this class

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,154 @@
...
+        # Number of stocks (dams).
+        self.N = num_stocks
+        N = self.N
+        # Parameters describing the problem.
+        self.year_to_day_ratio = 2.  # Ratio between variation of consumption in the year and variation of consumption in the day
+        self.constant_to_year_ratio = 1.
+        self.back_to_normal = 0.5  # How much of the gap with normal is cancelled at each iteration.
+        self.consumption_noise = 0.1
+        self.num_thermal_plants = 7
+        self.number_of_years = 1
```
brozi: These parameters are defined in infra.py. They should be defined only once to make sure they are the same everywhere. If you keep them here,  it feels like most of them should be parameters of the init function instead of hardcoded constants (you may want to change the horizon or number of plants).

applications/infra.py
```diff
@@ -0,0 +1,241 @@
...
+    hydro_hour = deblock24(hydro_ts)
+    ax.plot(np.linspace(1,24,len(hydro_hour)), hydro_hour, label='dam ' + str(i) + ' prod')
+ax.set_xlabel('time step')
+ax.set_ylabel('production per hour')
+ax.legend(loc='best')
+plt.savefig("ps.png")
+plt.show()
+plt.waitforbuttonpress()
+recommendation = optimizer.recommend()  #minimize(square)
+print(recommendation)  # optimal args and kwargs
```
brozi: This stuff is mostly duplicated (very similar to powersystems/core.py). Keep only one or it will be very confusing.

applications/infra.py
```diff
@@ -0,0 +1,241 @@
...
+    """An agent has an input size, an output size, a number of layers, a width of its internal layers 
+    (a.k.a number of neurons per hidden layer)."""
+    def __init__(self, input_size, output_size, layers=2, layer_width=20):
+        assert(layers >= 2)
+        self.input_size = input_size
+        self.output_size = output_size
+        self.layers = []
+        self.layers += [np.random.rand(int(input_size), int(layer_width))]
+        for i in range(layers-2):
+            self.layers += [np.random.rand(int(layer_width), int(layer_width))]
```
brozi: I was thinking that maybe you should initialize your weights with Xavier instead of something uniform. I guess it would also work better with some of the genetic algorithms.

applications/infra.py
teytaud: I remove this file but report the corresponding modification in the other similar file.

applications/infra.py
teytaud: Done.

nevergrad/functions/powersystems/core.py
teytaud: Actually we must sort the list. This is fixed. Thanks a lot...

nevergrad/functions/powersystems/core.py
teytaud: Done, thx.

nevergrad/functions/powersystems/core.py
teytaud: Done.

nevergrad/functions/powersystems/core.py
teytaud: Should happen, if "needed" is too high ?

applications/infra.py
teytaud: Done.

applications/infra.py
teytaud: Done + some related modifications.

nevergrad/functions/powersystems/core.py
teytaud: Done.

nevergrad/functions/powersystems/core.py
teytaud: Done.

nevergrad/functions/powersystems/core.py
teytaud: Presumably it's negligible compared to matrix products (which are in numpy arrays).
I improve readability (at least a little bit).

nevergrad/functions/powersystems/core.py
brozi: What I mean is that dam_index is defined as `list(range(len(price))) + [-1] * len(price)` in l108 and l111.
You loop on i in range(len(price)) so dam_index[i] will take the values of range(len(price)) (by the way it would be clearer if you were to just do for dam_index in range(len(price)) instead of this loop).
Then your test `if dam_index[i] >=0` will always return true and you will never execute what is in else.

Maybe that's not the test you actually want to do ?

brozi: oh ok sorry I didn't read your update before reading this. You corrected it !

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,234 @@
...
+            # Water usage.
+            needed = consumption
+    
+            # Setting inputs for all agents.
+            base_x = [math.cos(2*pi*t/24.), math.sin(2*pi*t/24.), math.cos(2*pi*t/(365*24)), math.sin(2*pi*t/(365*24)), needed, self.average_consumption, self.year_to_day_ratio, self.constant_to_year_ratio, self.back_to_normal, self.consumption_noise]
+            x = np.concatenate((base_x, self.thermal_power_capacity, self.thermal_power_prices, stocks))
+    
+            # Prices as a decomposition tool!
+            price: np.ndarray = np.asarray([a.GetOutput(np.array(x))[0][0] for a in dam_managers])
+            volume: np.ndarray = np.asarray([s for s in stocks])
```
brozi: this is the same as np.asarray(stocks). You could also define it direcly on l144 and rename it as capacity since volume doesn't fit the thermal power plants.
`capacity = np.concatenate((stocks, self.thermal_power_capacity))`

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,234 @@
...
+                stocks[i] += 0.5*(1.+math.cos(2*pi*t/(24*365) + delay[i])) * np.random.rand()
+            # Consumption model.
+            base_consumption = (self.constant_to_year_ratio*self.year_to_day_ratio 
+                    +0.5*self.year_to_day_ratio*(1.+math.cos(2*pi*t/(24*365))) + 0.5*(1.+math.cos(2*pi*t/24)))
+            if t == 0:
+                consumption = base_consumption
+            else:
+                consumption = max(0., consumption + self.consumption_noise*(np.random.rand()-.5) + self.back_to_normal * (base_consumption - consumption))
+            consumption_per_time_step += [consumption]
+            # Water usage.
```
brozi: Shouldn't the comment say energy usage ? Since you can also use thermal power plants.

applications/infra.py
teytaud: Removed. Code rewritten for less lists and more np arrays.

nevergrad/functions/powersystems/core.py
teytaud: Fixed.

teytaud: Yes, fixed.

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,234 @@
...
+            needed = consumption
+    
+            # Setting inputs for all agents.
+            base_x = [math.cos(2*pi*t/24.), math.sin(2*pi*t/24.), math.cos(2*pi*t/(365*24)), math.sin(2*pi*t/(365*24)), needed, self.average_consumption, self.year_to_day_ratio, self.constant_to_year_ratio, self.back_to_normal, self.consumption_noise]
+            x = np.concatenate((base_x, self.thermal_power_capacity, self.thermal_power_prices, stocks))
+    
+            # Prices as a decomposition tool!
```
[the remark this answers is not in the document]
teytaud: Removed for simplicity.

nevergrad/functions/powersystems/core.py
teytaud: Done.

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,233 @@
...
+import math
+import matplotlib.pyplot as plt
+from typing import Any
+from typing import List
+import numpy as np
+from nevergrad.common.typetools import ArrayLike
+from ... import instrumentation as inst
+from ...instrumentation.core import Variable
+from ...instrumentation.multivariables import Instrumentation
+pi = np.pi
```
brozi: You are already doing from `math import sqrt, tan, pi` so you can remove this (unless you really need the pi from numpy instead of math, in that case do `from numpy import pi`)

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,233 @@
...
+        w = [w for w in ww]
+        assert(len(w) == self.GetParamNumbers())
+        for l in self.layers:
+            s = np.prod(l.shape)
+            l = np.reshape(np.array(w[:s]), l.shape)
+            w = w[s:]
+
+    def GetOutput(self, inp):
+        output = np.array(inp).reshape(1, len(inp))
+        for l in [self.layers[i] for i in range(len(self.layers) - 1)]:
```
brozi: You can write it more simply as 
`for l in self.layers[:-1]:`

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,233 @@
...
+            year_to_day_ratio: float = 2.,  # Ratio between std of consumption in the year and std of consumption in the day.
+            constant_to_year_ratio: float = 1.,  # Ratio between constant baseline consumption and std of consumption in the year.
+            back_to_normal: float = 0.5,  # Part of the variability which is forgotten at each time step.
+            consumption_noise: float = 0.1,  # Instantaneous variability.
+            num_thermal_plants: int = 7,  # Number of thermal plants.
+            num_years: int = 1,  # Number of years.
+            failure_cost: float = 500.,  # Cost of not satisfying the demand. Equivalent to an expensive infinite capacity thermal plant.
+            ) -> None:
+        # Number of stocks (dams).
+        self.N = num_stocks
```
brozi: Call it `self.num_stocks` not `N`. Actually `num_dams` would probably be clearer (you wouldn't need a comment to understand what it is).

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,233 @@
...
+        return [m.GetParamNumbers() for m in self.dam_managers]
+
+    def _simulate_power_system(self, x: np.ndarray) -> float:
+        failure_cost = self.failure_cost  # Cost of power demand which is not satisfied (equivalent to a expensive infinite thermal group).
+        dam_managers = self.dam_managers
+        for a in dam_managers:
+            assert(len(x) >= a.GetParamNumbers())
+            a.SetParams(np.array(x[:a.GetParamNumbers()]))
+            x = x[a.GetParamNumbers():]
+        assert(len(x) == 0)
```
brozi: You wouldn't need this assert and the `x = x[a.GetParamNumbers():]` above if you were to assert `assert(len(x) == a.GetParamNumbers())` above

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,233 @@
...
+        N = int(self.N)
```
brozi: Same as above, please rename N.

nevergrad/functions/powersystems/core.py
```diff
@@ -0,0 +1,233 @@
...
+        cost = 0.
+        # Loop on time steps.
+        num_time_steps = int(365*24*self.number_of_years)
+        consumption = 0.
+        hydro_prod_per_time_step: List[Any] = []
+        consumption_per_time_step: List[float] = []
+        for t in range(num_time_steps):
+    
+            # Rain
+            for i in range(N):
```
brozi: Change i to `dam` or `dam_index`

nevergrad/functions/powersystems/core.py
teytaud: Ok!

nevergrad/functions/powersystems/core.py
teytaud: Yeah much better :-) done.

nevergrad/functions/powersystems/core.py
teytaud: Done.

nevergrad/functions/powersystems/core.py
teytaud: No comprendo! x decreases at each pass in the loop.
assert(len(x) == a.GetParamNumbers())  is not true.

nevergrad/functions/powersystems/core.py
teytaud: done.

nevergrad/functions/powersystems/core.py
brozi: Oh yes you're right I thought there was something to get all the parameters in one go. I would have done it differently (without modifying x) but I guess this is fine.

PR #344, opened 2019-11-26
Adding an experiment including chaining, an experiment including cheap constraints, and fixing a few bad docstrings

teytaud:
## Types of changes

There was no Experiment with constraints.
There was no experiment with chaining, now there is an experiment with chaining.
We also add an option for initialization with budget = sqrt(whole optimization budget).

not much

## Checklist


nevergrad/optimization/optimizerlib.py
```diff
@@ -1217,7 +1217,7 @@ def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optiona
     def _optimizers(self) -> List[base.Optimizer]:
         if not self._optimizers_:
             self._optimizers_ = []
-            conv = {"num_workers": self.num_workers, "dimension": self.dimension}
+            conv = {"num_workers": self.num_workers, "dimension": self.dimension, "sqrt": int(np.sqrt(self.budget)) if self.budget else self.num_workers}
             budgets = [conv[b] if isinstance(b, str) else b for b in self._parameters.budgets]
```
brozi: Reading l 1212 I think self._parameters.budgets and Chaining will always be [10] unless you modify it explicitely outside of the methods in _Chain. Then budgets will always be [10] and this conv thing doesn't change anything. Am I missing something ?

nevergrad/optimization/optimizerlib.py
```diff
@@ -1217,7 +1217,7 @@ def __init__(self, instrumentation: Union[int, Instrumentation], budget: Optiona
     def _optimizers(self) -> List[base.Optimizer]:
         if not self._optimizers_:
             self._optimizers_ = []
-            conv = {"num_workers": self.num_workers, "dimension": self.dimension}
+            conv = {"num_workers": self.num_workers, "dimension": self.dimension, "sqrt": int(np.sqrt(self.budget)) if self.budget else self.num_workers}
```
brozi: I know it isn't part of this commit but conv really makes me think of convolution while I guess it means conversion here ? Maybe we should call it conversion or budget_conversion instead (or call a method called compute_budget(budget:str) instead).

teytaud:
> Unless I missed something, there seems to be an issue with a previous commit that's affecting the behaviour of your code.

looks like everything compiles and runs now ?

nevergrad/optimization/optimizerlib.py
teytaud: i chose "converter" and will let Jeremy make a decision for this.

nevergrad/optimization/optimizerlib.py
teytaud: I added a bunch of asserts displaying self._optimizers and self._parameters, and I am confident
that the value of self._parameters is overwritten correctly.
(that's the magical @jrapin code for ParametrizedFamily, and I am so happy that it is also not trivial for you!)

nevergrad/benchmark/experiments.py
```diff
@@ -173,6 +194,22 @@ def illcondi(seed: Optional[int] = None) -> Iterator[Experiment]:
                 yield Experiment(function.duplicate(), optim, budget=budget, num_workers=1, seed=next(seedg))
 
 
+@registry.register
+def chainillcondi(seed: Optional[int] = None) -> Iterator[Experiment]:
```
brozi: same, use _

nevergrad/benchmark/experiments.py
```diff
@@ -191,6 +228,24 @@ def illcondipara(seed: Optional[int] = None) -> Iterator[Experiment]:
                 yield Experiment(function.duplicate(), optim, budget=budget, num_workers=1, seed=next(seedg))
 
 
+@registry.register
+def constrainedillcondipara(seed: Optional[int] = None) -> Iterator[Experiment]:
```
brozi: constrained_ill_conditioned_params

nevergrad/benchmark/experiments.py
teytaud: Done.

nevergrad/benchmark/experiments.py
```diff
@@ -76,6 +76,26 @@ def discrete(seed: Optional[int] = None) -> Iterator[Experiment]:
                 yield Experiment(func.duplicate(), optim, budget=budget, num_workers=1, seed=next(seedg))
 
 
+@registry.register
+def chaindeceptive(seed: Optional[int] = None) -> Iterator[Experiment]:
```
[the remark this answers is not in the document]
teytaud: Done.

nevergrad/benchmark/experiments.py
teytaud: Done (with a slight variant - it's parallel instead of params, which validates your claim that I should be explicit and use long names :-)

PR #347, opened 2019-12-01
Adding an unordered discrete variable

teytaud:
## Types of changes

There was no variable for the unordered case for people who don't want to use the softmax transformation.

## Checklist

### DIFF
--- a/docs/instrumentation.md
+++ b/docs/instrumentation.md
@@ -13,6 +13,7 @@ The aim of instrumentation is to turn a piece of code with parameters you want t
 6 types of variables are currently provided:
 - `SoftmaxCategorical(items)`: converts a list of `n` (unordered) categorial items into an `n`-dimensional space. The returned element will be sampled as the softmax of the values on these dimensions. Be cautious: this process is non-deterministic and makes the function evaluation noisy.
 - `OrderedDiscrete(items)`: converts a list of (ordered) discrete items into a 1-dimensional variable. The returned value will depend on the value on this dimension: low values corresponding to first elements of the list, and high values to the last.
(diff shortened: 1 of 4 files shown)


docs/instrumentation.md
```diff
@@ -13,6 +13,7 @@ The aim of instrumentation is to turn a piece of code with parameters you want t
 6 types of variables are currently provided:
 - `SoftmaxCategorical(items)`: converts a list of `n` (unordered) categorial items into an `n`-dimensional space. The returned element will be sampled as the softmax of the values on these dimensions. Be cautious: this process is non-deterministic and makes the function evaluation noisy.
 - `OrderedDiscrete(items)`: converts a list of (ordered) discrete items into a 1-dimensional variable. The returned value will depend on the value on this dimension: low values corresponding to first elements of the list, and high values to the last.
+- `UnorderedDiscrete(items)`: converts a list of (ordered) discrete items into a 1-dimensional variable. The returned value will depend on the value on this dimension: low values corresponding to first elements of the list, and high values to the last. The behavior is exactly the same as OrderedDiscrete variables -- the optimizer is just informed, by the instrumentation, that the order should not be used.
```
brozi: I don't get it. How is the optimizer informed that the order should not be used ? Don't we need to create a much higher dimension variable if the input is not ordered ? With this name, most people would assume that this is the the variable they should use when they have categorical variables. Is that the case ?

teytaud: The optimizer has a "instrumentation", in which it can read all information about variables.
In a further PR, we use it for guessing the arity of a categorical variable.

Categorical variables can be ordered or not ordered, yeah maybe most people use "categorical" for unordered, but I think it's good to have both names for making the difference explicit.

PR #367, opened 2019-12-16
add a bit for noise in instrumentation

teytaud:
## Types of changes

noise is one of the few things we usually know a priori in an objective function.
this change is critical for competence maps

CI

## Checklist

### DIFF
--- a/nevergrad/instrumentation/multivariables.py
+++ b/nevergrad/instrumentation/multivariables.py
@@ -189,7 +189,7 @@ class Instrumentation(NestedVariables):
         self.keywords: Tuple[Optional[str], ...] = ()
         self.variables: List[Variable] = []
         self._set_args_kwargs(args, kwargs)
-        self._objective_function_is_noiy: Bool = False
+        self._objective_function_is_noiy: bool = False
 
     def _set_args_kwargs(self, args: Tuple[Any, ...], kwargs: Dict[str, Any]) -> None:
         super()._set_args_kwargs(args, kwargs)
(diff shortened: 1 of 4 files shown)


nevergrad/instrumentation/multivariables.py
```diff
@@ -189,6 +189,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.keywords: Tuple[Optional[str], ...] = ()
         self.variables: List[Variable] = []
         self._set_args_kwargs(args, kwargs)
+        self._is_noiy: Bool = false
```
jrapin: `False` not `false`  :p 
What is the aim though?
- notifying that the instrumentation is noisy? This already exists (there's a "noisy" attribute in instrumentations, which will be renamed in the next version btw)
- notifying that the function is noisy? where will it be used? If so, I think this is somehow related to the `UnorderedDiscrete` variable I would say, in the next version of instrumentation I'm trying to merge all these into some kind of "descriptor" class giving information to the function.

teytaud: This is about the fact that the objective function is noisy.
But this is not related to unordered discrete --- an objective function can be ordered and noise-free, unordered and noisy, or the domain can be continuous and the objective function can be noisy, etc.

teytaud: in the code: "noisy=any(v.noisy for v in self.variables)  so this refers to noise induced by the instrumentation, not to noise in the objective function.

PR #379, opened 2019-12-17
[PR on instrumentation_v3] Convert all subparameters into Parameter

jrapin:
Convert all non-parameters to constants

### DIFF
--- a/nevergrad/parametrization/core.py
+++ b/nevergrad/parametrization/core.py
@@ -357,6 +357,12 @@ class Constant(Parameter):
     def _get_name(self) -> str:
         return str(self._value)
 
+    def get_value_hash(self) -> t.Hashable:
+        try:
+            return super().get_value_hash()
+        except NotSupportedError:
+            return "#non-hashable-constant#"
+
     @property
     def value(self) -> t.Any:
         return self._value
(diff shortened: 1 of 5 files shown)


jrapin:
this removes 8 `isisntance`

jrapin:
@danthe3rd do you prefer it this way?

nevergrad/parametrization/core.py
```diff
@@ -444,10 +447,10 @@ def value(self, value: t.Dict[str, t.Any]) -> None:
             as_parameter(self._parameters[key]).value = val
 
     def get_value_hash(self) -> t.Hashable:
-        return tuple(sorted((x, y.get_value_hash()) for x, y in self._parameters.items() if isinstance(y, Parameter)))
+        return tuple(sorted((x, y.get_value_hash()) for x, y in self._parameters.items() if not isinstance(y, Constant)))
```
danthe3rd: I understand you might not need to hash the value of constants, but maybe the user needs? If he has 2 Parameters with different constants inside? (PLUS we remove one `isinstance` call :D)

jrapin: Could be useful indeed, however that won't remove an isinstance :p We can only hash what is hashable, so we'll need to check it is...

jrapin: Not an isinstance actually then, but an additional try except and a placeholder if it's not easily hashable, and that works ;)

PR #391, opened 2019-12-19
Plugging new parametrization inside instrumentation and optimization

jrapin:
## Types of changes

FYI: this is expected to land after January 8th

This plugs parametrization as created in #323 
To do so, the new `Instrumentation` is extended with old `Variable` methods, and the old `Variable` inherits from `Instrumentation`  and overrides the new methods.  

This is not expected to change anything for you, but some printed information may change, and deprecation warnings will appear for some methods (Eg.: `set_cheap_constraint_checker` -> `register_cheap_constraint` with a slightly different API). Some probably unused functions will disappear (`get_description`) making the changes breaking anyway. At this point you will be able to use new `Parameter` classes, but their initialization may be inaccurate so it will not be advised in some cases (you would get a user warning). Also, because of the compatibility, we expect optimizers to be about 25% slower in this version. The optimizer will still only take an `Instrumentation` (new version) as input for its initialization.

## Checklist


docs/parametrization.md
```diff
@@ -0,0 +1,100 @@
+# Parametrization
+
+**Please note that parametrization is still a work in progress with heavy changes comming soon! We are trying to update it to make it simpler and simpler to use (all feedbacks are welcome ;) ), with the side effect that there will be breaking changes.**
+
+The aim of parametrization is to specify what are the parameters that the optimization should be performed upon.
+The parametrization subpackage will help you do thanks to:
+- the `parameter` modules providing classes that should be used to specify each parameter.
+- the `Instrumentation`, and `InstrumentedFunction` classes which provide an interface for converting any arguments into the data space used for optimization, and convert from data space back to the arguments space.
```
jrapin: TODO remove this

jrapin:
For the record:
Locally `pytest nevergrad/optimization` has gone up from 45s to 70s. Next PR will hopefully revert the change, which is due to the need of keeping both compatibilities.

PR #397, opened 2019-12-20
better naming

teytaud:
## Types of changes

requested by the big boss

## Checklist

### DIFF
--- a/nevergrad/benchmark/xpbase.py
+++ b/nevergrad/benchmark/xpbase.py
@@ -147,12 +147,12 @@ class Experiment:
         assert isinstance(function, instru.InstrumentedFunction), ("All experiment functions should derive from InstrumentedFunction")
         self.function = function
         # Conjecture on the noise level.
-        if not self.function.instrumentation.probably_noisy:
+        if not self.function.instrumentation.handled_as_noisy:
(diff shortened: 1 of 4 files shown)


nevergrad/benchmark/xpbase.py
```diff
@@ -147,12 +147,12 @@ def __init__(self, function: instru.InstrumentedFunction,
         assert isinstance(function, instru.InstrumentedFunction), ("All experiment functions should derive from InstrumentedFunction")
         self.function = function
         # Conjecture on the noise level.
-        if not self.function.instrumentation.probably_noisy:
+        if not self.function.instrumentation.handled_as_noisy:
```
jrapin: I don't really like this term either, because "handled_as_noisy" assumes the it **will** be handles as noisy. While it's for the algorithm to decide.
In practice for the the new instrumentation I chose to use "deterministic" instead of "noisy" (cos its not quite noise either). I would therefore go for a flag in the instrumentation called deterministic_function, which defaults to True and which can be set to False manually.
In any case, you can leave "probably noisy" for now, and I'll do the modifications when we agree on a name, because I will not change only the name, but also how to set it, and how to access it (I'm grouping all these information into a "descriptors" attribute to make it more structured, see #323 )

PR #412, opened 2019-12-26
Cma can work as oneshot

teytaud:
## Types of changes

## Checklist

- [ x] I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).
- [ x] All tests passed, and additional code has been covered with new tests.


jrapin:
@teytaud in my understanding the `one_shot` tag was for algorithms that would not make use of any previous `tell` for `ask`ing a new point, which is not true for CMA, so why do you want to consider it as such? You can always use all `ask` before doing any `tell` for all algotihms, but if I'm correct that only sums up to evaluating the initialization method

PR #467, opened 2020-01-19
Use standard initialization for PSO instead of full real space sampling

### LINKED ISSUE #296: PSO has unusually high population std
Consider the following code: 

```
import nevergrad as ng
import numpy as np
def square(x, y=12):
    return sum((x - .5)**2) + abs(y)
popsize = 100
episodes = 500
budget = popsize * episodes
op = ng.optimizers.registry['PSO'](
      instrumentation = 200,
      num_workers = popsize,
      budget = budget)
for ep in range(budget // popsize):
  solutions = []
  candidates = []
  for i in range(popsize):
    candidate = op.ask()
    candidates.append(candidate)
    solutions.append(candidate.args[0])
  solutions = np.asarray(solutions)
  candidates = np.asarray(candidates)
  fitness = []
  for i in range(popsize):
    value = square(solutions[i])
    fitness.append(value)
    op.tell(candidates[i], value)
  if ep % 10 == 0:
    print(np.std(solutions))
recommendation = op.provide_recommendation()
```
The standard deviation of the population here is unusually high. For other optimizers (e.g. DE, TBPSA, etc) the std seems normal. Is this behaviour expected?

jrapin:
## Types of changes

`PSO` initialization was very different from other algorithm: sampling in `[0,1]` and transforming to the real space with an arctan, instead of using a normal distribution. With this diff, `PSO` uses the sampling of the parametrization, which by default is a normal distribution, like other algorithms.
Closes #296 

## Checklist


teytaud:
PSO was performing great on deceptive, the question is whether the new version will be great as well; I can check this for you maybe ?

jrapin:
> PSO was performing great on deceptive, the question is whether the new version will be great as well; I can check this for you maybe ?

Yes it's important to check, I expect PSO will not be as good: if it's good currently, it may have been because the initialization was completely different to other algorithms', so not a fair comparison

teytaud:
As I'm underwater I propose that we merge and we'll see in the next global runs if PSO is still dominant in categories of problems for which it is dominant - or we really wait a little bit, I have papers to finish :-)

Regarding fair comparisons - if results with the new initialization are worse, then maybe we must keep a version with the old initialization somewhere.

jrapin:
@teytaud now:
- `PSO` uses standard initialization
- `WidePSO` uses the old initialization
- all `chainPSO` use the standard `PSO` (with new initialization)

Should I modify all `chainPSO` to use the "old" initialization with `WidePSO`?

jrapin:
> Should I modify all `chainPSO` to use the "old" initialization with `WidePSO`?

approved as it is by Olivier

PR #476, opened 2020-01-27
faster test

teytaud:
I'm pretty sure that @jrapin is surprised that I remember that I've promised to make this test faster.

## Types of changes

## Checklist


jrapin:
> I'm pretty sure that @jrapin is surprised that I remember that I've promised to make this test faster.

I did not even remember asking, so I'm surprised indeed :D 
Seems like a good idea indeed. Then maybe one additional thought (which is orthogonal to the PR): if the aim is to make sure it's random, should we make sure it's not always a tie?

PR #534, opened 2020-02-21
Bug in photonics.chirped fixed.

AnMoreau:
## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

The chirped function actually worked as expected (and described in https://arxiv.org/abs/1808.04689) only in some specific cases. The bug was tricky to find (the tests done
earlier could not allow to spot it, but the results of the optimization were not satisfactory).

We have run a DE on this problem (among the most efficient methods for this given testbed)
and found satisfactory results which correspond to naturally occurring structures/artificially engineered ones. The cost function is thus operational. 

## Checklist

- [X ] The documentation is up-to-date with the changes I made.
- [ ] I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).

### DIFF
--- a/nevergrad/functions/photonics/photonics.py
+++ b/nevergrad/functions/photonics/photonics.py
@@ -27,7 +27,7 @@ def bragg(X: np.ndarray) -> float:
     n = np.concatenate(([1], np.sqrt(X[0:bar]), [1.7320508075688772]))
     Type = np.arange(0, bar + 2)
     hauteur = np.concatenate(([0], X[bar : 2 * bar], [0]))
-    tmp = np.tan(2 * np.pi * n[Type] * hauteur[Type] / lam)
+    tmp = np.tan(2 * np.pi * n[Type] * hauteur / lam)
     # Specific to this substrate.
     Z = n[-1]
     for k in range(np.size(Type) - 1, 0, -1):


nevergrad/functions/photonics/photonics.py
```diff
@@ -27,7 +27,7 @@ def bragg(X: np.ndarray) -> float:
     n = np.concatenate(([1], np.sqrt(X[0:bar]), [1.7320508075688772]))
     Type = np.arange(0, bar + 2)
```
jrapin: to help have uniform and readable code, in the future I would suggest you avoid using capitalized names for instances like `Type` (cf [pep8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)). In this case naming it `type` would actually be worth though, since this is a reserved name :D

nevergrad/functions/photonics/photonics.py
```diff
@@ -27,7 +27,7 @@ def bragg(X: np.ndarray) -> float:
...
     hauteur = np.concatenate(([0], X[bar : 2 * bar], [0]))
-    tmp = np.tan(2 * np.pi * n[Type] * hauteur[Type] / lam)
+    tmp = np.tan(2 * np.pi * n[Type] * hauteur / lam)
     # Specific to this substrate.
     Z = n[-1]
     for k in range(np.size(Type) - 1, 0, -1):
```
jrapin: FYI `np.size(Type)` is equivalent to `Type.size` (shorter^^)

jrapin:
Only thing left before I can merge though is that you need to sign the CLA (see automatic comment above). This just means you comply with the license in the repository, i.e. MIT license which is really open

nevergrad/functions/photonics/photonics.py
AnMoreau: How do you guess I'm not yet completely familiar with Python :) ?

PR #536, opened 2020-02-21
Rename a_min and a_max to lower and upper and add them to Scalar init

jrapin:
## Types of changes

## Checklist

### DIFF
--- a/nevergrad/parametrization/data.py
+++ b/nevergrad/parametrization/data.py
@@ -342,7 +342,7 @@ class Scalar(Array):
     Parameters
     ----------
     init: optional float
-        initial value of the scalar
+        initial value of the scalar (defaults to 0.0 if both bounds are not provided)
     lower: optional float
         minimum value if any
     upper: optional float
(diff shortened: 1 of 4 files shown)


nevergrad/parametrization/data.py
```diff
@@ -332,19 +341,39 @@ class Scalar(Array):
...
+        upper: tp.Optional[float] = None,
+        mutable_sigma: bool = True
+    ) -> None:
+        bounded = all(a is not None for a in (lower, upper))
+        if bounded:
+            if init is None:
+                init = (lower + upper) / 2.0
         super().__init__(init=np.array([init]), mutable_sigma=mutable_sigma)
+        if bounded:
+            self.set_mutation(sigma=(upper - lower) / 6)  # type: ignore
```
jrapin: @teytaud 
what do you think of these defaults for scalars:
if bound lower and upper bound are provided:
- sigma is set to a sixth of b - a (so that the full range is covered from start)
- we also set `full_range_sampling=True` which means that parametrization.sample() will sample uniformly within the bounds ( this is used by some new algorithms)
- if init is not provided, it is set it is set to the middle point

jrapin: If neither or only one is provided, we are back to normal and the user need to use the Scalar methods to update sigma

jrapin: Also forgot to mention, by default it's using clipping, but once again to use a different method one can set it manually through the set_bounds method

PR #547, opened 2020-02-26
Simplify crossover

jrapin:
## Types of changes

This reimplements crossovers with a simplified API and scope.
We now only deal with 2 points crossover of 2 individuals, in any dimension, while we could before perform k-points crossover of n individuals but only in 1 dimension.
The code is simpler as well, and there is no clear impact of integrating other cases (more than 2 individuals and more points).

## Checklist

### DIFF
--- a/nevergrad/functions/arcoating/test_core.py
+++ b/nevergrad/functions/arcoating/test_core.py
@@ -34,7 +34,7 @@ def test_arcoating_recombination() -> None:
         arrays.append(func.parametrization.spawn_child())  # type: ignore
         arrays[-1].value = num * np.ones(arrays[0].value.shape)
     arrays[0].recombine(arrays[1])
-    expected = [[3., 3., 5., 5., 5., 3.]]
+    expected = [3., 3., 3., 5., 3., 3.]
     np.testing.assert_array_equal(arrays[0].value, expected)
 
 
--- a/nevergrad/parametrization/utils.py
+++ b/nevergrad/parametrization/utils.py
@@ -170,33 +170,34 @@ class Crossover:
     """ Experimental, the API will evolve
     """
 
-    def __init__(self, num_points: int = 0, structured_dimensions: tp.Iterable[int] = ()) -> None:
-        self.num_points = num_points
-        self.structured_dimensions = sorted(structured_dimensions)
+    def __init__(self, axis: tp.Optional[tp.Union[int, tp.Iterable[int]]]) -> None:
+        self.axis = (axis,) if isinstance(axis, int) else axis
 
     def __repr__(self) -> str:
-        return f"{self.__class__.__name__}({self.num_points}, {self.structured_dimensions})"
+        return f"{self.__class__.__name__}({self.axis})"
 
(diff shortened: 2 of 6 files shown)


nevergrad/parametrization/utils.py
```diff
@@ -152,51 +151,69 @@ def __call__(self, *args: tp.Any, **kwargs: tp.Any) -> str:
...
+    Examples:
+    ---------
+    - 2-dimensional array, with crossover on dimension 1:
+      0 1 0 0
+      0 1 0 0
+      0 1 0 0
+    - 2-dimensional array, with crossover on dimensions 0 and 1:
+      0 0 0 0
+      0 1 1 0
+      0 1 1 0
```
jrapin: @teytaud is that what you need? ;)

nevergrad/parametrization/utils.py
```diff
@@ -152,51 +151,69 @@ def __call__(self, *args: tp.Any, **kwargs: tp.Any) -> str:
...
+    """Operator for merging part of an array into another one
+
+    Parameters
+    ----------
+    axis: None or int or tuple of ints
+        the axis (or axes) on which the merge will happen. This axis will be split into 3 parts: the first and last one will take
+        value from the first array, the middle one from the second array.
+    max_size: None or int
+        maximum size of the part taken from the second array. By default, this is at most around half the number of total elements of the
+        array to the power of 1/number of axis.
```
jrapin: @teytaud do you agree this is a sensible default?

nevergrad/functions/arcoating/test_core.py
```diff
@@ -34,7 +34,7 @@ def test_arcoating_recombination() -> None:
         arrays.append(func.parametrization.spawn_child())  # type: ignore
         arrays[-1].value = num * np.ones(arrays[0].value.shape)
     arrays[0].recombine(arrays[1])
-    expected = [[3., 3., 5., 5., 5., 3.]]
+    expected = [3., 3., 3., 5., 3., 3.]
```
teytaud: can you tell me a bit more how this expected is built ? this is a test of seedability ?

nevergrad/parametrization/utils.py
teytaud: YES :-)

nevergrad/parametrization/utils.py
teytaud: i have abolutely no idea... the most classical thing, for historical reasons, is to use k1 and k2 random uniform and split accordingly, but some recent excellent papers have shown that small values lead to excellent results on images.

anyway don't stop for that, we'll have to play with those values later...

nevergrad/functions/arcoating/test_core.py
jrapin: Exactly, this both:
- tests that it is seedable (seed is set to 24 a few lines above)
- that we obtain something expected (here all the 3 come from the first array, and the 5 from the second), so we indeed have a 2 points crossover.

The value is different from before because I changed the implementation.

PR #551, opened 2020-02-26
Emna

fteytaud:
## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

- First PR about EMNA based algorithms. To be added in the future to have a complete set of algorithms:
     - reweighting emna
     - tbpsa with emna (parallel version of tbpsa)

- I have added a first version of a harder parallel benchmark.

pytest nevergrad/optimization/test_optimizerlib.py

I have deactivated the check_optimizer test (for emna algorithms) as this kind of algorithms needs a num_workers >> 1

## Checklist

- [ X] The documentation is up-to-date with the changes I made.
- [X ] I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).


nevergrad/optimization/optimizerlib.py
```diff
@@ -1425,32 +1425,126 @@ def _internal_tell_not_asked(self, candidate: p.Parameter, value: float) -> None
...
+@registry.register
+class AnisoEMNA(EMNA):
+    """Simple Anisotropic Estimation of Multivariate Normal Algorithm (EMNA).
+    """
+
+    # pylint: disable=too-many-instance-attributes
+
+    def __init__(self, parametrization: IntOrParameter, budget: Optional[int] = None, num_workers: int = 1) -> None:
+        super().__init__(parametrization, budget=budget, num_workers=num_workers)
+        self.sigma = np.ones(self.dimension)
```
jrapin: the typing issue is here because you're setting a np.ndarray to `sigma` which is a float in the parent class.
You'd need to type it to `sigma: tp.Union[float, np.ndarray]` in the **parent** class

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -129,6 +129,7 @@ def test_optimizers(name: str) -> None:
     if isinstance(optimizer_cls, base.ConfiguredOptimizer):
         assert any(hasattr(mod, name) for mod in (optlib, xpvariants))  # make sure registration matches name in optlib/xpvariants
     verify = not optimizer_cls.one_shot and name not in SLOW and not any(x in name for x in ["BO", "Discrete"])
+    print("Alors : {} et {}".format(name, verify))
```
jrapin: please remove this

nevergrad/optimization/optimizerlib.py
```diff
@@ -1425,32 +1425,126 @@ def _internal_tell_not_asked(self, candidate: p.Parameter, value: float) -> None
...
-class EMNA_TBPSA(TBPSA):
-    """Test-based population-size adaptation with EMNA.
+@registry.register
+class EMNA(base.Optimizer):
+    """Simple Estimation of Multivariate Normal Algorithm (EMNA).
     """
 
+    # pylint: disable=too-many-instance-attributes
+
+    def __init__(self, parametrization: IntOrParameter, budget: Optional[int] = None, num_workers: int = 1) -> None:
```
jrapin: Would it be possible to have additional inputs `naive`  and `anisotropic` here, and then use these options in the rest of the code?
The aim is to do something like the implementation of OnePlusOne here which is a good example of what we are aiming for (TBPSA is not a good example, I need to update it):
https://github.com/facebookresearch/nevergrad/blob/b8eaf6221ec8fc2985b7363ed39f7d2943a3fc2a/nevergrad/optimization/optimizerlib.py#L38-L160

That would avoid code duplication, and make it way easier to iterate upon on the long range.

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -70,7 +70,7 @@ def check_optimizer(
             warnings.filterwarnings("ignore", category=FinishedUnderlyingOptimizerWarning)
             # now optimize :)
             candidate = optimizer.minimize(fitness)
-        if verify_value and "chain" not in str(optimizer_cls):
+        if verify_value and "chain" not in str(optimizer_cls) and "EMNA" not in str(optimizer_cls):
```
jrapin: > I have deactivated the check_optimizer test (for emna algorithms) as this kind of algorithms needs a num_workers >> 1

Can you elaborate on this? I don't understand how that is a problem. If num_workers is 1 the algorithms can always mock that the `ask` are played in parallel.

nevergrad/optimization/test_optimizerlib.py
fteytaud: Oups sorry ; I missed that during my cleaning!

nevergrad/optimization/test_optimizerlib.py
fteytaud: If I understood the test correctly, the result is that the population size of emna (with such a parametrization) will be be too small ; and sigma is computed as the standard deviation of the mu (roughly lambda/4) best parents. Maybe I should add a warning in the algorithm to prevent its use with a too small popsize.

nevergrad/optimization/optimizerlib.py
fteytaud: Ok thx for all comments ; I try to correct it quickly, some stuff are easy to change ; the correction of code duplication is a good idea, I will look into that (it is not intuitive right now, as I haven't looked deeply enough in the parametrization but I will work on it :).

nevergrad/optimization/test_optimizerlib.py
jrapin:
```python
self.llambda = 4 * self.dimension
if num_workers is not None:
    self.llambda = max(self.llambda, num_workers)
```
This sets the population size to **at least** `num_workers` so `num_workers=1` should not matter

nevergrad/optimization/optimizerlib.py
jrapin: Let me know if you need help, at some point I can make a PR on top of your PR to suggest changes if you need

jrapin: That's great!
There's a second step to make it work:
You'll need to add a `ConfiguredOptimizer` and register variants like it's done here for OnePlusOne:
https://github.com/facebookresearch/nevergrad/blob/e07ecefe0c3a7fdffe72abdef4eb0abf7ea7da11/nevergrad/optimization/optimizerlib.py#L117-L166

For you that would mean this:
```python
class ConfiguredEMNA(base.ConfiguredOptimizer):
    """TODO
    """

    # pylint: disable=unused-argument
    def __init__(
        self,
        *,
            isotropic: bool = True,
            naive: bool = True
    ) -> None:
        super().__init__(EMNA, locals())
```

Actually if you are OK, I would call this new class `EMNA`, and the one you did `_EMNA` to hide it since the API that we need to document is the one in the new class.
The syntax is then for instance `EMNA(isotropic=False)(parametrization, budget)` to instantiate an optimizer.
You can preregister variants, like:
```python
NaiveIsoEMNA = EMNA().set_name("NaiveIsoEMNA", register=True)
NaiveAnisoEMNA = EMNA(isotropic=False).set_name("NaiveAnisoEMNA", register=True)
IsoEMNA = EMNA(naive=False).set_name("IsoEMNA", register=True)
AnisoEMNA = EMNA(inaive=False, sotropic=False).set_name("AnisoEMNA", register=True)
```

Then a few things to consider:
- please add a docstring in the new class, it will be automatically visible in the documentation when it's merged. What could be of interest for users is when this algorithm should be used (which cases will it be good at? bad at?) and what is the impact of each parameter
- add the new configured optimizer to optimization/families.py
- I'm trying to limit the number of variants in optimizerlib (work in progress...), to keep only the most useful variants for user, if you can leave only one or two there it would be great, the others should be added to optimization/experimentalvariants.py so that they can be straightforwardly used in the benchmarks.

fteytaud: Hello, thx a lot for your remarks. I think I have taken into account everything, but do not hesitate to tell me if something is not correct.

nevergrad/optimization/test_optimizerlib.py
jrapin: Can it be removed or the test is failing for some variants? if so which one? We'd need to figure out why, because num_workers=1 shouldnt be a problem

nevergrad/optimization/optimizerlib.py
```diff
@@ -1425,33 +1425,105 @@ def _internal_tell_not_asked(self, candidate: p.Parameter, value: float) -> None
...
+
+        Parameters
+        ----------
+        - `"isotropic"`: if true: isotropic version on EMNA, i.e. we have an
+                                    identity matrix for the Gaussian
+                         else: anisotropic version, we here consider the
+                                    separable version, meaning we have a diagonal
+                                    matrix for the Gaussian
+        - `"naive"`: optimistic version. This parameter can be set to true for
+                    noiseless problems
```
jrapin: A few details for the documentation to be able to be correctly generated:
- first the parameters should be presented slightly differently
```suggestion
        isotropic: bool
             isotropic version on EMNA if True, i.e. we have an
             identity matrix for the Gaussian, else  we here consider the separable
             version, meaning we have a diagonal matrix for the Gaussian (anisotropic)
        naive: bool
             set to False for noisy problem, so that the best points will be an average of the final
             population.
```
- then notice (if that makes sense), I changed the way you describe "naive", since by default it is True
- finally, thee whole docstring (except the first line, is overindented: all lines should be aligned with the initial `"""`

nevergrad/optimization/recorded_recommendations.csv
```diff
@@ -84,6 +86,8 @@ MultiCMA,1.012515477,-0.9138805701,-1.029555946,1.2098418178,,,,,,,,,,,,
 MultiScaleCMA,0.0002149759,-0.0003843636,-0.0002539104,7.32548e-05,,,,,,,,,,,,
 MutDE,1.1400386808,0.3380024444,0.4755144618,2.6390460807,0.6911075733,1.111235567,-0.2576843178,-1.1959512855,,,,,,,,
 NGO,0.0,-0.3451057176,-0.1327329683,1.9291307781,,,,,,,,,,,,
+NaiveAnisoEMNA,1.012515477,-0.9138691467,-1.0295302074,1.2097964496,,,,,,,,,,,,
+NaiveEMNA,1.012515477,-0.9138691467,-1.0295302074,1.2097964496,,,,,,,,,,,,
```
jrapin: Please merge master and then rerun pytest to record the new values. Since names have changed this should be updated. Hopefully that should solve the continuous integration issues

nevergrad/optimization/optimizerlib.py
fteytaud: Ok thx ! I think it is ok now.
Just the test deactivation remains, but I am not sure on how to correct it. It seems for me that the test will fail for all variants. I have just tried to put algorithms in the "SLOW" list into this test and they fail too, so I don't know. From here, as I am not sure to understand why I would be tempted to add EMNA in the SLOW list, in that way it will skip the test, but I think you will disagree :-)

nevergrad/optimization/recorded_recommendations.csv
```diff
@@ -28,7 +29,7 @@ AvgSmallHaltonSearchPlusMiddlePoint,0.0031863936,0.0076470967,-0.0175068607,0.00
 AvgSmallHammersleySearchPlusMiddlePoint,0.0052440051,-0.0115034938,-0.001397103,0.0084162123,,,,,,,,,,,,
 AvgSmallScrHaltonSearchPlusMiddlePoint,-0.0115034938,0.0122064035,-0.0084162123,0.0106757052,,,,,,,,,,,,
 AvgSmallScrHammersleySearchPlusMiddlePoint,-0.0128155157,0.0,0.004307273,0.0084162123,,,,,,,,,,,,
-BO,-1.7186359043,-0.0719657219,-0.2263452946,-0.0245468071,,,,,,,,,,,,
+BO,-1.1230418754,-0.563634709,2.3536712172,5.1757172963,,,,,,,,,,,,
```
jrapin: There was an update in the bayesian optimization package, and you have been running the old one while the CI uses the new one.
To correct this:
- `pip install -U bayesian-optimization` 
- remove all lines containing BO in this file
- rerun pytest
- commit
Hopefully, all will be good then!

fteytaud: Thx, all tests on CI are ok now.

jrapin:
I'm still concerned by the line in test_optimizerlib.py, what happens if you remove it? apart from that everything seems good

fteytaud:
> I'm still concerned by the line in test_optimizerlib.py, what happens if you remove it? apart from that everything seems good

The "np.testing.assert_array_almost_equal" will fail. I am looking more deeply into this now.

fteytaud:
Ok I have added and corrected a bug that I had introduced in the aniso implementation and now all tests are ok. I add a warning too in the algorithm, to advertise the use with a too small population size.

jrapin:
Thank you, this looks perfect!
Just one final step: can you merge master (or maybe just update the BO recorded recommendations with the new bayes opt version?), there seems to be a couple of changes in the recorded recommendantion file which should not be there.
When it's done, I'll merge ;)

fteytaud:
Ok I have merged and lauched pytest, I hope this is ok.

jrapin:
Seems good, I'll merge this!
Thank you!!

PR #554, opened 2020-02-27
remove dirty experiments, simplify others

teytaud:
## Types of changes

## Checklist


nevergrad/benchmark/experiments.py
```diff
@@ -95,57 +95,23 @@ def yawidebbob(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
-    ]
-    for func in functions:
-        for optim in optims:
-            for budget in [100, 200, 300, 400, 500, 600, 700, 800, 900, 1000, 1100, 1200, 1300, 1400, 1500, 1600, 1700,
-                           1800, 1900, 2000, 2100, 2200, 2300, 2400, 2500, 2600, 2700, 2800, 2900, 3000]:  # , 10000]:
-                yield Experiment(func, optim, budget=budget, num_workers=1, seed=next(seedg))
+    for nv in [10, 50, 200]:
+        for arity in [2, 7]:
+            variables = list(ng.p.TransitionChoice(list(range(arity))) for _ in range(nv))
+            instrum = ng.p.Instrumentation(ng.p.Tuple(*variables))
```
jrapin: I've had to add a Tuple layer here because onemax function and others need to have a tuple as single input, while using an instrumentation directly would give it several positional arguments instead

**Caution**: Since I needed to do it here, I'm a bit concerned that yawidebbob does not work! We'll need to investigate this :s

nevergrad/functions/base.py
```diff
@@ -117,8 +117,8 @@ def copy(self: EF) -> EF:
                                                   f"{output} with descriptors {output._descriptors}.")
         else:
             # back to standard ExperimentFunction
-            ouptut = self.__class__(self.function, self.parametrization.copy())
-            ouptut._descriptors = self.descriptors
+            output = self.__class__(self.function, self.parametrization.copy())
```
jrapin: There was a stupid typo of mine here! which caused the bug you had

jrapin:
So, there is still some reproducibility problem I did not expect, I'll have to check that in more details tomorrow

nevergrad/benchmark/experiments.py
```diff
@@ -72,8 +72,7 @@ def yawidebbob(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
     # Discrete, unordered.
     for nv in [10, 50, 200]:
         for arity in [2, 7]:
-            variables = list(ng.p.TransitionChoice(list(range(arity))) for _ in range(nv))
-            instrum = ng.p.Instrumentation(*variables)
+            instrum = ng.p.Tuple(*(ng.p.TransitionChoice(range(arity)) for _ in range(nv)))
```
jrapin: @teytaud What you wanted is to have 1 argument which was a tuple of the integer choices, so you should just build it. The `Instrumentation` class is for when you want multiple arguments (positional and/or keyword arguments). So this is not what you needed here.

nevergrad/benchmark/xpbase.py
```diff
@@ -129,6 +135,9 @@ def __init__(self, function: fbase.ExperimentFunction,
         self.result = {"loss": np.nan, "elapsed_budget": np.nan, "elapsed_time": np.nan, "error": ""}
         self.recommendation: Optional[p.Parameter] = None
         self._optimizer: Optional[obase.Optimizer] = None  # to be able to restore stopped/checkpointed optimizer
+        # make sure the random_state of the base function is created, so that spawning copy does not
+        # trigger a seed for the base function, but only for the copied function
+        self.function.parametrization.random_state  # pylint: disable=pointless-statement
```
jrapin: Here was the reproducibility issue, that was a problem of my code. There was an extra call of the default random generator to seed this parametrization after the beginning of the first experiment only, while this parametrization is not used in the experiments.

nevergrad/benchmark/experiments.py
teytaud: Ok thx!

PR #571, opened 2020-03-05
Make pruning much much less conservative

jrapin:
## Types of changes

Optimization may slow down dramatically at large number of iterations because the archive becomes big. this makes it much smaller.
The archive is therefore limited to 1000 points most of the time (more for large num_workers), which seems big enough in much cases

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
   Using old names will raise a deprecation warning for the time being.
 - Scalar now takes optional `lower` and `upper` bounds at initialization, and `sigma` (and optionnally `init`)
   if is automatically set to a sensible default [#536](https://github.com/facebookresearch/nevergrad/pull/536).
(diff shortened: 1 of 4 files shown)


jrapin:
@teytaud what do you think? I feel the archive is wasting time :s

teytaud:
> @teytaud what do you think? I feel the archive is wasting time :s

Ok my educated guess, taking into account the fact that the archive is the number of points, not the number of points with their repetitions, is something like "7 times num_workers should be enough" (or 7 times the population size); and possibly much more than that when we apply something like "clearing"...

but I might be completely wrong :-)

jrapin:
> > @teytaud what do you think? I feel the archive is wasting time :s
> 
> Ok my educated guess, taking into account the fact that the archive is the number of points, not the number of points with their repetitions, is something like "7 times num_workers should be enough" (or 7 times the population size); and possibly much more than that when we apply something like "clearing"...
> 
> but I might be completely wrong :-)

If that fits 99% if use case then I go for it. This is only a default value, which can be modified by the optimizer if need be.

PR #601, opened 2020-04-03
Implement automatic deployement

jrapin:
## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,14 +105,14 @@ jobs:
             . venv/bin/activate
             rm -rf dist  # make sure it's clean
             python setup.py sdist bdist_wheel
+            mkdir nevergrad-build
+            cp -r dist nevergrad-build/
             python -c "from pathlib import Path;files = Path('nevergrad.egg-info/SOURCES.txt').read_text().split(); assert 'LICENSE' in files"
             python3 -m venv test_wheel
             . test_wheel/bin/activate
             cd ..  # don't use nevergrad locally
             pip install repo/dist/nevergrad-*any.whl
             python -c "from nevergrad import functions;f = functions.ArtificialFunction(name='sphere', block_dimension=2);f([2, 3])"
-            mkdir nevergrad-build
-            mv dist nevergrad-build/
 
       - persist_to_workspace:
           root: nevergrad-build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -238,13 +238,11 @@ jobs:
           command: |
             echo -e "[pypi]" >> ~/.pypirc
             echo -e "username = jrapin" >> ~/.pypirc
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -103,14 +103,22 @@ jobs:
           when: always
           command: |
             . venv/bin/activate
+            rm -rf dist  # make sure it's clean
             python setup.py sdist bdist_wheel
+            mkdir nevergrad-build
+            mv dist nevergrad-build/
```
danthe3rd, line 109: (btw if you copy-pasted the script from HiPlot: I used a different `dist` folder because TypeScript will generate JS files in `dist` as well, but it might not be necessary here)

.circleci/config.yml
```diff
@@ -233,3 +269,12 @@ workflows:
...
               only: master
+      - pypi-deploy:
+          requires:
+            - install
+            - mypy
+            - pytests
+            - windows-pytests
+          filters:
+            tags:
+              only: /(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(\.post(0|[1-9][0-9]*))?/
```
danthe3rd, line 280: It is possible that you need to add the `filters/tags` to all CI jobs as well. IIRC CircleCI is not smart enough to run dependencies (for instance `install`) when a child job (`pypi-deploy`) has to run

.circleci/config.yml
```diff
@@ -218,6 +223,37 @@ jobs:
...
+
+      - run:
+          name: Verify tag
+          command: |
+            python setup.py verify
+
+      - run:
+          name: upload to pypi
+          command: |
+            python3 -m venv_deploy venv
```
danthe3rd, line 251: Did you mean:
```suggestion
            python3 -m venv venv_deploy
```

.circleci/config.yml
jrapin: I considered removing it, but I needed a root (nevergrad-build) and a path (dist), so this was the simplest way I found anyway

.circleci/config.yml
jrapin: Yes I figured as well indeed :D thanks ;)

.circleci/config.yml
jrapin: Oh? That would be a pain :s 
I'll try to find a way without this :s but I suppose you're right :(

jrapin: For now at least, I could get it to not run when tag is missing by adding
```
branches:
  ignore: /.*/
```
which does not make sense to me.
What your saying is that if I push a tag, only this part may run? That seems weird

jrapin: Oh the branch ignore is because by default all branches are included, and there's a OR condition so any branch OR a tagged branch would trigger it if we dont ignore all branches... makes more sense

PR #620, opened 2020-04-22
Tentatively fix EDA variants (more work to do)

jrapin:
## Types of changes

## Checklist


nevergrad/optimization/optimizerlib.py
```diff
@@ -371,22 +371,26 @@ def _internal_ask_candidate(self) -> p.Parameter:
...
-        candidate._meta["loss"] = value
         self.children.append(candidate)
         if self._POPSIZE_ADAPTATION:
             self.popsize.add_value(value)
         if len(self.children) >= self.popsize.llambda:
-            self.children = sorted(self.children, key=lambda c: c._meta["loss"])
+            self.children = sorted(self.children, key=lambda c: c.loss)
             population_data = [c.get_standardized_data(reference=self.parametrization) for c in self.children]
-            self.covariance *= 0.9 if self._COVARIANCE_MEMORY else 0
-            self.covariance += 0.1 * np.cov(np.array(population_data).T)
```
jrapin, line 382: maybe I'm wrong here, but it seems to me that without the memory, the coeff should be 1 and not 0.1.
I've added a dump factor in the fix to solve this (if it did need to be solved)

nevergrad/optimization/optimizerlib.py
```diff
@@ -371,22 +371,26 @@ def _internal_ask_candidate(self) -> p.Parameter:
...
-            # Computing the new parent
             mu = self.popsize.mu
             arrays = population_data[:mu]
+            # covariance
+            centered_arrays = np.array([x - self.current_center for x in arrays])
+            dump = 0.9 if self._COVARIANCE_MEMORY else 0
+            self.covariance *= dump
+            self.covariance += (1 - dump) * centered_arrays.T.dot(centered_arrays)
```
jrapin, line 386: this is the covariance of:
- the selected points (before: all the generated points)
- with respect to the mean used for generation (before: with respect to the mean of all the generated points)

This makes more sense to me since only with these 2 fixes does the covariance adapts to the direction that is useful. But I may be wrong here.

jrapin:
Correction, it works "most of the time" but still early converges often

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -114,17 +114,14 @@ def check_optimizer(
     "ASCMA2PDEthird",
     "MultiScaleCMA",
     "PCEDA",
```
jrapin, line 116: PCEDA still failed. The others ones may fail as well pretty often if they early converge, they may require adding again to the list of slow optimizers which convergence cannot be tested

PR #624, opened 2020-04-26
Add a first test case on images

teytaud:
just for discussion next monday :-)

## Types of changes

## Checklist


nevergrad/functions/images/core.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from math import sqrt, tan, pi
+import numpy as np
+import nevergrad as ng
+import PIL
```
jrapin, line 9: I think you can remove `import PIL` since you import `PIL.Image`

nevergrad/functions/images/core.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # Storing high level information.
+        self.domain_shape = (256, 256, 3)
+        self.problem_type = problem_type
+        self.index_pb = index_pb
+
+        # Storing data necessary for the problem at hand.
+        assert problem_type == "recovering"  # For the moment we have only this one.
+        assert index_pb == 0  # For the moment only 1 target.
+        path = os.path.dirname(__file__) + "/headrgb_olivier.png"
```
jrapin, line 34: Prefer `Path` over `os.path`. I think the following would work:
```python
from pathlib import Path
path = Path(__file__).with_name("headrgb_olivier.png")
```

nevergrad/functions/images/core.py
```diff
@@ -0,0 +1,69 @@
...
+        #     self.data = ..... (here we load the imagee correspnding to index_pb and problem_type; this is
+        #         # the attacked image.)
+
+        array = ng.p.Array(shape=self.domain_shape, mutable_sigma=True,)
+        array.set_mutation(sigma=35)
+        array.set_bounds(lower=0, upper=255, method="clipping", full_range_sampling=True)
+        max_size = ng.p.Scalar(lower=1, upper=200).set_integer_casting()
+        array.set_recombination(ng.p.mutation.Crossover(axis=(0, 1), max_size=max_size)).set_name("")  # type: ignore
+        super().__init__(self._loss, array)
+        self.register_initialization()
```
jrapin, line 49: You need to add the initialization parameter here, otherwise the benchmarks will fail

nevergrad/functions/images/core.py
```diff
@@ -0,0 +1,69 @@
...
+        self._descriptors.update()
```
jrapin, line 50: This does nothing (descriptors is a dict): You should add descriptor of the problem (problem type and index?)

nevergrad/functions/images/core.py
```diff
@@ -0,0 +1,69 @@
...
+import nevergrad as ng
+import PIL
+import PIL.Image
+import os
+from nevergrad.common.typetools import ArrayLike
+from .. import base
+
+
+class Image(base.ExperimentFunction):
+    def __init__(self, problem_type: str = "recovering", index_pb: int = 0) -> None:
```
jrapin, line 17: I'd rather use `problem_name` instead of `problem_type`: `type` seems too code-related to me
and either `problem_index` or `index`, it's weird to use a shortcut for one variable and not the other
Also, how many  indices do you plan to add? Adding a large number of files to the repo is not recommended

nevergrad/functions/images/core.py
```diff
@@ -0,0 +1,69 @@
...
+        path = os.path.dirname(__file__) + "/headrgb_olivier.png"
+        image = PIL.Image.open(path).resize((self.domain_shape[0], self.domain_shape[1]), PIL.Image.ANTIALIAS)
+        self.data = np.asarray(image)[:,:,:3]  # 4th Channel is pointless here, only 255.
+
+        # if problem_type == "adversarial": 
+        #     assert index_pb <= 100  # If we have that many target images.
+        #     self.data = ..... (here we load the imagee correspnding to index_pb and problem_type; this is
+        #         # the attacked image.)
+
+        array = ng.p.Array(shape=self.domain_shape, mutable_sigma=True,)
```
jrapin, line 43: remember this will start at 0, starting in the could be better, since at 0, half of the mutation will do nothing

nevergrad/functions/images/core.py
teytaud: Done, thx.

nevergrad/functions/images/core.py
teytaud: Done, thx.

nevergrad/functions/images/core.py
teytaud: Done.

nevergrad/functions/images/core.py
teytaud: Done.

nevergrad/functions/images/core.py
teytaud: Right, thx, done.

nevergrad/functions/images/core.py
teytaud: sorry I didn't get it.. can you develop ?

teytaud: I think I understand ("the" is for "the middle", right ?), I modify accordingly.

nevergrad/functions/images/core.py
jrapin: you probably did not push then :p

nevergrad/functions/images/core.py
```diff
@@ -0,0 +1,71 @@
...
+        # Define the loss, in case of recovering: the goal is to find the target image.
+        assert self.problem_type == "recovering"
+        assert self.index_pb == 0
+        value = np.sum(np.fabs(np.subtract(x, self.data)))
+
+        # Here we should implement "adversarial" and others.
+        return value
+
+    # pylint: disable=arguments-differ
+    def evaluation_function(self, x: np.ndarray) -> float:  # type: ignore
```
jrapin, line 69: you can remove this method, by default it does exactly that. It's only useful if you want to add specific evaluation behaviors: Eg write to the leaderboard, or run the function several times because it's noisy

nevergrad/functions/images/test_core.py
```diff
@@ -0,0 +1,18 @@
...
+from . import core
+
+
+def test_images() -> None:
+    func = core.Image()
+    x = 7 * np.fabs(np.random.normal(size=func.domain_shape))
+    #data = func.parametrization.spawn_child().set_standardized_data(x.flatten()).value
+    value = func(x)  # should not touch boundaries, so value should be < np.inf
+    print(value)
+    assert value < np.inf
```
jrapin, line 18: you should check if func.copy() works

nevergrad/functions/images/core.py
jrapin: yes exactly, sorry.
However, you should specify either `shape` or `init`, not both, and then it will work

nevergrad/functions/images/test_core.py
teytaud: Added!

nevergrad/functions/images/core.py
teytaud: Done!

nevergrad/functions/images/core.py
teytaud: yeah sorry for that...

PR #625, opened 2020-04-26
dimension vs budget bbob

teytaud:
adding a test case for studying specifically the link between budget and dimension

## Types of changes

I think our bound "2000" in Shiva is too big for switching between CMA and DE.
Maybe something like "budget < dim^2" or whatever... this testbed for investigating this in a more subtle manner.

## Checklist

### DIFF
--- a/nevergrad/benchmark/experiments.py
+++ b/nevergrad/benchmark/experiments.py
@@ -307,7 +307,7 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
     
     # Special case: for this big xp we do not want too many optims or too many test functions.
     if hd and big and not parallel:
-        optims = ["Shiva", "CMA", "ChainCMAPowell", "DE", "PSO", "OnePlusOne"]
+        optims = ["Shiva", "CMA", "chainCMAPowell", "DE", "PSO", "OnePlusOne"]
         names = ["cigar", "ellipsoid", "hm", "griewank", "sphere"]
         
     # Deceptive path is related to the sharp ridge function; there is a long path to the optimum.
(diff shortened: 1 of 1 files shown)


nevergrad/benchmark/experiments.py
```diff
@@ -298,22 +298,29 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
     if not parallel and not small:
         optims += ["Cobyla"]
     optims += ["NGO", "Shiva"]
-    # optims += [x for x, y in ng.optimizers.registry.items() if "chain" in x]
+    if hd and big and not parallel:  # We want to compare the Shiva variants.
+        optims += [x for x, y in ng.optimizers.registry.items() if "hiva" in x]
```
jrapin, line 302: are there several optimizers with "hiva" in it? It's really hard to read that way. Depending on the optimizers, there are several options to include them in a readable way

nevergrad/benchmark/experiments.py
```diff
@@ -326,6 +333,14 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
                     yield xp
 
 
+@registry.register
+def yadimbudgetbbob(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Yabbob variant aimed at exploring the link between dimension and budget specifically."""
+    internal_generator = yabbob(seed, parallel=False, big=True, hd=True)
```
jrapin, line 339: I don't think this case is interesting: this is not parallel, so Powell will win whatever happens as far as I understing.
If you turn it on, the num_workers will be 100 and it wont make sense either for budgets smaller than 500

PR #627, opened 2020-04-26
[WIP]prune descriptors

teytaud:
Nevergrad provides plots as follows:
* x-axis = budget
* y-axis = regret
We do this separately for each test case.
Unfortunately, in some cases, we prefer to have several test cases gathered on the same plot.
This is the case in particular in one-shot optimization: we do not want to separate test cases which corresponds to different parallelisms, because the parallelism is equal to the budget.
So, we remove descriptors which are uniquely determined by the budget.
This should remove num_workers from the descriptors, when num_workers = budget.

## Types of changes

in some sense a bugfix, as the result was technically correct but not consistent with what users want.

The one-shot setting is the key motivation.

Just the CI.

## Checklist

### DIFF
--- a/nevergrad/benchmark/plotting.py
+++ b/nevergrad/benchmark/plotting.py
@@ -188,7 +188,7 @@ def create_plots(
     for d in all_descriptors:
         acceptable = False
         for b in df.budget.unique():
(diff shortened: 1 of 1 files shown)


teytaud:
tested, it works, cool :-)

PR #635, opened 2020-05-03
check suggest

teytaud:
## Types of changes

Checking if suggest is not ignored.

CI

## Checklist

### DIFF
--- a/nevergrad/optimization/test_optimizerlib.py
+++ b/nevergrad/optimization/test_optimizerlib.py
@@ -166,7 +166,7 @@ def test_optimizers_suggest(name: str) -> None:  # pylint: disable=redefined-out
             optimizer.tell(candidate, 12)
             # The optimizer should recommend its suggestion, except for a few optimization methods:
             if name not in ["SPSA", "TBPSA", "StupidRandom"]:
-                np.testing.assert_array_almost_equal(optimizer.provide_recommendation().args[0], [12.0] * 4)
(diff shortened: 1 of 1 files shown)


nevergrad/optimization/test_optimizerlib.py
```diff
@@ -164,6 +164,9 @@ def test_optimizers_suggest(name: str) -> None:  # pylint: disable=redefined-out
         candidate = optimizer.ask()
         try:
             optimizer.tell(candidate, 12)
+            # The optimizer should recommend its suggestion, except for a few optimization methods:
```
jrapin, line 167: `suggest` is something completely different, so the comment is misleading, what do you want to try here?

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -164,6 +164,9 @@ def test_optimizers_suggest(name: str) -> None:  # pylint: disable=redefined-out
...
+            if name not in ["SPSA", "TBPSA", "StupidRandom"]:
+                np.testing.assert_array_almost_equal(optimizer.provide_recommendation().args[0], [12.0] * 4)
```
jrapin, line 169: you should use `parameter.value` here instead of `args[0]` which is not as clear

nevergrad/optimization/test_optimizerlib.py
jrapin: it's suggest indeed, didnt get it sorry

PR #638, opened 2020-05-03
Tentative adaptation of Shiwa to high dimension

teytaud:
no idea if that works

testing a collaborative coevolution around Shiwa

## Types of changes

Collaborative coevolution is not yet used inside Shiwa.

## Checklist


nevergrad/optimization/optimizerlib.py
```diff
@@ -1676,3 +1676,27 @@ def __init__(self, parametrization: IntOrParameter, budget: Optional[int] = None
                     self.optims = [CMA(self.parametrization, budget, num_workers)]
             else:
                 self.optims = [NGO(self.parametrization, budget, num_workers)]
+
+
+@registry.register
+class Alacrite(NGO):
```
jrapin, line 1682: why yet another optimizer? and where does the name come from :D

teytaud: Why another optimizer: black-box optimization is wide, and moves in mysterious ways -- Shiwa performs greatly, but does not use collaborative coevolution, so I'd like to try this additional combination.

From the discussions with various people and Gecco reviewers, we might try this, and more active portfolios, and develop the discrete case -- let me do that.
We might wait for merging -- this will probably move depending on experimental results.

Naming, well... random conflict-free name.

jrapin: i was not arguing for why this needs to be tried but rather why create yet another version. Ok to wait for merging, though maybe eventually NGO needs to be updated, or maybe keep One experimental version (`ExperimentalNGO`?) and one up to date version (`NGO`?)?
No hurry to decide, but we'll need to find a better way eventually, random names can be fun but from a user perspective they do not make sense and don't help to understand what it does

nevergrad/optimization/optimizerlib.py
```diff
@@ -1676,3 +1676,27 @@ def __init__(self, parametrization: IntOrParameter, budget: Optional[int] = None
...
+    """Nevergrad optimizer by competence map. You might modify this one for designing your own competence map."""
```
jrapin, line 1683: it may be useful to describe how it differs from other versions

PR #646, opened 2020-05-06
Add adversarial examples

LaurentMnr95:
## Types of changes

Add adversarial real examples

## Checklist


nevergrad/functions/images/core.py
```diff
@@ -32,36 +34,60 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
...
         assert index == 0  # For the moment only 1 target.
-        #path = os.path.dirname(__file__) + "/headrgb_olivier.png"
+        # path = os.path.dirname(__file__) + "/headrgb_olivier.png"
         path = Path(__file__).with_name("headrgb_olivier.png")
         image = PIL.Image.open(path).resize((self.domain_shape[0], self.domain_shape[1]), PIL.Image.ANTIALIAS)
-        self.data = np.asarray(image)[:,:,:3]  # 4th Channel is pointless here, only 255.
+        self.data = np.asarray(image)[:, :, :3]  # 4th Channel is pointless here, only 255.
 
-        # if problem_name == "adversarial": 
+        # if problem_name == "adversarial":
```
teytaud, line 42: You might remove this then.

nevergrad/functions/images/core.py
```diff
@@ -12,10 +12,12 @@
 import os
 from nevergrad.common.typetools import ArrayLike
 from .. import base
+import torch.nn as nn
+import torch
 
 
 class Image(base.ExperimentFunction):
-    def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
+    def __init__(self, problem_name: str = "recovering", index: int = 0, params: dict = {}) -> None:
```
teytaud, line 20: Cool, the change of signature is worth it :-)

nevergrad/optimization/optimizerlib.py
```diff
@@ -95,13 +95,28 @@ def _internal_ask_candidate(self) -> p.Parameter:
                                      mutator.get_roulette(self.archive, num=2))
             return pessimistic.set_standardized_data(data, reference=ref)
         # mutating
+<<<<<<< HEAD
```
teytaud, line 98: here you have a conflict between versions, which should be easy to solve -- just come back to the current master.

nevergrad/optimization/optimizerlib.py
```diff
@@ -95,13 +95,28 @@ def _internal_ask_candidate(self) -> p.Parameter:
...
+            return self.current_bests["pessimistic"].x + self._sigma * self._rng.normal(0, 1, self.dimension)
+        elif mutation == "cauchy":
+            # type: ignore
+            return self.current_bests["pessimistic"].x + self._sigma * self._rng.standard_cauchy(self.dimension)
+        elif mutation == "crossover":
+            if self._num_ask % 2 == 0 or len(self.archive) < 3:
+                return mutator.portfolio_discrete_mutation(self.current_bests["pessimistic"].x)
+            else:
+                return mutator.crossover(self.current_bests["pessimistic"].x, mutator.get_roulette(self.archive, num=2))
+=======
```
teytaud, line 111: ...i.e. remove from line HEAD to ==== in both cases.

nevergrad/functions/images/test_core.py
```diff
@@ -7,12 +7,64 @@
...
+        self.std = torch.Tensor(std)
+
+    def forward(self, x):
+        return (x - self.mean.type_as(x)[None, :, None, None]) / self.std.type_as(x)[None, :, None, None]
+
+
+class Classifier(nn.Module):
+    def __init__(self):
+        super(Classifier, self).__init__()
+        self.norm = Normalize(mean=[0.485, 0.456, 0.406],
```
teytaud, line 30: I would say that the Classifier should be created in init, not here, right ?

teytaud:
Missing dependencies for the CI: you can add that in .circleCI/config.yml

teytaud:
In nevergrad/benchmark/experiments.py, you can see below the experiment associated to the first case about images:

@registry.register
def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
    """AR coating. Problems about optical properties of nanolayers."""
    seedg = create_seed_generator(seed)
    algos = ["CMA", "Shiwa", "DE", "PSO", "RecES", "RecMixES", "RecMutDE", "ParametrizationDE"]
    for budget in [100 * 5 ** k for k in range(3)]:
        for num_workers in [1]:
            for algo in algos:
                for func in [Image()]:
                    xp = Experiment(func, algo, budget, num_workers=num_workers, seed=next(seedg))
                    if not xp.is_incoherent:
                        yield xp


You can replace [Image()] by [Image(), Image("adversarial")]
so that your case is included.

nevergrad/functions/images/core.py
jrapin: you'll need to `import typing as tp`
then `params: tp.Optional[tp.Dict[str, tp,Any]] = None`
As a general rule, *never* set a dict or a list as a default argument, this can lead to reaaaaally unexpected behavior, I can show you how at some point :D 
Better initialize it as None, then
```python
if params is None:
    params = {}
```

LaurentMnr95:
`            pip install torch==1.4.0+cpu torchvision==0.5.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
`

torchvision already seems to be in CI

nevergrad/functions/images/core.py
```diff
@@ -32,36 +34,54 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
...
-        #         # the attacked image.)
-
-        array = ng.p.Array(init=128 * np.ones(self.domain_shape), mutable_sigma=True,)
+        array = ng.p.Array(init=128 * np.ones(self.domain_shape), mutable_sigma=True, )
         array.set_mutation(sigma=35)
         array.set_bounds(lower=0, upper=255.99, method="clipping", full_range_sampling=True)
         max_size = ng.p.Scalar(lower=1, upper=200).set_integer_casting()
         array.set_recombination(ng.p.mutation.Crossover(axis=(0, 1), max_size=max_size)).set_name("")  # type: ignore
+
+        if problem_name == "adversarial":
```
jrapin, line 48: what is shared between the previous Image class and the new case? It seems to me there's hardly anything similar, in that case a different class would be much better

nevergrad/functions/images/test_core.py
```diff
@@ -7,12 +7,64 @@
...
+        self.model = resnet50(pretrained=True)
+
+    def forward(self, x):
+        return self.model(self.norm(x))
+
+
+def test_images_adversarial() -> None:
+    image_size = 224
+    classifier = Classifier()
+    classifier.cuda()
```
jrapin, line 41: does that require a gpu? If so, this wont work in the tests

nevergrad/functions/images/test_core.py
```diff
@@ -7,12 +7,64 @@
...
+    classifier.eval()
+    classifier = torch.nn.DataParallel(classifier, device_ids=range(torch.cuda.device_count()))
+    image_folder = datasets.ImageFolder("/datasets01_101/imagenet_full_size/061417/val",
```
jrapin, line 44: Hard path cannot be used. For tests you need some fake data

nevergrad/functions/images/test_core.py
```diff
@@ -7,12 +7,64 @@
...
+                                        transforms.Compose([transforms.Resize(image_size),
+                                                            transforms.CenterCrop(image_size),
+                                                            transforms.ToTensor()]))
+    data_loader = torch.utils.data.DataLoader(image_folder,
+                                              batch_size=1,
+                                              shuffle=True,
+                                              num_workers=8,
+                                              pin_memory=True)
+    image, label = next(iter(data_loader))
```
jrapin, line 53: I dont understand how this is useful

nevergrad/functions/images/core.py
```diff
@@ -32,36 +34,54 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
...
+            self.targeted = params["targeted"] if ("targeted" in params) else False
+            self.epsilon = params["epsilon"] if ("epsilon" in params) else 0.05
+            self.image_size = params["image_size"]
+            self.domain_shape = (self.image_size, self.image_size, 3)
+            self.image = params["image"]
+            self.label = params["label"]
+            self.classifier = params["classifier"]
```
jrapin, line 55: I don't think anyone will be able to use this if they need to provide a classifier, I agree with Olivier, there should be one provided, or a selection of classifiers to choose from?
Also, if they are big, they must not be submitted within the repo, and need to be hosted somewhere else and downloaded

nevergrad/functions/images/test_core.py
teytaud: For this to work everywhere, you might use "if".
E.g." if /datasets01_101 exists then use it, otherwise use xxxx1.jpg and xxx2.jpg (with these two images inside Nevergrad)"

nevergrad/functions/images/core.py
```diff
@@ -32,36 +56,66 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
...
-        assert self.problem_name == "recovering"
         assert self.index == 0
         value = np.sum(np.fabs(np.subtract(x, self.data)))
-
-        # Here we should implement "adversarial" and others.
         return value
 
+
+class ImageAdversarial(base.ExperimentFunction):
+    def __init__(self, params: dict = {}) -> None:
```
jrapin, line 86: this needs a simplified signature, with maybe a couple of parameters, like `target: str, classifier:str, epsilon: float`, and depending on the names, you instantiate the corresponding classifier etc. Otherwise this wont be useable out of the box, and hard to use in benchmarks

nevergrad/functions/images/test_core.py
```diff
@@ -7,15 +7,56 @@
...
 from . import core
+from torchvision.models import resnet50
+import torch.nn as nn
+import torch
+import torchvision.models
+from torchvision.models.resnet import model_urls
+
+model_urls['resnet50'] = model_urls['resnet50'].replace('https://', 'http://')
+
+class Normalize(nn.Module):
```
jrapin, line 18: this can (should?) be removed from here

jrapin:
The new error in windows is really weird but unrelated, i'll need to check why it's there, but don't worry about it

nevergrad/functions/images/core.py
```diff
@@ -29,34 +33,131 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
         # Storing data necessary for the problem at hand.
         assert problem_name == "recovering"  # For the moment we have only this one.
         assert index == 0  # For the moment only 1 target.
```
teytaud, line 35: In your case, Laurent, there is no "assert index == 0"; index is the index of the image to be attacked.

nevergrad/functions/images/core.py
```diff
@@ -29,34 +33,131 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
...
+    def forward(self, x):
+        return self.model(x.flatten(x.shape[0], -1))
+
+
+# pylint: disable=too-many-arguments
+class ImageAdversarial(base.ExperimentFunction):
+
+    def __init__(
+        self,
+        classifier: nn.Module,
```
teytaud, line 99: Most arguments should have a default value, so that the benchmark can run create "ImageAdversarial()" and get an objective function; or maybe just one integer, which is the index of the attacked image in ImageNet.

nevergrad/functions/images/test_core.py
teytaud: More precisely, this and almost everything in this file should be in image/core.py and called in __init__. The test should not specify the classifier or whatever; the classifier should be created from __init__ in image/core.py.

nevergrad/functions/images/core.py
teytaud: I agree with this. Or everything has a default value.
One of the arguments could be the index in ImageNet, with default value 0. The code should look for the default path in /datasetxxxxx/xxxx, use those images if they are there --- otherwise we can use one or two images (we will not copy the entire image net inside Nevergrad :-) ).

(we could also do something like https://stackoverflow.com/questions/40251150/using-python-to-download-images-via-urllib-and-testing-if-the-website-exists, maybe, but maybe not for now...)

nevergrad/functions/images/core.py
jrapin: No I would say no need:
- the aim for Laurent is to be able to plug anything inside it, so let's leave it this way and avoid weird behavior with default values which do not match some other inputs etc
- it would not work on benchmarks anyway since we cant get a descriptor for a module etc
- there's a `from_testbed` classmethod where we can define testbeds from a name. This way they can be used for test beds easily.

nevergrad/benchmark/experiments.py
```diff
@@ -892,6 +893,20 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+@registry.register
+def adversarial_images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Adversarial attacks."""
+    seedg = create_seed_generator(seed)
+    algos = ["CMA", "Shiwa", "DE", "PSO", "RecES", "RecMixES", "RecMutDE", "ParametrizationDE"]
+    for budget in [100 * 5 ** k for k in range(3)]:
+        for num_workers in [1]:
+            for algo in algos:
+                # Simple example with just the default case.
+                for func in [ImageAdversarial()]:
```
jrapin, line 905: this would need to be `ImageAdversarial.from_testbed(name)`. You won't be able to use it this way reliably.
Also, please instantiate the function before all the for loops, that will save a lot of computation.

nevergrad/functions/images/core.py
jrapin: I'm not sure we are ready yet for actual experiments from this. Cf my comments.
I recommend merging if this work for Laurent, and working on the testbed classmethod to provide support for experiments.

jrapin: (once the test work, i.e. after updating or removing the code in experiments.py and merging master)

nevergrad/functions/images/core.py
teytaud: Ok for no default value.
But we should really include this in benchmarks.py, and the user should not have to create the classifier and the images. "from_testbed" is ok (I had in mind something similar, like a string that would be the name of the testcase and would build all parameters -- from_testbed is quite analogous to that),

jrapin: it's exactly that actually ;) with a few more (default) parameters

nevergrad/benchmark/experiments.py
```diff
@@ -892,6 +893,20 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
                     if not xp.is_incoherent:
                         yield xp
 
+@registry.register
+def adversarial_images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Adversarial attacks."""
+    seedg = create_seed_generator(seed)
```
teytaud, line 899: So we create resnet50(pretrained=True)   here ?

jrapin: No, it should be in https://github.com/facebookresearch/nevergrad/pull/646/files#diff-9237fdd46c25be2ca72b153f11ed0cebR130 ;)

nevergrad/functions/images/core.py
```diff
@@ -6,8 +6,11 @@
 from pathlib import Path
 
 import numpy as np
-import nevergrad as ng
 import PIL.Image
+import torch.nn as nn
+import torch
+
+import nevergrad as ng
```
teytaud, line 13: importing Nevergrad inside Nevergrad ? you just have to import parameters as p, right ?

nevergrad/benchmark/experiments.py
```diff
@@ -1181,5 +1217,45 @@ def bragg_structure(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+            self.norm = Normalize(mean=[0.485, 0.456, 0.406],
+                                  std=[0.229, 0.224, 0.225])
+            self.model = resnet50(pretrained=True)
+
+        def forward(self, x):
+            return self.model(self.norm(x))
+
+    image_size = 224
+    data_folder = "/datasets01_101/imagenet_full_size/061417/val"
+    data_loader = torch.utils.data.DataLoader(
```
teytaud, line 1248: If the path does not exist just stop (no yield command at all), so that this does not crash on CircleCI and the CI will be ok, whereas it will work on the dashboard.
For testing that everything works you can do 
python -m nevergrad.benchmark AdversarialAttack --seed=12 --repetitions=1 --plot

nevergrad/benchmark/experiments.py
```diff
@@ -1181,5 +1217,45 @@ def bragg_structure(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+                                             transforms.CenterCrop(image_size),
+                                             transforms.ToTensor()])),
+        batch_size=1,
+        shuffle=True,
+        num_workers=8,
+        pin_memory=True)
+
+    classifier = Resnet50()
+
+    # TODO: Add experiment here
```
teytaud, line 1261: The end of the function is just something like "optims = ["DE", "PSO", "DiagonalCMA", "CMA"] and the kind of loop you can see in other xps.

nevergrad/functions/images/core.py
jrapin: I like it better that way actually, so that it's closer to a user's code.
I do this mostly for parameters

nevergrad/benchmark/experiments.py
jrapin: Also, all this should be in the tesbed method of Adversarial, not in this file

nevergrad/functions/images/core.py
LaurentMnr95: That's jeremy's line of code  so I tend to agree with him ^^

nevergrad/benchmark/experiments.py
LaurentMnr95: OK will correct that! but  why in testbed? I do not want my loader to load each  time, it is very costful no?

nevergrad/benchmark/experiments.py
LaurentMnr95: yes but I do not know exactly what  to put inside, let see on Wed

nevergrad/benchmark/experiments.py
jrapin: > but why in testbed?

this way all adversarial example code stays in the corresponding package, and is reusable in other experiments without copy pasting everything, which is a **very** bad idea.

> I do not want my loader to load each time, it is very costful no?

How costful is it? That could be a good reason to avoid implementing this in testbed. Then again we could make it work without copying the loader, but that could require some work and I would be the one doing it. So if it's costy lets keep it that way for now.

LaurentMnr95: Costly like loading ImageNet every single time :p, so will keep it this way for now, and we'll see later :)

nevergrad/functions/images/core.py
```diff
@@ -29,34 +32,102 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
...
+        array.set_recombination(ng.p.mutation.Crossover(axis=(1, 2), max_size=max_size))  # type: ignore
+
+        super().__init__(self._loss, array)
+        self.register_initialization(classifier=classifier, image=image, label=label,
+                                     targeted=targeted, epsilon=epsilon)
+        # classifier and image cant be set as descriptors
+        self.add_descriptors(label=label, targeted=targeted, epsilon=epsilon)
+
+    @classmethod
+    def from_testbed(
```
teytaud, line 101: For creating an objet, adversarial_image = ImageAdversarial.from_testbed("resnet50")

nevergrad/benchmark/experiments.py
LaurentMnr95: actually I also want it to be modulable (ImageNet is one dataset, but I may want use any image), I want to specify an Image in my attack, not a dataset :)

nevergrad/functions/images/core.py
```diff
@@ -29,34 +32,102 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
...
+            cls,
+            name: str,
+            label: int = 0,
+            targeted: bool = False,
+            epsilon: float = 0.05
+    ) -> "ImageAdversarial":
+        if name == "test":
```
teytaud, line 108: elif name == "resnes50":
    classifier = ResNet50()

(and move the definition of ResNet50 to the present file!)

nevergrad/benchmark/experiments.py
```diff
@@ -1257,5 +1299,61 @@ def bragg_structure(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+            self.model = resnet50(pretrained=True)
+
+        def forward(self, x):
+            return self.model(self.norm(x))
+
+    image_size = 224
+    data_folder = "/datasets01_101/imagenet_full_size/061417/val"
+    if not os.path.exists(data_folder):
+        print("this path does not exist")
+        return
```
teytaud, line 1331: I see two solutions.
Solution 1: Use a trivial data loader with one single image. This is the cleanest solution, as it has a better code coverage.
Solution 2: Easier. Maybe define an exception, of the form ExperimentSkippedOnCI. Instead of "return", do "raise ExperimentSkippedOnCI", and in the test you do <<try: .... except ExperimentSkippedOnCI: pass>>.

requirements/bench.txt
```diff
@@ -4,10 +4,12 @@ xlrd>=1.2.0
...
 torch>=1.2.0
+torchvision
 hiplot
 fcmaes>=0.9.5.6
 pandas>=0.23.4
 pyproj>=2.6.1
 Pillow
 tqdm
-pyomo>=5.7
+torchvision
```
jrapin, line 14: added twice

nevergrad/benchmark/experiments.py
LaurentMnr95: solved

jrapin (approved):
Lets merge, I'll make minor edits later on

PR #651, opened 2020-05-07
Adding boston and diabetes and more artificial

teytaud:
## Types of changes

Real world is more fun.

## Checklist

### DIFF
--- a/nevergrad/functions/ml/mlfunctionlib.py
+++ b/nevergrad/functions/ml/mlfunctionlib.py
@@ -13,6 +13,7 @@ from nevergrad.parametrization import parameter as p
 from nevergrad.common import tools
 from nevergrad.common.typetools import ArrayLike
 from sklearn.tree import DecisionTreeRegressor  # type: ignore
+from sklearn.neural_network import MLPRegressor  # type: ignore
 
 from ..base import ExperimentFunction
 from .. import utils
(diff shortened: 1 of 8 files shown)


nevergrad/functions/ml/mlfunctionlib.py
```diff
@@ -77,10 +79,12 @@ def _ml_parametrization(self,
 
         return result / self.num_data  # We return a 10-fold validation error.
 
-    def __init__(self, regressor: str, data_dimension: int):
+    def __init__(self, regressor: str, data_dimension: tp.Optional[int] = None, dataset_name: str = "artificial"):
```
jrapin, line 82: at may be simpler to just have `dataset: tp.Union[str, int]`: either it's a int and it's artificial, or it's not and it's the given dataset. That avoids coupled parameters

nevergrad/functions/ml/mlfunctionlib.py
```diff
@@ -174,16 +178,36 @@ def __init__(self, regressor: str, data_dimension: int):
...
+            self.num_data = num_train_data
+            for cv in range(10):
+                train_range = np.arange(num_train_data) % 10 != cv 
+                valid_range = np.arange(num_train_data) % 10 == cv
+                self.X_train += [self.X[train_range]]
+                self.y_train += [self.y[train_range]]
+                self.X_valid += [self.X[valid_range]]
+                self.y_valid += [self.y[valid_range]]
+            self.X_test = data[0][1::2]
+            self.y_test = data[1][1::2]
```
jrapin, line 201: all the attribues need to be typed in the init

nevergrad/functions/ml/mlfunctionlib.py
```diff
@@ -174,16 +178,36 @@ def __init__(self, regressor: str, data_dimension: int):
...
 
         num_data: int = 120  # Training set size.
         self.num_data = num_data
         
         # Training set.
         X = np.arange(0., 1., 1. / (num_data * data_dimension))
         X = X.reshape(-1, data_dimension)
-        random_state = np.random.RandomState(17)
-        random_state.shuffle(X)
+        np.random.shuffle(X)
```
jrapin, line 210: you need to use `self.parametrization.random_state` and avoid pulling from the default random state.
to do so, you may need to make sure you call `super` before `get_dataset` in the `init`

nevergrad/functions/ml/mlfunctionlib.py
teytaud: This means that we have a single artificial dataset ?
I would go for having dozens of artificial cases :-)

nevergrad/functions/ml/mlfunctionlib.py
teytaud: only num_data is missing; I add it.

nevergrad/functions/ml/mlfunctionlib.py
teytaud: I'm lost! In xpbase.py, I see:
==================
        if self.seed is not None and self._optimizer is None:
            # Note: when resuming a job (if optimizer is not None), seeding is pointless (reproducibility is lost)
            np.random.seed(self.seed)  # seeds both functions and parametrization (for which random state init is lazy)
            random.seed(self.seed)
            torch.manual_seed(self.seed)  # type: ignore
===========
which suggests that we are precisely pulling from the default random state.

jrapin: it's more a hack to make sure it works well in all cases, but it's much better to avoid it

nevergrad/functions/ml/mlfunctionlib.py
jrapin: if you think you will have more, ok ;)
but then it's a bit weird that your default parameters make the function crash since data_dimension cannot be 0 for the artificial dataset
Also, `dataset` may be enough instead of `dataset_name`

nevergrad/functions/ml/mlfunctionlib.py
teytaud: Ok I see. Fixed.

nevergrad/functions/ml/test_mlfunctionlib.py
```diff
@@ -12,30 +12,54 @@
 
 
 def test_ml_tuning() -> None:
+    # Testing a decision tree with only depth as a parameter.
+    np.random.seed(17)
```
jrapin, line 16: to prove it's working well, you should go for `func.rng.seed(17)` instead

teytaud: Done!

teytaud:
Added more artiticial cases.
Fixed names as requested.

teytaud:
Added more tests.

nevergrad/functions/ml/test_mlfunctionlib.py
teytaud: Done, thx!

PR #684, opened 2020-05-16
Add an experimental bouncing bounding method

jrapin:
## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context / Related issue

Hard clipping on the bounds can lead to very detrimental loss of diversity. Bouncing on the border can help avoid this will being efficient and simple

## Checklist

- [x] The documentation is up-to-date with the changes I made.
- [x] I have read the [**CONTRIBUTING**](https://facebookresearch.github.io/nevergrad/contributing.html) document and completed the CLA (see [**CLA**](https://facebookresearch.github.io/nevergrad/contributing.html#contributor-license-agreement-cla)).
- [x] All tests passed, and additional code has been covered with new tests.

PR #691, opened 2020-05-20
Make bouncing the default for constraints

### LINKED ISSUE #621: Bounded variable with clipping can lead to early converging for PSO and DE
With small population, clipped variables (eg. `Scalar(lower=0, upper=1)`) can lead to early convergence of PSO and DE. Indeed, if for some reasons all of the individuals reach the same bound, they will not be able to move away from it anymore.

Currently, we do not have any idea of how to fix this. Using other methods than clipping for bounds can help, but will lead to other issues (inefficiency, transformed space etc)

jrapin:
## Types of changes

This is a follow-up of #684 making `bouncing` the default for bounding constraints. Hard clipping on the bounds can lead to very detrimental loss of diversity. Bouncing on the border can help avoid this in being closer to an actual constraint through resampling, while being much more efficient in large dimension
Closes #621 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   Still, some optimizers (like `TBPSA`) may recommend a non-evaluated point.
 - `Choice` now takes a new `repetitions` parameters for sampling several times,
   it is equivalent to :code:`Tuple(*[Choice(options) for _ in range(repetitions)])` but can be be around 30x faster for large numbers of repetitions [#670](https://github.com/facebookresearch/nevergrad/pull/670).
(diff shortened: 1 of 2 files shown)


jrapin:
> Do I understand correctly that bouncing is tested once and then clipping is applied ?

What do you mean tested once?
It "bounces" once on the border, and if it reaches the other border then it clips:
Eg: starting point 1, and bounds in [0, 10]:

- moving of -3 leads to 2
- moving of 1 leads to 2
- moving of 10 leads to 9 (bounces once on border)
- moving of -11 or lower leads to 10 (bounces once at 0, then clips at 10)

PR #722, opened 2020-06-09
Ctulo2

teytaud:
## Types of changes

Shiwa is dirty (due to me).
So I reimplemented it.
I also modified a few things.
This also includes an improvement of cGA.
Other improvements are expected.
The plan is (i) merge this PR (ii) check performance (iii) if everything fine, remove Shiwa and rename Ctulo as NGO10 or something like that.

CI and a few preliminary tests.
Additional tests have been added.

## Checklist


nevergrad/optimization/optimizerlib.py
```diff
@@ -1755,3 +1756,70 @@ def _internal_ask_candidate(self) -> p.Parameter:
 
     def _internal_tell_candidate(self, candidate: p.Parameter, value: float) -> None:
         self._optim.tell(candidate, value)
+
+
+@registry.register
+class Ctulo(base.Optimizer):
```
jrapin, line 1762: why don't you rename it directly NGO2?

jrapin:
Please don't remove Shiwa so soon, now that it was advised to use it and published, it is not user-friendly to remove it.
When NGO2 is working, I would recommend to make Shiwa inherit from it and just add a deprecation warning to it. This way it will still works for some time, remove most of the code, and allow users to adapt.
Also, if there are more NGO versions coming, we need to make `NGO` point to the latest version. This way we can keep NGO2, NGO3 etc as long as they make sence, and advise people to always use NGO as it is the one supposed to have the best results.

jrapin:
Also, in what is it cleaner that Shiwa? :s

nevergrad/optimization/optimizerlib.py
teytaud: Ok I do so.

teytaud: Well NGO10 because we have had several variants with various numbers.

teytaud:
Ok for a NGO pointing to NGO-last.
I have no plan to remove Shiwa in the near future, and ok I will not remove it at all actually given that there are already several publications mentioning it.

teytaud:
> Also, in what is it cleaner that Shiwa? :s

Shiwa has dead branches.
It is also longer than necessary, with several equivalent branches.

PR #750, opened 2020-07-03
 update experiments.py: plenty of docs.

teytaud:
## Types of changes

Docs are now visible on the dashboard, so more doc ==> more cool.

## Checklist


nevergrad/benchmark/experiments.py
```diff
@@ -36,15 +36,19 @@
 # for black (since lists are way too long...):
 # fmt: off
 
+default_optims = None  # ["NGO10", "CMA", "Shiwa"]
```
jrapin, line 39: why do you do that, you can do it better using the optimizer groups

jrapin: optimizer groups will be way more sustainable, and allow shuffling, I would stronlgy advise to remove this

teytaud: This does not do anything as long as its known. The only purpose is to replace or complete the list of optimizers when it's not None. So it's not the same purpose as optimizer group. The point is to run specifically a short list of optimizers on all benchmarks. I do not see any other option for this.

Ok for removing this feature if you don't like it.

teytaud:
@jrapin  I modified; ok under this form ?

nevergrad/benchmark/experiments.py
jrapin: You can do the same think using the optimizer groups, just have one group that is always there (or always added) and contains nothing by default (and that you can change when needed)
you will also avoid duplicates this way

PR #751, opened 2020-07-03
Stopping optimizers which will never finish in time given their current time consumption.

teytaud:
## Types of changes

I've considered this as a bugfix. Explanation:
- Currently, if optimizer A is super slow and fails, and if optimizer B is behind optimizer A in a job,
    then both optimizer A and optimizer B are penalized (at least if I understand correctly our Slurm method for 
    running jobs :-) ). In addition, we waste a lot of computational power if it's visible after 1 hour that we are not 
    going to finish in time and we still run until 72 hours.
- This introduces a bias as B is biased without a good reason.
- Currently, we fix it by running the dashboard on sufficiently small instances for having everything running fine.
- The proposed fix is possibly not perfect, but I think it's a significant improvement.

## Checklist


nevergrad/optimization/base.py
```diff
@@ -461,6 +464,14 @@ def minimize(
                     sleeper.stop_timer()
                 while self._finished_jobs:
                     x, job = self._finished_jobs[0]
+
+                    if max_time >= 0:
```
jrapin, line 468: I would rather this was not included in minimize , a callback added in the xps would be much better

jrapin: Callbacks: https://facebookresearch.github.io/nevergrad/optimizers_ref.html#callbacks

The reason is that this would avoid changing the API, especially since eventually we need to add some kind of time out, but probably not this way

teytaud:
I'll just entirely revisit long xps.

PR #788, opened 2020-08-07
Improve pyomo typing

jrapin:
## Types of changes

## Checklist

### DIFF
--- a/nevergrad/functions/pyomo/core.py
+++ b/nevergrad/functions/pyomo/core.py
@@ -2,21 +2,27 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-import typing as tp
 from functools import partial
 import numpy as np
-import pyomo.environ as pyomo # type: ignore
+import pyomo.environ as pyomo
+import nevergrad.common.typing as tp
 from nevergrad.parametrization import parameter as p
 from .. import base
 
-def _convert_to_ng_name(pyomo_var_key):
+
+ParamDict = tp.Dict[str, p.Parameter]
+
+
+def _convert_to_ng_name(pyomo_var_key: tp.Any) -> str:
     if isinstance(pyomo_var_key, str):
         return '"' + str(pyomo_var_key) + '"'
     else:
         return str(pyomo_var_key)
 
 
-def _make_pyomo_range_set_to_parametrization(domain : pyomo.RangeSet, params : dict, params_name : str):
+def _make_pyomo_range_set_to_parametrization(
+    domain: pyomo.RangeSet, params: ParamDict, params_name: str
+) -> ParamDict:
     # https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Sets.html
     # Refer to the implementation in pyomo/core/base/set.py
     ranges = list(domain.ranges())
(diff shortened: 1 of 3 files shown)


nevergrad/functions/pyomo/core.py
```diff
@@ -129,28 +134,27 @@ def __init__(self, model : pyomo.Model) -> None:
...
 
-
-    def _pyomo_obj_function_wrapper(self, i, **k_model_variables) -> float: # type: ignore
+    def _pyomo_obj_function_wrapper(self, i: int, **k_model_variables: tp.Dict[str, tp.Any]) -> float:
         for k, v in k_model_variables.items():
-            exec(f"self._model_instance.{k} = {v}") # exec-used: ignore
-        return pyomo.value(self.all_objectives[i] * self.all_objectives[i].sense) #Single objective assumption
-
+            # TODO find a way to avoid exec
+            exec(f"self._model_instance.{k} = {v}")  # pylint: disable=exec-used
```
jrapin, line 148: @pacowong do you think there is any way we can avoid exec?
I think that would require packing some variables together through `Tuple` for instance

pacowong: @jrapin I have just tested locally. You can eliminate exec by storing the accessing method in a class variable (self.instru_param_types). The solution is as follows:

Step 1:
```python
    def __init__(self, model : pyomo.Model) -> None:
        ...
        instru_params: ParamDict = {}
        self.instru_param_types : dict = {}
        ...
        for v in self._model_instance.component_objects(pyomo.Var, active=True):
            self.all_vars.append(v)
            _make_pyomo_variable_to_parametrization(v, instru_params, self.instru_param_types)
        ...
```

Step 2: the function will also update the param_types
```python
def _make_pyomo_variable_to_parametrization(
    model_component: pyomo.Var, params: ParamDict, param_types : dict) -> ParamDict:
    ...
    for k, v in model_component._data.items():
        if isinstance(v, pyomo.base.var._GeneralVarData):
            if v.is_fixed():
                raise NotImplementedError
            if k is None:
                params_name = str(model_component.name)
                param_types[params_name] = (0,)
            elif isinstance(k, str):
                params_name = f"{model_component.name}[{_convert_to_ng_name(k)}]"
                param_types[params_name] = (1, model_component.name, k)  
            else:
                params_name = f"{model_component.name}[{_convert_to_ng_name(k)}]"
                param_types[params_name] = (2, model_component.name, k) 
            if isinstance(v.domain, pyomo.RangeSet):
                params = _make_pyomo_range_set_to_parametrization(v.domain, params, params_name)
            elif isinstance(v.domain, pyomo.Set) and v.domain.isfinite():
                if v.domain.isordered():
                    params[params_name] = p.Choice(list(v.domain.ordered_data()))
                else:
                    params[params_name] = p.Choice(list(v.domain.data()))
            else:
                raise NotImplementedError(f"Cannot handle domain type {type(v.domain)}")
        else:
            raise NotImplementedError(f"Cannot handle variable type {type(v)}")
    return params
```

Step 3: replace the two statements with exec in the wrapper functions. Note that var_entry can be a string, None, or int.
```python
            if self.instru_param_types[k][0] == 0:
                setattr(self._model_instance, k, v)
            else:
                var_name, var_entry = self.instru_param_types[k][1], self.instru_param_types[k][2]
                getattr(self._model_instance, var_name)[var_entry] = v
```

jrapin: if the code is not much longer that can alleviate some (mild) security concerns. If it's adding many lines, it's not worth it.
A solution based on `Tuple` may be more native/natural though, is there a way to capture the underlying structure of pyomo variables? figuring out that there is x[0] x[1] x[2] we could just create x which encapsulates all this,
Then affectation becomes straightforward.

PR #803, opened 2020-09-01
Add a specific ArrayExperimentFunction using symmetry for diversity

teytaud:
## Types of changes

One of the biggest sources of poor experimental result is running optimization tests
on a single objective function.
The worst and most classical case is when the optimum is at zero and experimenters did not carefully translate.
A simple solution consists in translating the objective function.
However, for real-world objective functions, we can not translate.
What we can reasonably do, maybe, is to symmetrize parametrization:
if there is a variable which is a power between 10 MW and 20 MW, we replace x by 30-x ==> same domain, same geometry of the objective function.
For diversity, we can do it with probability 50% for each variable.
This PR is here for that.

## Checklist


nevergrad/functions/base.py
```diff
@@ -204,6 +204,36 @@ def update_leaderboard(identifier: str, loss: float, array: np.ndarray, verbose:
         pass  # better avoir bugs for this
 
 
+class ArrayExperimentFunction(ExperimentFunction):
+    """Adds a "symmetry" parameter, which allows the creation of many symmetries of a given function."""
+
+    def symmetrized_function(self: AEF, x: tp.Any):
```
jrapin, line 210:
```suggestion
    def symmetrized_function(self, x: np.ndarray) -> tp.Loss:
```

`AEF` is only necessary in very specify cases (typically for copy constructors, which return the same type as `self` while you don't necessary know the type of `self` since it can be a subclass)

nevergrad/functions/base.py
```diff
@@ -204,6 +204,36 @@ def update_leaderboard(identifier: str, loss: float, array: np.ndarray, verbose:
...
+                y[i] = -x[i]  # We should rather symmetrize w.r.t the center of Parameter. TODO
+            symmetry = symmetry // 2
+        return self._inner_function(y)  # type: ignore
+
+    def __init__(self: AEF, function: tp.Callable[..., tp.Loss], parametrization: p.Parameter, symmetry: int = 0) -> None:
+        """ Same parameters as ExperimentFunction, plus "symmetry".
+        symmetry: an int, 0 by default.
+        if not zero, a symmetrization is applied to the input; each of the 2^d possible values
+        for symmetry % 2^d gives one different function.
+        Makes sense if and only if (1) the input is a single ndarray (2) the domains are symmetric."""
```
jrapin, line 226: Please move the init as the first method, and all class related docstring should be in the init, hence you should remove 
"""Adds a "symmetry" parameter, which allows the creation of many symmetries of a given function."""
from  above and add it here

nevergrad/functions/base.py
```diff
@@ -204,6 +204,36 @@ def update_leaderboard(identifier: str, loss: float, array: np.ndarray, verbose:
...
+            if symmetry % 2 == 1:
+                y[i] = -x[i]  # We should rather symmetrize w.r.t the center of Parameter. TODO
+            symmetry = symmetry // 2
+        return self._inner_function(y)  # type: ignore
+
+    def __init__(self: AEF, function: tp.Callable[..., tp.Loss], parametrization: p.Parameter, symmetry: int = 0) -> None:
+        """ Same parameters as ExperimentFunction, plus "symmetry".
+        symmetry: an int, 0 by default.
+        if not zero, a symmetrization is applied to the input; each of the 2^d possible values
+        for symmetry % 2^d gives one different function.
```
jrapin, line 225: maybe symmetry >= 2d should not be allowed?

jrapin (commented):
can you add a few tests? it seems a bit dangereous without it

nevergrad/functions/base.py
teytaud: Done.

nevergrad/functions/base.py
teytaud: Done.

teytaud: Maybe the creation of all symmetries, of a subset of all symmetries, should be possible... for later.

teytaud:
We accept p.Array(shape=(10,)), but not p.Instrumentation(p.Array(shape=(10,)))
This just means that I'm not an instrumentation wizard --- I should.

nevergrad/functions/base.py
```diff
@@ -204,6 +203,59 @@ def update_leaderboard(identifier: str, loss: float, array: np.ndarray, verbose:
...
+        self._symmetry = symmetry
+        if self._symmetry != 0:
+            self._function = self.symmetrized_function
+        else:
+            self._function = function
+
+    def symmetrized_function(self, x: np.ndarray) -> tp.Loss:
+        assert isinstance(x, np.ndarray), "symmetry != 0 works only when the input is an array."
+        assert len(x.shape) == 1, "only one-dimensional arrays for now."
+        y = x
```
jrapin, line 246:
```suggestion
        y = np.array(x, copy=True)
```
This is a security to avoid border effects

nevergrad/functions/base.py
```diff
@@ -204,6 +203,59 @@ def update_leaderboard(identifier: str, loss: float, array: np.ndarray, verbose:
...
+        symmetry: int = self._symmetry  # type: ignore
+        for i in range(len(y)):
+            if symmetry % 2 == 1:
+                if self.parametrization.bounds[0] is not None and self.parametrization.bounds[1] is not None:  # type: ignore
+                    middle = (self.parametrization.bounds[0][i] + self.parametrization.bounds[1][i]) / 2.  # type: ignore
+                else:
+                    middle = 0.
+                y[i] = middle - x[i]  # We should rather symmetrize w.r.t the center of Parameter. TODO
+            symmetry = symmetry // 2
+        return self._inner_function(y)  # type: ignore
```
jrapin, line 256: once this is tested and works (and possibly merged) a better option of doing this is to precomute the transform that must be applied (it's basically  a x + b  if I am not forgetting anything, so we only need a and b

nevergrad/functions/test_base.py
```diff
@@ -23,7 +23,13 @@ def test_experiment_function() -> None:
         constkwarg="blublu",
         plop=p.Choice([3, 4]),
     ))
+    iarrayfunc = base.ArrayExperimentFunction(_arg_return,   # type: ignore
+        p.Array(shape=(10,)),
+        symmetry=247,
+    )
     np.testing.assert_equal(ifunc.dimension, 8)
+    np.testing.assert_equal(iarrayfunc.dimension, 10)
+    iarrayfunc(np.zeros(10))
```
jrapin, line 32: you should check that some coefficients were indeed changed, and that it handles boundary. An independent tests would be better.

jrapin: Also, I suppose iarrayfunc.copy() should return the same func, does that work?

teytaud: Tests added, fixes applied, thx for your help.

teytaud: (now adding cases with box-constraints)

nevergrad/functions/test_base.py
```diff
@@ -23,7 +23,18 @@ def test_experiment_function() -> None:
...
+        symmetry=247,
+    )
+    iarrayfunc2 = base.ArrayExperimentFunction(lambda x: sum(x),   # type: ignore
+        p.Array(shape=(10,)).set_bounds(-0.5, 6.),
+        symmetry=111,
+    )
     np.testing.assert_equal(ifunc.dimension, 8)
+    np.testing.assert_equal(iarrayfunc.dimension, 10)
+    assert iarrayfunc(np.zeros(10)) == iarrayfunc.copy()(np.zeros(10))
+    assert iarrayfunc(np.ones(10)) != iarrayfunc2(np.ones(10))
```
jrapin, line 37: please put this is a dedicated test, tests should be unitary as much as possible.

jrapin (changes requested):

nevergrad/functions/test_base.py
jrapin: Also, probably test against an actual value that you *expect*, to make sure it works.

nevergrad/functions/base.py
```diff
@@ -204,6 +203,62 @@ def update_leaderboard(identifier: str, loss: float, array: np.ndarray, verbose:
...
+    """
+
+    def __init__(self, function: tp.Callable[..., tp.Loss], parametrization: p.Parameter, symmetry: int = 0) -> None:
+        """Adds a "symmetry" parameter, which allows the creation of many symmetries of a given function.
+
+        symmetry: an int, 0 by default.
+        if not zero, a symmetrization is applied to the input; each of the 2^d possible values
+        for symmetry % 2^d gives one different function.
+        Makes sense if and only if (1) the input is a single ndarray (2) the domains are symmetric."""
+        super().__init__(function, parametrization)
```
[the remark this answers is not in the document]
teytaud: Oh my god.

nevergrad/functions/test_base.py
teytaud: > please put this is a dedicated test, tests should be unitary as much as possible.

Done.

teytaud: > Also, probably test against an actual value that you _expect_, to make sure it works.

Done.

nevergrad/functions/base.py
```diff
@@ -234,6 +233,60 @@ def update_leaderboard(identifier: str, loss: float, array: np.ndarray, verbose:
...
+        if not zero, a symmetrization is applied to the input; each of the 2^d possible values
+        for symmetry % 2^d gives one different function.
+        Makes sense if and only if (1) the input is a single ndarray (2) the domains are symmetric."""
+        self._inner_function = function
+        super().__init__(self.symmetrized_function, parametrization)
+        assert isinstance(parametrization, p.Array), f"{type(parametrization)} is not p.Array; {parametrization.parameters}."
+        assert (parametrization.bounds[0] is None) == (parametrization.bounds[1] is None)
+        assert len(parametrization._constraint_checkers) == 0
+        assert symmetry >= 0
+        assert symmetry < 2 ** self.dimension
```
jrapin, line 266: if you plan to apply this on experiment functions, maybe you'd need to extract its descriptors? or it's not a plan?

nevergrad/functions/test_base.py
```diff
@@ -48,6 +48,27 @@ def test_experiment_function() -> None:
...
+        symmetry=247,
+    )
+    iarrayfunc2 = base.ArrayExperimentFunction(lambda x: sum(x),   # type: ignore
+        p.Array(shape=(10,)).set_bounds(-0.5, 6.).set_name(""),
+        symmetry=111,
+    )
+    iarrayfunc3 = base.ArrayExperimentFunction(lambda x: sum(x),   # type: ignore
+        p.Array(shape=(10,)).set_bounds(-0.5, 6.).set_name(""),
+        symmetry=111,
+    )
```
jrapin, line 63:
```suggestion
    iarrayfuncs = [
        base.ArrayExperimentFunction(
            lambda x: sum(x),   # type: ignore
            p.Array(shape=(10,)).set_bounds(-0.5, 6.).set_name(""),
            symmetry=s,
        ) for s in [247, 111, 111]
    ]
```
always avoid duplicated code

nevergrad/functions/test_base.py
```diff
@@ -48,6 +48,27 @@ def test_experiment_function() -> None:
...
+    np.testing.assert_equal(iarrayfunc.dimension, 10)
+    assert iarrayfunc(np.zeros(10)) == iarrayfunc.copy()(np.zeros(10))
+    assert iarrayfunc(np.zeros(10)) == 8.25
+    assert iarrayfunc2(np.zeros(10)) == 11.0
+    assert iarrayfunc(np.ones(10)) != iarrayfunc2(np.ones(10))
+    assert iarrayfunc3(np.ones(10)) == iarrayfunc2(np.ones(10))
```
jrapin, line 69:
```suggestion
    np.testing.assert_equal(iarrayfunc[0].dimension, 10)
    assert iarrayfunc[0](np.zeros(10)) == iarrayfunc[0].copy()(np.zeros(10))
    assert iarrayfunc[0](np.zeros(10)) == 8.25
    assert iarrayfunc[1](np.zeros(10)) == 11.0
    assert iarrayfunc[0](np.ones(10)) != iarrayfunc[1](np.ones(10))
    assert iarrayfunc[2](np.ones(10)) == iarrayfunc[1](np.ones(10))
```

nevergrad/functions/base.py
jrapin: maybe it's not a good idea at all

jrapin (approved):
see changes for the tests, to avoid duplication, then you can merge if tests pass and you are confident that it works well, I did not crosscheck the logic

nevergrad/functions/base.py
teytaud: Reimplemented that way.

nevergrad/functions/base.py
teytaud: For the moment we have no constructor based on ExperimentFunction.

PR #807, opened 2020-09-03
Add penalizations for constraints

teytaud:
## Types of changes

Constraints provided a number quantifying the violation are frequent, let us use this information.
This is just a beginning, I will play a lot with this during the forthcoming weeks.

## Checklist


teytaud:
To be tested on mltuning, yawidebbob, constrained_illconditioned_parallel
for finding a decent parametrization, before merging something cool.

nevergrad/optimization/optimizerlib.py
```diff
@@ -1869,6 +1869,68 @@ def __init__(self, parametrization: IntOrParameter, budget: tp.Optional[int] = N
         optim = self.optim if not isinstance(self.optim, NGO) else self.optim.optim
         logger.debug("%s selected %s optimizer.", *(x.name for x in (self, optim)))
 
+@registry.register
+class ShiwaC1(Shiwa):
+    def __init__(self, parametrization: IntOrParameter, budget: tp.Optional[int] = None, num_workers: int = 1) -> None:
+        super().__init__(parametrization, budget=budget, num_workers=num_workers) 
+        self.set_constraints_management(max_constraints_trials = 1000, constraint_penalization = 1., constraint_penalty_exponent = 1.0001,
+            penalize_cheap_violations = True, memorize_constraint_failures = False)
```
jrapin, line 1877: Please remove all this variants, this is a lot of duplicated code and would make maintenance too complicated.
You should create a ConstrainedNGOpt ConfiguredOptimizer, and then instantiate the ones you need in the experiments. No need to register them because they are too specific for being defined here.

nevergrad/parametrization/core.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, **parameters: tp.Any) -> None:
         # Additional convenient features
         self._random_state: tp.Optional[np.random.RandomState] = None  # lazy initialization
         self._generation = 0
-        self._constraint_checkers: tp.List[tp.Callable[[tp.Any], bool]] = []
+        self._constraint_checkers: tp.List[tp.Union[tp.Callable[[tp.Any], bool], tp.Callable[[tp.Any], float]]] = []
```
jrapin, line 41: would that work? `tp.List[tp.Callable[[tp.Any], tp.Union[bool, float]]]`
not sure though, but that would simplify the code

nevergrad/parametrization/core.py
```diff
@@ -276,9 +276,17 @@ def satisfies_constraints(self) -> bool:
...
+        def ok(x: tp.Union[bool, float]) -> bool:
+            return x >= 0. if isinstance(x, float) else x
+        return all(ok(func(val)) for func in self._constraint_checkers)
 
-    def register_cheap_constraint(self, func: tp.Callable[[tp.Any], bool]) -> None:
+    def penalty(self, num_ask: int, budget: int, exponent: float):
+        def float_penalty(x: tp.Union[bool, float]) -> float:
+            return 1 if x is False else 0 if x is True else -x if x < 0 else 0
+        val = self.value
+        return (exponent ** (num_ask / np.sqrt(budget))) * sum(float_penalty(func(val)) for func in self._constraint_checkers)
```
jrapin, line 287: this is way too linked to the optimization to be here, let's move this in the optimizer

nevergrad/optimization/base.py
```diff
@@ -398,7 +425,7 @@ def ask(self) -> p.Parameter:
             callback(self)
         current_num_ask = self.num_ask
         # tentatives if a cheap constraint is available
-        MAX_TENTATIVES = 1000
+        MAX_TENTATIVES = self._max_constraints_trials if self._max_constraints_trials else 1000
```
jrapin, line 428: just inititialize `self._max_constraints_trials` to 1000 and use it directly in the code (no need for `MAX_TENTATIVES`

nevergrad/optimization/base.py
```diff
@@ -164,6 +170,22 @@ def num_tell_not_asked(self) -> int:
         """
         return self._num_tell_not_asked
 
+    def set_constraints_management(self, max_constraints_trials: int, constraint_penalization: float, constraint_penalty_exponent: float,
```
jrapin, line 173: add defaults parameters to it, so that we can modify them one by one
Or better yet, default values in the init, and defaults to None in this function (and only non-None inputs will update the value)

nevergrad/optimization/base.py
```diff
@@ -164,6 +170,22 @@ def num_tell_not_asked(self) -> int:
...
+            penalize_cheap_violations: bool, memorize_constraint_failures: bool):
+        """We try max_constraints_trials random explorations for satisfying constraints.
+        The finally chosen point, if it does not satisfy constraints, is penalized as shown in the tell function.
+        """
```
jrapin, line 177: Please improve the docstring, in the standard format. And avoid `We`

jrapin (changes requested):
Also add information in register_constraint about the difference between floats and bools

nevergrad/benchmark/experiments.py
```diff
@@ -57,6 +57,7 @@ def mltuning(seed: tp.Optional[int] = None, overfitter: bool = False, seq: bool
               "DoubleFastGADiscreteOnePlusOne", "PSO", "BO", "MetaTuneRecentering"]
     if default_optims is not None:
         optims = default_optims
+    optims += ["Shiwa"] + ["ShiwaC" + str(i) for i in range(9)]
```
jrapin, line 60: you should use an optimizer group to create this list everywhere

nevergrad/parametrization/core.py
```diff
@@ -276,9 +276,17 @@ def satisfies_constraints(self) -> bool:
         if not self._constraint_checkers:
             return True
         val = self.value
-        return all(func(val) for func in self._constraint_checkers)
+        def ok(x: tp.Union[bool, float]) -> bool:
```
jrapin, line 279: add a comment below to explain why this is needed. I know why now, but eventually that informationn may get lost.

jrapin (commented):
Also, a test in test_optimizerlib.py where this is actually used would be useful

nevergrad/optimization/base.py
jrapin: I'm also not sure I would make this method public for now, since it may be unstable,
The docstring should state it.

nevergrad/benchmark/experiments.py
teytaud: This is going to be entirely deleted actually, only one of those ShiwaC should survive.

nevergrad/optimization/base.py
teytaud: Done, thx.

nevergrad/optimization/base.py
teytaud: Modified accordingly.

nevergrad/optimization/base.py
teytaud: Definitely right.

nevergrad/optimization/optimizerlib.py
teytaud: As mentioned in the comments, this is exactly the intent, this is here just for discussion until experiments conclude the debate.

nevergrad/parametrization/core.py
teytaud: Works!

nevergrad/parametrization/core.py
teytaud: Done.

nevergrad/parametrization/core.py
teytaud: Done.

teytaud:
I think https://dl.fbaipublicfiles.com/nevergrad/allxps/constrained_illconditioned_parallel/fight_all.png clearly shows that this works :-)

jrapin:
> I think https://dl.fbaipublicfiles.com/nevergrad/allxps/constrained_illconditioned_parallel/fight_all.png clearly shows that this works :-)

This is not a unit tests, it is not run by the CI everytime we commit, so we could end up breaking the feature without knowing it if we do not have a dedicated tests.
Also, in my experience of the many mistakes I make, results like these can be due to unexpected reasons, so I tend to prefer pinning down all features independently of everything else.

nevergrad/optimization/optimizerlib.py
jrapin: Then it should be in a separate PR that is not merged, anything merged is fair game and should be considered as a breaking changes if removed. So I definitely don't want to add something that will be removed right after.

nevergrad/optimization/base.py
```diff
@@ -165,6 +171,53 @@ def num_tell_not_asked(self) -> int:
         """
         return self._num_tell_not_asked
 
+    def penalty(self, candidate: p.Parameter):
+        def float_penalty(x: tp.Union[bool, float]) -> float:
+            """Unifies penalties as float (bool=False becomes 1)."""
+            return 1 if x is False else 0 if x is True else -x if x < 0 else 0
```
jrapin, line 177: please **never** use inline functions, this is causing issues when pickling. Although this is probably not important here, another problem is that it cannot be tested independently. In this case testing may definitely be required.
you can define it outside as `def _float_penalty`

nevergrad/optimization/base.py
```diff
@@ -165,6 +171,53 @@ def num_tell_not_asked(self) -> int:
         """
         return self._num_tell_not_asked
 
+    def penalty(self, candidate: p.Parameter):
```
jrapin, line 174: this should be hidden, please append with `_`.
I'll probably update this with a specific class anyway

nevergrad/optimization/base.py
```diff
@@ -165,6 +171,53 @@ def num_tell_not_asked(self) -> int:
...
+            """Unifies penalties as float (bool=False becomes 1)."""
+            return 1 if x is False else 0 if x is True else -x if x < 0 else 0
+        # Extract useful data from self.
+        val = candidate.value
+        budget = self.budget
+        num_ask = self._num_ask
+        exponent = self._constraint_penalty_exponent
+        return self._constraint_penalization * (exponent ** (num_ask / np.sqrt(budget))) * sum(float_penalty(func(val)) for func in candidate._constraint_checkers)
+
+    def set_constraints_management(self, max_constraints_trials: tp.Optional[int] = None,
```
jrapin, line 185: please hide it with a `_` for the time being, I'll update the API afterwards

jrapin (changes requested):
Please add a test in test_optimizerlib showing that this works.

nevergrad/optimization/base.py
teytaud: Done.

nevergrad/optimization/base.py
teytaud: Done.

nevergrad/optimization/base.py
```diff
@@ -339,6 +397,11 @@ def tell(self, candidate: p.Parameter, loss: tp.Loss) -> None:
             # multiobjective reference is not handled :s
             # but this allows obtaining both scalar and multiobjective loss (through losses)
             callback(self, candidate, loss)
+        if not candidate.satisfies_constraints() and self.budget is not None:
+            penalty = self.penalty_(candidate)
+            assert isinstance(loss, float)
```
jrapin, line 402: this is going to explode in multiobjective cases with constraints. What should be the "penalty" in this case?

jrapin:
@teytaud for simplicity, I am now wondering if the constraint should not be a positive only function:
- the meaning is clearer: this is a penalty and it's 0 if the constraint is satisfied.
- it can prevent errors: while one may confuse positive for negative (I do), with positive only penalty there is no confusion (if someone sends a negative, we trigger an explicit error).

Or does the positivity could bring anything for the optimizer? (for now it does not)

jrapin (approved):

PR #822, opened 2020-09-25
Long photonics runs -- WIP, not sure the best way for doing this

teytaud:
## Types of changes

## Checklist


nevergrad/benchmark/experiments.py
```diff
@@ -1297,6 +1298,33 @@ def photonics(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
                         yield xp
 
 
+@registry.register
+def big_photons(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
```
teytaud, line 1302: i'll store results in separate directories per date, so that we get multiple results for reruns...
the point in this function is that there is an aautomatic "suggest" based on the leaderboard.

jrapin: the runs will never finish, I'm still unconvinced by this xp, what is the aim?

teytaud: Each time we run the benchmarks, all algorithms start with the same information, i.e. the best so far.
Depending on the speed of machines, some runs will never finish if there is a walltime. 
If this runs on a devfair, this will actually finish as there is no wall time.
This does not prevent the leaderboard to be updated. 
What we should get is:
(1) run number 1: comparison between algorithms, possibly an update of the leaderboard.
(2) run number 2: comparison between algorithms, possibly an update of the leaderboard. The situation might be different from (1) as the starting point was not the same thanks to the get from leaderboard.

The goal is to have a continuous improvement: better results at each time we run the experiment.
This could be applied to other real world problems that really matter and are not here just for testing algorithms.

teytaud:
Deprecated

Issue #820, opened 2020-09-23
Feature Request: NSGA-II Optimizer


pacowong:
## Background

NSGA-II is a well known, fast sorting and elite multi-objective genetic algorithm. 

It is in the wishlist: https://github.com/facebookresearch/nevergrad/issues/761

## What to do
Let me be a volunteer. I'll add NSGA-II optimizer following the examples in DE and ES in Nevergrad.

As an alternative, we may consider adding an interface to jMetalPy. When I compare the ES implementation in jMetalPy ([here](https://github.com/jMetal/jMetalPy/blob/master/jmetal/algorithm/singleobjective/evolution_strategy.py)) and Nevergrad ([here](https://github.com/facebookresearch/nevergrad/blob/master/nevergrad/optimization/es.py)), my feeling is that it will be hard to maintain.

jrapin:
Since we are doing research with nevergrad, I tend to prefer when things are implemented internally, and that would be one fewer requirement ;)

If you are up for implementing it I am perfectly fine with it (and I expect @teytaud will be as well ;) ).
Still be careful, the multiobjective support within optimizer class is still a bit hacky and will probably evolve (I don't like that there is now both a `loss` and a `losses` in the `Parameter` class, but don't know how to do something simpler and generic enough for now). The only optimizer which actually use the multiobjective loss is currently `DE` indeed, so that should be the main example.

PR #823, opened 2020-09-27
NSGA-II implementation

pacowong:
## Types of changes

Related issue: https://github.com/facebookresearch/nevergrad/issues/820
Add the NSGA-II algorithm. The key procedures are the computation of the nondomination rank and crowding distance for the candidates. The initialization, crossover, and mutation procedures are borrowed from DE and ES in the current library.

## How Has This Been Tested 
1. The nondomination rank and crowding distance are tested by applying them on manually created candidates (i.e. p.Parameter variables), whose losses are preset.

2. The NSGA-II algorithm is tested on a multiobjective function, which is originally used to test CMA.

3. Test for correct population size as defined in test_optimizerlib.py

## Things to Review
1. In the current implementation, I'm not quite sure if I use "ask" and "tell" methods correctly.
2. Not sure why "_internal_tell_not_asked" will happen as in DE.
3. Is the custom "recommend" method in DE necessary? I found that ES does not implement "recommend".
4. Do I need to update the "heritage" variable for a candidate?

## Checklist


jrapin:
Hey,
Sorry for the delay.
I'm starting reading the code. The first thing that comes to mind is that it would be simpler if you put all the NSGA related code in a folder nevergrad/optimization/nsga, instead of having it in several folders and modules.
Then again, I'll need to reread, I am wondering if it wouldn't be simpler to update EvolutionStrategy in the case of multivariate optimization, instead of adding a whole new optimizer.

- In the current implementation, I'm not quite sure if I use "ask" and "tell" methods correctly.

Looks ok to me, I'll reread when I have a better overview.

- Not sure why "_internal_tell_not_asked" will happen as in DE.

That happens only when a user tells a candidate which has not been asked by the optimizer. In DE that has to be dealt with in a particular way because it must explicitely replace one element of the population, instead of just adding to the population like in ES.

- Is the custom "recommend" method in DE necessary? I found that ES does not implement "recommend".

It's only required for custom recommendation like the "noisy" case of DE. It shouldn't matter here.

- Do I need to update the "heritage" variable for a candidate?

I expect you should not have to deal with this, it's only necessary for very specific use cases which won't matter here.

I think I'll have more time for a second read beginning of next week, right now I'm a bit under the weather sorry :s

teytaud:
> if it wouldn't be simpler to update EvolutionStrategy in the case of multivariate optimization

you mean, having NSGA2 inheriting something from EvolutionStrategy ? not sure I understand your point, but I know for sure that NSGA2 is a specific, very well known algorithm for multiobjective optimization.

jrapin:
> you mean, having NSGA2 inheriting something from EvolutionStrategy ? not sure I understand your point, but I know for sure that NSGA2 is a specific, very well known algorithm for multiobjective optimization.

No I meant ES using the NSGA2 routine for selection.

Let's do it step by step:
 1. If I get it right NSGA2 is a ranking algorithm, not an optimizer, so let's focus first on this and we'll deal with optimizers later on.
 2. The whole algorithm can be defined as a function `List[Parameters] -> List[float]` right? Is there any reason for using a class for it? 
3. Since this method will be a tool for multiobjective, I think it should be in nevergrad/optimization/multiobjective, probably in a tools.py file or something like this. We'll move it around afterwards if need be.
4. It should be tested independently.

Just this should already be a PR on its own, and we can after interface it with optimizers as a selection step for instance. I think all the code is already there (except the tests), but correct me if I am wrong

pacowong:
2. As a contributor, I found it hard to modify DE or ES code to fit into EC framework, and many variables that I didn't really understand what they are. In my opinion, ES and DE (and NSGA) should share a EC base class, which clearly provides interface for evaluation, (parent) selection, crossover, mutation, replacement of individuals, and ranker. In the future, we can plug in different components in order to support more EC algorithms.

4. The test cases have been added for the ranker and NSGA (in test_optimizerlib.py). I want to use the same test cases that are used to test the multi-objective variants of ES or DE.

pacowong:
Let's reduce the mess in the codebase. I have reduced the amount of new code of NSGA-II. Now, NSGA2 is regarded as a selection operator for ES.

As for `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[float])`, do you mean this function should return 0 if a candidate is selected to survive to the next population and 1 otherwise? Because the key component of NSGA-II is a selection operator.

jrapin:
> Let's reduce the mess in the codebase. I have reduced the amount of new code of NSGA-II. Now, NSGA2 is regarded as a selection operator for ES.

Thanks for the simplification, I still think we need to go further so that NSGA-II can be easily applied with other optimizers. For now it is still tightly entwined with ES. Don't change it just yet though, I want to make sure I understand it better (through this thread) before that, in order to avoid useless work.

> As for `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[float])`, do you mean this function should return 0 if a candidate is selected to survive to the next population and 1 otherwise? Because the key component of NSGA-II is a selection operator.

That could be an option, but I feel we'd lose a lot of information in the process, and that would make it less general.
From what I see (again, I am not familiar with the algorithm), the output that are useful for selections are the rank and the crowding distance. The rank being the main selection criterion, and the crowding distance the secondary selection criterion.

So instead, I would tend to frame it as `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[tp.Tuple[int, float]])`, that is, for each parameter we return the rank and the crowding distance. With this, "user" optimizers would just have to call this function, then sort, and select the population they need. Am I correct? If we have this, I feel that modifying the ES algorithm is near to a oneliner (i.e. instead of sorting on loss, we'd sort on those values). If this is a one liner for ES, then that will be a one liner for other optimizers, and this is definitely what I am looking for.

jrapin:
Another way of phrasing it then, is what is the selection operator? if this is taking primarily lower ranks then secondarily the crowding distance, then those values are definitely what I want to extract, because after then we are a `sort` away from selection.

pacowong:
Thank you for the reply.

> So instead, I would tend to frame it as nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[tp.Tuple[int, float]]), that is, for each parameter we return the rank and the crowding distance. With this, "user" optimizers would just have to call this function, then sort, and select the population they need.

As you observed, there are two sorting procedures (let say SortA and SortB) involved in nsga2_ranking. Given a candidate, we can choose to output the following results:

1. Rank of the candidate after applying SortA
2. Fitness of the candidate under the distance function of SortA
3. Rank of the candidate after applying SortB
4. Fitness of the candidate under the distance function of SortB (after SortA is applied)
5. Final rank of the candidate 

Your proposal `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[tp.Tuple[int, float]])` meaning that you want to return the results 1 and 5 listed above. In general, a ranking function does not require to output the fitness value (i.e. `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[int])`, which should be the basic form. 

Since some ranking functions use 1 or more sorting criterion, we may provide another interface specifically for `tp.List[Parameters]) -> tp.List[tp.Tuple[int, float, float]])`. Do we really need these values? Once we decide what information to be kept, only minor modification is needed.

> With this, "user" optimizers would just have to call this function, then sort, and select the population they need

If we let "user" optimizer to decide what to do with the NSGA2 rank, this does not follow the description in the NSGA2 paper, which said selecting top K candidates. Possibly insert a note in the code to tell the users (mainly researchers) that "NSGA2 rank + top K selection = NSGA2 algo". 

Note also that top K selection leads to faster algorithm (not affecting the total time complexity though), hence more scalable for large population size, when we compare "assign the rank, then sort them all" approach using the proposed `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[tp.Tuple[int, float]])` interface.

jrapin:
> As you observed, there are two sorting procedures (let say SortA and SortB) involved in nsga2_ranking. Given a candidate, we can choose to output the following results:
> 
> 1. Rank of the candidate after applying SortA
> 2. Fitness of the candidate under the distance function of SortA
> 3. Rank of the candidate after applying SortB
> 4. Fitness of the candidate under the distance function of SortB (after SortA is applied)
> 5. Final rank of the candidate
>
> Your proposal `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[tp.Tuple[int, float]])` meaning that you want to return the results 1 and 5 listed above. In general, a ranking function does not require to output the fitness value (i.e. `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[int])`, which should be the basic form.
> 
> Since some ranking functions use 1 or more sorting criterion, we may provide another interface specifically for `tp.List[Parameters]) -> tp.List[tp.Tuple[int, float, float]])`. Do we really need these values? Once we decide what information to be kept, only minor modification is needed.

I must have something wrong, I see the "fast_non_dominant_rank" and the "crowding_distance" as the two values which allow sorting, no third one. Which one is that?
If we get a list (fast_non_dominant_rank, crowding_distance), we can easily sort later on either on the first value (non dominant rank) or directly on the tuples (which means first the rank, then on the crowding distance for equal ranks). Sorting on the crowding distance is not  on its own if i understand well.


> > With this, "user" optimizers would just have to call this function, then sort, and select the population they need
> 
> If we let "user" optimizer to decide what to do with the NSGA2 rank, this does not follow the description in the NSGA2 paper, which said selecting top K candidates. Possibly insert a note in the code to tell the users (mainly researchers) that "NSGA2 rank + top K selection = NSGA2 algo".

I'm always happy with notes in docstrings with as much information as possible :D 
On why I was considering the full computation: I am thinking of other use cases such as DifferentialEvolution in which case we compare 1 child to its parent, and not to the whole population, hence "top K selection" cannot be used, but NSGA2 rank can, since we need to make pairwise comparisons.


> Note also that top K selection leads to faster algorithm (not affecting the total time complexity though), hence more scalable for large population size, when we compare "assign the rank, then sort them all" approach using the proposed `nsga2_ranking(parameter: tp.List[Parameters]) -> tp.List[tp.Tuple[int, float]])` interface.

what you mean is that crowding_distance does not always need to be computed, is that it? That's indeed something I overlooked., maybe we can add `nsga2_ranking(parameter: tp.List[Parameters], n_selected: tp.Optional[int]) -> tp.Dict[str, tp.Tuple[int, float]])`? In other words:
- adding a `n_selected` parameters, by default all results are returned, otherwise only the `n_selected` best
- converting the output to a dict, with `parameter.uid` as key. With a list we would have had to return something even for the non-selected parameters, with a dict we get more flexibility.

In any case, such a function will probably evolve with time, but making sure it is very flexible will be helpful to test it and integrate it in multiple locations.

pacowong:
The new code should be flexible enough to be used in evolutionary algorithms.

nevergrad/optimization/es.py
```diff
@@ -64,13 +73,21 @@ def _internal_tell_candidate(self, candidate: p.Parameter, value: float) -> None
...
-                for uid in self._population:
-                    self._uid_queue.tell(uid)
+                self._simple_selection()
+
+
+    def _simple_selection(self):
+        choices = self._waiting + ([] if self._config.only_offsprings else list(self._population.values()))
+        if self._ranker is not None:
+            choices_rank = self._ranker.rank(choices, self._config.popsize)
+            choices.sort(key=lambda x: choices_rank[x.uid][0] if x.uid in choices_rank else float('inf'))
```
jrapin, line 83: why do we need to sort? this looks like a selection already, since it takes the popsize as input. The name should be updated accordingly

nevergrad/optimization/es.py
```diff
@@ -32,6 +33,13 @@ def __init__(
         self._waiting: tp.List[p.Parameter] = []
         # configuration
         self._config = EvolutionStrategy() if config is None else config
+        if self._config.ranker == "simple":
+            self._ranker = None
+        elif self._config.ranker == "nsga2":
+            self._ranker = rankers.NSGA2Ranking() #Note this does not control the selection mechanism
+        else:
+            raise NotImplementedError
```
jrapin, line 41:
```suggestion
        self._ranker: tp.Any = None   # TODO better typing (eventually)
        if self._config.ranker == "nsga2":
            self._ranker = rankers.NSGA2Ranking()
        elif self._config.ranker != "simple":
            raise NotImplementedError(f"Unknown ranker {self._config.ranker}")
```

Is the class adding anything to a mere function? Aka is there any sort of memoization? If not, then a function is much better, as it has a clearer API.

nevergrad/optimization/es.py
```diff
@@ -64,13 +73,21 @@ def _internal_tell_candidate(self, candidate: p.Parameter, value: float) -> None
...
-                choices.sort(key=lambda x: x._meta["value"])
-                self._population = {x.uid: x for x in choices[:self._config.popsize]}
-                self._uid_queue.clear()
-                self._waiting.clear()
-                for uid in self._population:
-                    self._uid_queue.tell(uid)
+                self._simple_selection()
+
+
+    def _simple_selection(self):
```
jrapin, line 79:
```suggestion
                self._select()


    def _select(self):
```

I put a lot of care to naming stuff precisely enough because it saves a lot of maintenance time.

jrapin (commented):
How will that work in non-multiobjective case? just like before?

jrapin:
It's getting way clearer in my opinion, thanks for the simplifications. I'd like just to clarify names and API, the clearer they are, the easier it will be to upgrade and maintain.

nevergrad/optimization/es.py
pacowong: Line 83 has been replaced by extracting the choices from the result choices_rank.

nevergrad/optimization/test_nsga.py
```diff
@@ -0,0 +1,38 @@
...
+    print("My Pareto front:", [x[0][0] for x in f.pareto_front()])
+
+    # It can also provide a subset:
+    print("My Pareto front:", [x[0][0] for x in f.pareto_front(2, subset="random")])
+    print("My Pareto front:", [x[0][0] for x in f.pareto_front(2, subset="loss-covering")])
+    print("My Pareto front:", [x[0][0] for x in f.pareto_front(2, subset="domain-covering")])
+    # DOC_MULTIOBJ_1
+    assert len(f.pareto_front()) > 1
+    assert len(f.pareto_front(2, "loss-covering")) == 2
+    assert len(f.pareto_front(2, "domain-covering")) == 2
```
jrapin, line 29: all this test is a copy paste of a test which is only aiming at providing documentation, it should not be duplicated. Also, it does not actually test the feature you have added because the MultiobjectiveFunction is actually converting multiobjective to scalar using a hypervolume (what you added should work for the new way of dealing with multiobjtive, without this MultiobjectiveFunction layer)
You can remove this file entirely, I can deal with adding a test later on.

nevergrad/optimization/evolution_ops/rankers.py
```diff
@@ -0,0 +1,224 @@
+import numpy as np
```
jrapin, line 1: I guess this file can be moved and renamed as nevergrad/optimization/multiobjective/nsga2.py to clarify things, and same thing for the corresponding test file (the one that is currently name test_nsga2.py should be removed, it's actually not testing the feature, see comment below)

jrapin (commented):
Still a few structural changes to simplify maintenance sorry (maintenance takes a huge toll so whatever simplifies the structure is a big win for me)
Also merge master and the mypy issues will just disappear.
I'll probably do a path later on to update a bit the API, but that can wait for later.

jrapin:
please merge the current master within your branch and we should be good to merge ;)

pacowong:
Done. Thank you for your review!

jrapin:
> Done. Thank you for your review!

sorry to bother again, can you fix the mypy error? The simpler option at this point is just a `# type: ignore` at the end of the line, no need for anything else (their is indeed an inconsistency, but that's because of nevergrad design, not sure yet how I'll update it)

nevergrad/optimization/multiobjective/nsga2.py
```diff
@@ -0,0 +1,224 @@
...
+                if objective_maxn - objective_minn == 0:
+                    pass #undefined
+                else:
+                    distance = distance / float(objective_maxn - objective_minn)
+                print(f"front[j]: {front[j].uid} distance: {distance}")
+                # The overall crowding-distance value is calculated as the sum of
+                # individual distance values corresponding to each objective.
+                front[j]._meta['crowding_distance'] += distance
+        else:
+            front = sorted(front, key=lambda x: x.loss)
```
jrapin, line 43:
```suggestion
            front = sorted(front, key=lambda x: x.loss)  # type: ignore
```

jrapin:
actually I could commit myself, hopefully that will be enough

jrapin:
All clear, thank you @pacowong

PR #630, opened 2020-04-27
Add arity descriptor in xp

jrapin:
## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] The documentation is up-to-date with the changes I made.
- [x] I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).
- [x] All tests passed, and additional code has been covered with new tests.

PR #828, opened 2020-09-30
[Open Optimization Competition 2020] - Nevergrad Performance Capture Benchmark

alexd314:
## Types of changes

## Motivation and Context 

This pull request regards the addition of a new benchmark to the nevergrad platform for the real-world noisy problem of 3D Human Performance Capture via fitting an animatable template 3D mesh to post-processed sensed 3D data.

The dataset consists of 3D captures of 11 human performances, recorded via the Multi-View Volumetric Capture System of [[1]]. For the purposes of this benchmark and for each performance, a 3D animatable template mesh has been created along with the selection of one individual target frame. The benchmark aims to assess the performance of nevergrad optimizers to the task of fitting the animatable template mesh (via its animation parameters) to the sensed 3D data of the selected performance’s target frame.

To realize the concept, along with this pull request, we release Windows binaries of a Performance Capture Benchmark Server (Perfcap benchmark server) which is able to serve Objective Function evaluations for arbitrary template mesh animation parameter values for each one of the designed experiments. To the nevergrad side, we implemented a python client to this benchmark server, which acts as an Objective Function proxy for every benchmarked optimizer.

I think this pull request is compatible with the latest version of nevegrad-master and benchmarking has been integrated natively, respecting nevergrad interfaces.
This code, mostly, contains newly added code compatible with the existing one. A small exception is that we added logging of the optimizer's recommendation value to the nevergrad logs as well as a naive mechanism to exclude fields from fight plots based on a naming convention.

While the current implementation supports single objective optimization, a future implementation of the python client may be able to support multi-objective optimization as well, without the need for updated binaries for the benchmark server.

A detailed documentation of this work (which will be continuously updated until our upcoming paper's submission) can be found in https://alexd314.github.io/nevergrad/.
A draft pre-liminary version of our paper can be found in https://drive.google.com/drive/folders/1vtCmmnM4BT-nuqpBMG3I2vZqibHcGWID?usp=sharing
Further, documentation of this work will be submited to GECCO 2021 for the purposes of the Open Optimization Competition 2020.

[1]: https://github.com/VCL3D/VolumetricCapture

## Things to check

Modification to existing code-base:
- `nevergrad/benchmark/frozenexperiments.py` (benchmark experiment registration)
- `nevergrad/benchmark/xpbase.py` (Added recommendation value logging in benchmark experiment under "{recommendation}" key)
- `nevergrad/benchmark/plotting.py` (Modified plotting.py to exclude description fields contained in "{}" from fight plots (Currently Performance Capture Benchmark adds an {experiment_tag_id} value to the description of the Experiment)).
Newly added code:
- `nevergrad/benchmark/perfcap/*`

For this benchmark the objective function, can be implicitly parameterized via json files found in `nevegrad/benchmark/perfcap/experiment_configuration`.

RabbitMQ connection details (for communication with the benchmark server) are currently being read from `rmqsettings.json` found in the `experiment_configuration` folder, since I did not want to introduce changes to the benchmark interface.

We have tested the submitted code by following the steps described in the run section of https://alexd314.github.io/nevergrad/.

## Checklist

- [X] I have read the [**CONTRIBUTING**](https://facebookresearch.github.io/nevergrad/contributing.html) document and completed the CLA (see [**CLA**](https://facebookresearch.github.io/nevergrad/contributing.html#contributor-license-agreement-cla)).

### DIFF
--- a/nevergrad/benchmark/frozenexperiments.py
+++ b/nevergrad/benchmark/frozenexperiments.py
@@ -12,8 +12,65 @@ from nevergrad.functions import ArtificialFunction
 from .xpbase import registry
 from .xpbase import create_seed_generator
 from .xpbase import Experiment
+from .perfcap.experiments import perfcap_experiment
+
 # pylint: disable=stop-iteration-return, too-many-nested-blocks
 
+
+@registry.register
+def perfcap_bench1(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment1.json", seedg)
+
+@registry.register
+def perfcap_bench2(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment2.json", seedg)
+
+@registry.register
+def perfcap_bench3(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment3.json", seedg)
+
+@registry.register
+def perfcap_bench4(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment4.json", seedg)
+
+@registry.register
+def perfcap_bench5(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment5.json", seedg)
+
+@registry.register
+def perfcap_bench6(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment6.json", seedg)
+
+@registry.register
+def perfcap_bench7(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment7.json", seedg)
+
+@registry.register
+def perfcap_bench8(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment8.json", seedg)
+
+@registry.register
+def perfcap_bench9(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment9.json", seedg)
+
+@registry.register
+def perfcap_bench10(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment10.json", seedg)
+
+@registry.register
+def perfcap_bench11(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment11.json", seedg)
 
 @registry.register
 def basic(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
--- a/nevergrad/benchmark/plotting.py
+++ b/nevergrad/benchmark/plotting.py
@@ -199,16 +199,19 @@ def create_plots(
     assert not any("Unnamed: " in x for x in df.columns), f"Remove the unnamed index column:  {df.columns}"
     assert "error " not in df.columns, f"Remove error rows before plotting"
     required = {"optimizer_name", "budget", "loss", "elapsed_time", "elapsed_budget"}
+    excluded = {x for x in set(df.columns) if x.startswith('{') and x.endswith('}')}  # ignore all descriptors starting enclosed in "{ }"
+
     missing = required - set(df.columns)
     assert not missing, f"Missing fields: {missing}"
     output_folder = Path(output_folder)
     os.makedirs(output_folder, exist_ok=True)
     # check which descriptors do vary
-    descriptors = sorted(set(df.columns) - (required | {"instrum_str", "seed", "pseudotime"}))  # all other columns are descriptors
+    # all other columns are descriptors
+    descriptors = sorted(set(df.columns) - (excluded | required | {"instrum_str", "seed", "pseudotime"}))
     to_drop = [x for x in descriptors if len(df.unique(x)) == 1]
     df = utils.Selector(df.loc[:, [x for x in df.columns if x not in to_drop]])
     # now those should be actual interesting descriptors
-    all_descriptors = sorted(set(df.columns) - (required | {"instrum_str", "seed", "pseudotime"}))
+    all_descriptors = sorted(set(df.columns) - (excluded | required | {"instrum_str", "seed", "pseudotime"}))
     print(f"Descriptors: {all_descriptors}")
     print("# Fight plots")
     #
@@ -245,6 +248,7 @@ def create_plots(
             if all([len(c) > 1 for c in df.unique(fixed)]):  # Let us try if data are adapted to competence maps.
                 # This is not always the case, as some attribute1/value1 + attribute2/value2 might be empty
                 # (typically when attribute1 and attribute2 are correlated).
+                # pylint:disable=consider-using-set-comprehension
                 try:
                     xindices = sorted(set([c[0] for c in df.unique(fixed)]))
                 except TypeError:
@@ -313,6 +317,7 @@ def create_plots(
             description = description[:140] + hash_ + description[-140:]
         out_filepath = output_folder / "xpresults{}{}.png".format("_" if description else "", description.replace(":", ""))
         data = XpPlotter.make_data(subdf)
+        # pylint: disable=broad-except
         try:
             xpplotter = XpPlotter(data, title=description, name_style=name_style, xaxis=xpaxis)
         except Exception as e:
@@ -499,6 +504,7 @@ def split_long_title(title: str) -> str:
     """
     if len(title) <= 60:
         return title
+    # pylint: disable=unnecessary-comprehension
     comma_indices = np.where(np.array([c for c in title]) == ",")[0]
     if not comma_indices.size:
         return title
--- a/nevergrad/benchmark/perfcap/core.py
+++ b/nevergrad/benchmark/perfcap/core.py
@@ -1,0 +1,460 @@
+#################################################################
+# Author: Alex Doumanoglou (al3x.doum@gmail.com / aldoum@iti.gr)
+# Information Technologies Institute - Visual Computing Lab (https://vcl.iti.gr)
+# 30 Sept 2020
+#################################################################
+import json
+import os
+import pathlib
+import typing as tp
+#import threading
+import logging
+import sys
+import traceback
+import itertools
+import copy
+from enum import Enum
+from uuid import uuid4 as uuid_gen
+from nevergrad.benchmark.perfcap.rmqservice import RMQService as _RMQService, RMQSettings as _RMQSettings
+import nevergrad as ng
+from nevergrad.benchmark.xpbase import Experiment
+from nevergrad.optimization import base as obase
+from nevergrad import functions as ngfuncs
+
+
+class Status(Enum):
+    OK = 0
+    Failed = 1
+
+
+class PerfcapExperiment(Experiment):
+
+    #pylint: disable=too-many-arguments
+    def __init__(self,
+                 experiment_manager: 'ExperimentManager',
+                 function: 'PerfcapFunction',
+                 optimizer: tp.Union[str, obase.ConfiguredOptimizer], budget: int, num_workers: int = 1,
+                 batch_mode: bool = True, seed: tp.Optional[int] = None,
+                 ) -> None:
+
+        kvps = {key: value for (key, value) in locals().items() if key not in ('experiment_manager', 'self', '__class__')}
+        super().__init__(**kvps)
+        self._experiment_manager = experiment_manager
+
+    def run(self) -> tp.Dict[str, tp.Any]:
+        try:
+            self._experiment_manager._start_experiment(self.function.experiment_tag_id)        # try to setup experiment
+        except Exception as e:  # pylint: disable=broad-except
+                                                            # experiment setup failed
+            self.result["error"] = e.__class__.__name__     # log error
+            print(f"Error when applying {self}:", file=sys.stderr)
+            traceback.print_exc()
+            print("\n", file=sys.stderr)
+            self._experiment_manager._stop_experiment(self.function.experiment_tag_id)     # disable rmq service if needed
+            return self.get_description()
+
+        ret = super().run()
+        # print the case and the traceback
+        # stop experiment. submit recommendation of the optimizer to Perfcap Benchmark Server in order for the server
+        # to dump the animated mesh in its mesh logs.
+        self._experiment_manager._stop_experiment(self.function.experiment_tag_id,
+                                                  self.function.compute_pose(*self.recommendation.value[0]))
+
+        return ret
+
+
+class VarNormalizer:
+    """
+    Normalizes parameter values to range [-1.0,1.0] based on bounds
+    This may improve optimizer convergence when scaling is different across parameters (i.e translation vs rotation)
+    Can be silently deactivated if normalize is set to False in the constructor.
+    """
+
+    def __init__(self, lower_bound: float, upper_bound: float, normalize: bool = False):
+        """
+        lower_bound: variable lower bound
+        upper_bound: variable upper bound
+        normalize: if False VarNormalizer is deactivated. If True, it will normalize input values in [-1.0,1.0
+        """
+        self._lower_bound = lower_bound
+        self._upper_bound = upper_bound
+        self._normalize = normalize
+
+    def denormalize(self, val: float) -> float:
+        """
+        Denormalize the parameter val
+        val: value to denormalize
+
+        Returns denormalized value
+        """
+        if not self._normalize:
+            return val
+        else:
+            return (val+1)*(self._upper_bound-self._lower_bound)/2.0 + self._lower_bound
+
+    def normalize(self, val: float) -> float:
+        """
+        Normalize the parameter val
+        val: value to normalize
+
+        Returns normalize value
+        """
+        if not self._normalize:
+            return val
+        else:
+            return 2.0 * (val - self.lower_bound) / (self._upper_bound - self._lower_bound) - 1.0
+
+    @property
+    def lower_bound(self):
+        return self._lower_bound
+
+    @property
+    def upper_bound(self):
+        return self._upper_bound
+
+    @property
+    def active_lower_bound(self):
+        """
+        effective lower variable bound
+        """
+        if self._normalize:
+            return -1.0
+        else:
+            return self._lower_bound
+
+    @property
+    def active_upper_bound(self):
+        """
+        effective upper variable bound
+        """
+
+        if self._normalize:
+            return 1.0
+        else:
+            return self._upper_bound
+
+    @property
+    def active(self):
+        """
+        Returns True, if VarNormalizer is effective
+        """
+        return self._normalize
+
+# pylint:disable=too-many-instance-attributes
+
+
+class PerfcapFunction(ngfuncs.ExperimentFunction):
+    """
+    Performance Capture Objective Function. This appears as a blocking call to the caller.
+    However, under the hood, it does a network request via ExperimentManager's RMQService to
+    Perfcap Benchmark Server to get the function evaluation at the specified point
+    """
+    # pylint:disable=too-many-locals
+
+    def _create_parameters(self, cfg: tp.Dict[str, tp.Any]) -> tp.List[ng.p.Scalar]:
+        """
+        Creates the parameters for the objective function based on configuration read from experimentX.json
+        """
+        variables = cfg["variables"]
+        jids = list(map(lambda x: x["joint_id"], variables))
+        jidset = set(jids)
+        assert len(jids) == len(jidset)
+        for v in variables:
+            # joint rotation exponential map has at most 3 degrees of freedom (DOF). Each DOF corresponds to one potential parameter.
+            # ["DOFs"] is a boolean list of length 3, indicating which one of the joint axis
+            # are considered for rotation. We can disable degrees of freedom per joint by
+            # inserting False to the related position index inside this list
+            # mutation_variances/bounds_min/bounds_max/mutable_mutation_variances
+            # are arrays of lenth equal to the number of DOFs of the joint
+            assert len(v["DOFs"]) == 3
+            nvars = sum(map(lambda x: 1 if x else 0, v["DOFs"]))
+
+            assert len(v["mutation_variances"]) == nvars
+            assert len(v["bounds_min"]) == nvars
+            assert len(v["bounds_max"]) == nvars
+            assert len(v["mutable_mutation_variances"]) == nvars
+
+        initial_pose = self._initial_pose
+        dofs_flattened = list(itertools.chain(*map(lambda x: x["DOFs"], variables)))
+        jids_flattened = list(itertools.chain(*map(lambda x: [x["joint_id"]]*len(list(filter(None, x["DOFs"]))), variables)))
+        #var_indices_flattend = list(np.cumsum(np.array(list(map(lambda x: 1 if x else 0, dofs_flattened))))-1)
+        var_indices_flattened = list(range(len(jids_flattened)))
+        dof_indices_flattened = list(itertools.chain(*map(lambda x: [index for index, v in enumerate(x["DOFs"]) if v], variables)))
+        self._variable_to_dof_index_map = dict(zip(var_indices_flattened, dof_indices_flattened))
+        self._variable_to_joint_index_map = dict(zip(var_indices_flattened, jids_flattened))
+        initial_values = list(itertools.compress(itertools.chain(*(initial_pose[jid] for jid in jids)), dofs_flattened))
+        # initial_values
+        #joint_vars = list(map(lambda x: sum(map(lambda y: 1 if y else 0, x["DOFs"])), variables))
+        bounds_min = list(itertools.chain(*map(lambda x: x["bounds_min"], variables)))
+        bounds_max = list(itertools.chain(*map(lambda x: x["bounds_max"], variables)))
+        mutation_variances = list(itertools.chain(*map(lambda x: x["mutation_variances"], variables)))
+        mutable_mut_var = list(itertools.chain(*map(lambda x: x["mutable_mutation_variances"], variables)))
+        # parameter normalizers, one for each parameter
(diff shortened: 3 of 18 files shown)


jrapin:
Hi
Thank you for your PR, this looks like a lot of work ;)
Before diving in (I'll need some time), can you explain why you need to subclass `Experiment` and what you mean by "communication with the benchmark server"? I am very uneasy at modifying this part of the code for one use case. Same thing goes for changing the plotting tools xpbase and plotting. Eg: logging the recommendation can be desastrous in very high dimension

alexd314:
Hi,

Thanks for your reply. 

**TL;DR** 
In case subclassing `Experiment` or modifying `xpbase` and `plotting` is prohibitive, this PR can be very easily refactored to drop these requirements at the cost of losing some logging capabilities.

**Detailed answer**

For this PR it is important to understand the underlying real-world problem of 3D performance capture. In this problem, given an animatable 3D human mesh template we try to fit it, via its pose parameters, to a 3D reconstructed human mesh (called `target` mesh), generated from multi-view depthmaps acquired via RGB-D sensors capturing a human performance. The objective function evaluation requires loading and 3D processing of the recorded depthmaps, computing a 3D distance field in the GPU, for efficient run-time, as well as multiple GPU-accelerated rendering operations among others. Implementing the objective function evaluation pipeline in python would be both impractical and costly in terms of development time and run-time performance, not to mention the increased code-base size which would be required to import in nevergrad. Thus, for this real-world problem we chose to develop an external application executable (called Performance Capture Benchmark Server) which is distributed in Windows binary form that servers objective function evaluations. Its interface is based on RabbitMQ messaging. On the nevergrad side we implemented the objective function (class `PerfcapFunction`) which talks to the Benchmark Server via RabbitMQ messaging with dependency on the `aio-pika` python package. `PerfcapFunction` respects the nevergrad interface and aims to hide the behind-the-scenes message talking to the benchmark server. In particular, `PerfcapFunction` looks to nevergrad as a standard Objective Function implementation with function evaluations being typical blocking calls, hiding the async messaging to the benchmark server.

In the current implementation and at the end of each experiment, the benchmark server is able to dump the fited animated template and the target 3D reconstructed human mesh to two .ply files ( a 3D mesh format which can be imported in popular 3D authoring tools like Meshlab, 3D Studio Max, Maya, Blender, etc) in order to allow visual inspection of the template’s fit to the target mesh. To realize this dump process, the benchmark server needs to know the time the experiment finishes along with the optimizer’s recommendation. Because the end of the experiment along with optimizer’s recommendation cannot be known to the `PerfcapFunction` alone, as its interface does not have any relevant API, subclassing `Experiment` served this purpose: To add communication in order to notify the benchmark server about the end of the experiment and flush the .ply dumps.

To synchronize logging between nevergrad logs and benchmark server .ply dumps, the `PerfcapFunction` appends an `{experiment_tag_id}` field to its description. The `.ply` files for the same experiment are dumped as `{experiment_tag_id}.ply` by the benchmark server. Thus, the relation between visual results and nevergrad logs is preserved. Since `{experiment_tag_id}` is a logging attribute we needed to exclude it from columns participating in the fight plots in `plotting.ply`. To make it generic, we introduced a scheme that if a column name is enclosed in brackets (“{}”) it is omitted from the fight plots. (changes in `plotting.py`)

Finally, in `xpbase.py` we added recommendation logging (under “{recommendation}” - column, inside “{}” to also omit from flight plots) for completeness. This modification has no major value since the recommendation is also communicated to the benchmark server and appears in server’s additional logs.

Conclusively, modifications in `xpbase.py` can be easily dropped. In case automatic association between an experiment’s log and its visualization to the real-world problem is not something of interest, we can also drop `{experiment_tag_id}` from `PerfcapFunction` and further omit subclassing `Experiment`, since `{experiment_tag_id}` is the only reason for subclassing it.

In case we omit `{experiment_tag_id}`, we could add benchmark server command line functionality to dump the .ply files after reading the optimizer’s  recommendation value from a text file. In this case, visualization would be an external process outside the benchmark framework and would require clients of nevergrad to write the optimizer’s recommendation to a specified text file (probably with us providing some helper functions for this purpose).

**However**, in this real-world problem (and probably in other real-world problems), the success of the solution found by an optimizer cannot be solely assessed by the objective function value at the recommended point. This stems from the fact that, in our case, the objective function is an **approximation** to another metric (hausdorff distance) which is costly to evaluate. From a real-world perspective, the quality of the solution though, is eventually measured by Hausdorff distance and not by the objective function value at the recommendation point. This is common to most problems using a proxy/surrogate loss function, instead of the final objective for various reasons. Thus, the aforementioned logging functionality and visual inspection of the solution may be of an important value to real-world practitioners. This applies to real-world problems needing to assess the resulting recommendation via a metric other than the objective function value.

We do not argue that it has to be done the way we propose in this PR, but in general, would you consider a generic mechanism to log the optimizer’s recommendation value in the nevegrad’s benchmark framework ? In case yes, then we would be happy to contribute.

teytaud:
There is a system for logging the recommended value, this is "update_leaderboard". This is for information, I'm not saying that your system should not be merged in Nevergrad, Jeremy is the right person for knowing this :-)

teytaud:
aio_pika is missing in the dependencies ?
you might fix that in requirements/bench.txt or in .circleCI/config maybe ?

teytaud:
I guess the remaining errors are not a big deal, just differences in the text in the test of some command lines:

    def test_commandline_launch() -> None:
        with tempfile.TemporaryDirectory() as folder:
            output = Path(folder) / "benchmark_launch_test.csv"
            # commandline test
            # TODO make it work on Windows!
            # TODO make it work again on the CI (Linux), this started failing with #630 for no reason
            with testing.skip_error_on_systems(FailedJobError, systems=("Windows", "Linux")):
                CommandFunction(command=[sys.executable, "-m", "nevergrad.benchmark", "additional_experiment",
                                         "--cap_index", "2", "--num_workers", "2", "--output", str(output),
                                         "--imports", str(Path(__file__).parent / "additional" / "example.py")])()
            assert output.exists()
            df = utils.Selector.read_csv(str(output))
>           testing.assert_set_equal(df.columns, DESCRIPTION_KEYS | {"offset"})  # "offset" comes from the custom function

nevergrad/benchmark/test_core.py:54: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

jrapin:
Sorry for taking so much time to read this, I have been on too many stuff at once and delayed this long reading a bit too much :s This should be better now.

My current thoughts (for discussions, this is not definitive):
1. OK for the `{}` pattern in the log, this can be useful, but we'll simplify it a bit (I'll explain later on in a review, no need to focus on this now)
2. "Because the end of the experiment along with optimizer’s recommendation cannot be known to the PerfcapFunction alone, as its interface does not have any relevant API"
  actually it does, it's the `evaluation_function` (probably not very well advertised in the doc sorry :s) https://github.com/facebookresearch/nevergrad/blob/bcf82bd125687b1d09bf3b8329c3be238ab86fea/nevergrad/functions/base.py#L156-L166
  With this you can do anything you want at the very end of the optimization, I've used it to log the results for [some functions](https://github.com/facebookresearch/nevergrad/blob/bcf82bd125687b1d09bf3b8329c3be238ab86fea/nevergrad/functions/photonics/core.py#L128-L133). Would that help avoid to subclass `Experiment`? Subclassing `Experiment` could have a huge cost for us in term of maintenance on the long run, so I am definitely against it, but it should be able to support most if not all requirements you can have.
3. With both the above points, I would tend to write the uid to nevergrad logs with `{}`, then write everything else (including recommendation, since it cannot reliably be dumped for all benchmarks without creating size problems) to dedicated files created by the `ExperimentFunction`. Would that work?
4. I have a concern for testing. If this runs only on Windows that will prevent a lot of the testing to be accurately done. Whatever we cannot test especially locally may end up either broken or take us too much time to maintain. We'd need to have mocked tests when possible, but even with this I cannot guarantee that we'll be able to maintain the code :s 
5. Is there any license related to the files that are added? If so we need to deal with this as well. One option would be to import them on the run from an appropriate repo.

Once everything is clarified, I'll provide a more detailled review to move on (which may be a bit painful :s sorry although I don't expect it to be difficult, except for the mocking tests part :s)

alexd314:
Hello,

Thanks for your reply. Regarding (1), (2) and (3) I think I can manage to adapt the codebase rather easily. I haven't done a "hands on" investigation yet, but I am going to draft a preliminary implementation conforming to your suggestions and I can notify you about the updated code when it is ready or in case I find a difficulty (probably in short time from now).

Regarding (4),  I definitely understand the difficulties of automated testing and maintenance. How would you envision testing a real-world data driven benchmark contribution ? What are your current thoughts? I am asking in order to better understand your viewpoint and make proper suggestions.

For (5), are you referring to the files included in this PR or the externally provided binaries for the benchmark server ? In any case, files in this PR do not have any license.

jrapin:
> Thanks for your reply. Regarding (1), (2) and (3) I think I can manage to adapt the codebase rather easily. I haven't done a "hands on" investigation yet, but I am going to draft a preliminary implementation conforming to your suggestions and I can notify you about the updated code when it is ready or in case I find a difficulty (probably in short time from now).

Great, while you're at it please move perfcap to nevergrad/functions. Also since we are getting more and more benchmarks, would there be a clearer name? performancecapture of perfcapture3d or something in these lines would be clearer in my opinion, but that's if the name is not "standard" already.

> Regarding (4), I definitely understand the difficulties of automated testing and maintenance. How would you envision testing a real-world data driven benchmark contribution ? What are your current thoughts? I am asking in order to better understand your viewpoint and make proper suggestions.

I don't have enough visibililty for now, but I would start by isolating as much as possible whatever cannot be run easily into 1 function with a simple enough input and simple enough output that can be mocked during the tests. The smaller this function, the better...

> For (5), are you referring to the files included in this PR or the externally provided binaries for the benchmark server ? In any case, files in this PR do not have any license.

both, where do they come from?

jrapin:
I can't see any binary in the PR (that's better this way actually, I don't think I would add a binary here if that's feasible to get it any other way), nor where it is supposed to be used, can you point to it?

alexd314:
Hello,

I have refactored the code-base based on your feedback. Indeed using `ExperimentFunction`'s `evaluation_function` method I was able to accomplish exactly was I intended. More specifically,
1. I no longer subclass `Experiment`, I only subclass `ExperimentFunction`
2. I left the "{ }" trick for excluding fields in the fight plots, untouched. I know you already mentioned that we should change that but since you said that we can leave this for later, I did not touch it, yet
3. I removed logging of recommendation value (i.e reverted back all changes to `nevergrad/benchmark/xpbase.py`)
4. I moved most of the code-base to `nevergrad/functions` as you suggested. The only part of this PR that is not in `nevergrad/functions` is the code edited in `frozenexperiments.py` in order to register experiments to the benchmark registry.

The part of this PR that has dependency in nevergrad is isolated in `Perfcap3DFunction` which subclasses `ExperimentFunction` and uses `nevergrad.parameterization.parameter.Scalar`. These are the only dependencies to nevergrad in this PR (unless I am forgetting something now, which I think I don't). Other future change that might break things, could be either a python language specification change, or a python native library interface change or aio_pika package upgrades (and updates to interfaces). This information maybe useful when we will consider mocking and I am just mentioning them here just to initiate the discussion.

Regarding the binary (ie. the benchmark server itself) can be downloaded from [here](https://github.com/VCL3D/PerformanceCapture/releases). Currently it does not come with any License either.

nevergrad/benchmark/frozenexperiments.py
```diff
@@ -4,16 +4,92 @@
 # LICENSE file in the root directory of this source tree.
 
 import typing as tp
+from uuid import uuid4 as uuid_gen
 import numpy as np
 from nevergrad import optimizers
 from nevergrad.optimization.base import ConfiguredOptimizer
 from nevergrad.optimization import experimentalvariants  # pylint: disable=unused-import
 from nevergrad.functions import ArtificialFunction
+from nevergrad.functions.perfcap3d.core import Perfcap3DExperimentConfig, Perfcap3DFunction, Perfcap3DServerComm
```
jrapin, line 13: for simplicity it would be better to only do `from nevergrad.functions import perfcap3d` since all the casses are imported in the `__init__.py`

Also, are all these classes necessary? It seems too complex for me from a user perspective, wouldnt it work to have `Perfcap3DFunction` take an `experiment_filename` and then set up all the necessary classes itself?

nevergrad/benchmark/frozenexperiments.py
```diff
@@ -4,16 +4,92 @@
...
+    # pylint: disable=stop-iteration-return
+    for budget in budgets:
+        for optim in optimizer_names:
+            yield Experiment(function=Perfcap3DFunction(server_comm, cfg.objective_function_config, str(uuid_gen())),
+                                optimizer=optim, budget=budget, seed=next(seedg))
+
+@registry.register
+def perfcap_bench1(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    seedg = create_seed_generator(seed)
+    return perfcap_experiment("experiment1.json", seedg)
```
jrapin, line 42: I tend to really hate repeated code (it's basically a maintenance hell in the long run) :D
So we'd need something else for this. Since benchmarks are only accessed through the register, we can probably convert all this into a for loop that does the registration over some object. Not sure exactly how yet though so I may end up doing it later on.
Another option would be to define all the experiments in one, do you run all them independently? I feel you would usually want to try them all together. If the experiment number is provided as a descriptor of the Experiment, then the plotting tools will differenciate between them.

nevergrad/benchmark/plotting.py
```diff
@@ -199,16 +199,19 @@ def create_plots(
     assert not any("Unnamed: " in x for x in df.columns), f"Remove the unnamed index column:  {df.columns}"
     assert "error " not in df.columns, f"Remove error rows before plotting"
     required = {"optimizer_name", "budget", "loss", "elapsed_time", "elapsed_budget"}
+    excluded = {x for x in set(df.columns) if x.startswith('{') and x.endswith('}')}  # ignore all descriptors starting enclosed in "{ }"
```
jrapin, line 202: instead of all these modifications, I would prefer you the columns directly from the dataframe as the first or second line of the function:
`df = df.loc[:, [x for x in df.columns if not x.startswith('{') and x.endswith('}')]`
This way the impact is really reduced, which is always better :D

nevergrad/functions/perfcap3d/core.py
```diff
@@ -0,0 +1,436 @@
...
+from uuid import uuid4 as uuid_gen
+from nevergrad.parametrization import parameter as p
+from .rmqservice import RMQService as _RMQService, RMQSettings as _RMQSettings
+from ...functions.base import ExperimentFunction
+
+class Status(Enum):
+    OK = 0
+    Failed = 1
+
+class VarNormalizer:
```
jrapin, line 18: cant these constraints be integrated within the parametrization? It's better if all this is handled in one place

nevergrad/functions/perfcap3d/core.py
```diff
@@ -0,0 +1,436 @@
...
+        # parameter creation (initial value, bounds, mutable_mutation_var)
+        params = [p.Scalar(init=n.normalize(init), lower=n.active_lower_bound, upper=n.active_upper_bound, mutable_sigma=ms)
+             for (n, ms, init) in zip(self._var_normalizers, mutable_mut_var, initial_values)]
+        # leave negative mutation variances to auto / ie. do not set
+        for (v, mv, n) in filter(lambda x: x[1] > 0.0, zip(params, mutation_variances, self._var_normalizers)):
+            v.set_mutation(n.normalize(mv), custom="gaussian")      # set variable mutation
+
+        return params
+
+    def __init__(self, server_comm: 'Perfcap3DServerComm', objective_func_cfg: tp.Dict[str, tp.Any], experiment_tag_id: str):
```
jrapin, line 157: it does look to me like this could take a mere index of the experiment to use, and then everything could be set up automatically. Can you detail why not?

nevergrad/functions/perfcap3d/core.py
```diff
@@ -0,0 +1,436 @@
...
+        """
+        return self._normalize
+
+# pylint:disable=too-many-instance-attributes
+
+class Perfcap3DFunction(ExperimentFunction):
+    """
+    Performance Capture Objective Function. This appears as a blocking call to the caller.
+    However, under the hood, it does a network request via Perfcap3DServerComm's RMQService to
+    the Perfcap Benchmark Server to get the function evaluation at the specified point
```
jrapin, line 101: It there are stuff to install to make this work, the docstring should explain it.
If the download of the executable could be automatized, it could be even better

nevergrad/functions/perfcap3d/core.py
```diff
@@ -0,0 +1,436 @@
+import json
+import os
+import pathlib
+import typing as tp
+import logging
+import itertools
+import copy
+from enum import Enum
+from uuid import uuid4 as uuid_gen
+from nevergrad.parametrization import parameter as p
```
jrapin, line 10: throughout the code we now preferably use:
`import nevergrad as ng`
then `ng.p` when needed

jrapin:
I am not yet clear where the executable is used, i don't see it being called in the called, what have I missed?
A test should basically mock this executable, so that all the code but it is actually tested, are you familiar with mocking?

nevergrad/functions/perfcap3d/core.py
alexd314: I am not sure about this. Can you point me to the correct direction ?
VarNormalizer normalizes domain to -1,1 in order to cope with optimizers that may have scaling issues wrt to parameters

nevergrad/benchmark/plotting.py
alexd314: Done

nevergrad/benchmark/frozenexperiments.py
alexd314: I added a dynamic way to register experiments. (dynexpreg.py)

nevergrad/benchmark/plotting.py
alexd314: Done

nevergrad/benchmark/frozenexperiments.py
alexd314: I added a simpler interface in this new version

nevergrad/functions/perfcap3d/core.py
alexd314: I followed a similar way. Initially it takes a filename and follows configuration from there.

nevergrad/functions/perfcap3d/core.py
alexd314: Done

nevergrad/functions/perfcap3d/core.py
alexd314: I added automated downloading and unzipping of executable and data

alexd314: I added automated downloading, unzipping of executable and data, and execution of server on benchmark start

alexd314:
Hi,

I have update the code base based on your comments. Basically, i added dynamic registration of experiments and automated downloading and execution of benchmark server executable and data. Interface has been simplified and other minor comments have also been followed according to your proposals.

Kind regards,
Alex

jrapin:
Thanks for the updates, I've started reviewing this but a few things make it difficult:
- can you please merge master
- add the missing packages, for what I have seen you need at least to add `py7zr>=0.11.3` and `multivolumefile>=0.1.2` or else nothing will work
- please try to add a few basic test on the function, mocking what is missing, without this it's a bit hard to figure out if it's working or not. I am currently trying to run the test suit but not sure if it will work at all, because if I understood correctly it's Windows only.
- is it useful to run each test case independently from the other ones? I would rather have just one test case which aggregates on all the different cases. For now I have tried to avoid code doing dynamic registration as you did, so having only one xp to register would be better.

Also, automatic download is great, but we should not rely on the internet for the tests, so we will probably need to deactivate part of the automatic tests for these experiments (making it even more important to be able to have independent tests for the function). Not sure what will be the best option for now, I'll need to try when the above is solved.

jrapin:
After about 1h of running the tests (which is way too long :s before this PR the whole test suit was taking 15min, which is already way too long :s) , it crashed with `ValueError: File is not a recognized excel file` :s (on Mac though)

alexd314:
Hello,

Thank you for your efforts to review the PR. I understand about the difficulties. Let's see if we can make something that meets all your requirements. Below is my feedback to your comments:

> 
> 
> Thanks for the updates, I've started reviewing this but a few things make it difficult:
> 
>     * can you please merge master
Done.
> 
>     * run `black nevergrad` for the code formatting (`pip install black` if you don't have it)
> 
Done.

>     * add the missing packages, for what I have seen you need at least to add `py7zr>=0.11.3` and `multivolumefile>=0.1.2` or else nothing will work
> 
Done.

>     * please try to add a few basic test on the function, mocking what is missing, without this it's a bit hard to figure out if it's working or not. I am currently trying to run the test suit but not sure if it will work at all, because if I understood correctly it's Windows only.

I am not sure I understand what you mean by "missing". Could you please explain more? Do you mean about the requirement to download and connect to the executable and data ?

> 
>     * is it useful to run each test case independently from the other ones? I would rather have just one test case which aggregates on all the different cases. For now I have tried to avoid code doing dynamic registration as you did, so having only one xp to register would be better.
> 

Currently we have 11 different cases named 'perfcap_bench1" to "perfcap_bench11". What exactly do you mean run each test case independently from the other ones? Currently we actually have 11 different cases running independently. Maybe I there is something that I did not fully understand. 

Concerning the dynamic registration, I'd be open to suggestions. That was a quick way that I could think of, in order to avoid duplicate code.

> 
> Also, automatic download is great, but we should not rely on the internet for the tests, so we will probably need to deactivate part of the automatic tests for these experiments (making it even more important to be able to have independent tests for the function). Not sure what will be the best option for now, I'll need to try when the above is solved.
> 

Do you have already some logic integrated regarding this matter (ie. code path for no internet connection)?
What exactly do you mean independent tests for the function ? Do you mean skipping the connection to the actual benchmark server and serving dummy evaluations ?


> This will all take some time sorry :s most PRs already require a lot of back and forth, but PRs which I cannot run easily may require even more :s

I understand. Let's see if we make it work.

alexd314:
> 
> 
> After about 1h of running the tests (which is way too long :s before this PR the whole test suit was taking 15min, which is already way too long :s) , it crashed with `ValueError: File is not a recognized excel file` :s (on Mac though)

Hmm. I've never tried on Mac and this 'Value Error' for excel file I am sure I've never run into it. We need to investigate. Regarding the run time of the tests, they are actually very time consuming, and certainly depend on the number of repetitions. Currently, the budgets and optimizers evaluated are based on our paper submission regarding this benchmark, to GECCO '21. We kept the configuration the same as the one we used for our paper, for reproducibility. However, we can be flexible on this, if this is not very desirable for the purposes of this PR. Let's discuss further.

jrapin:
> Currently we have 11 different cases named 'perfcap_bench1" to "perfcap_bench11". What exactly do you mean run each test case independently from the other ones? Currently we actually have 11 different cases running independently. Maybe I there is something that I did not fully understand.

One "experiment" function in nevergrad is usually a set of different test cases, so that everything is run, and results can be both aggregated and split upon different cases (the framework recognizes when the ExperimentFunction used different arguments). So instead of 11 registered experiments, I would just use "perfcap_experiment" and loop over the different arguments (= over `experiment_filename`) within the function. Also, if we do this, then no more need of dynamic registration.

> I am not sure I understand what you mean by "missing". Could you please explain more? Do you mean about the requirement to download and connect to the executable and data ?
Do you have already some logic integrated regarding this matter (ie. code path for no internet connection)?
Do you mean skipping the connection to the actual benchmark server and serving dummy evaluations ?

Exactly, we must not expect the tests to download GB of data, nor run something that takes a full hour, so during the tests we need to mock the part of your code that does the downloading and the computation, and replace it with a dummy. Eg (from https://docs.python.org/3/library/unittest.mock.html#the-patchers):
```python
>>> from io import StringIO
>>> def foo():
...     print('Something')
...
>>> @patch('sys.stdout', new_callable=StringIO)
... def test(mock_stdout):
...     foo()
...     assert mock_stdout.getvalue() == 'Something\n'
```
I've done this on several parts of the code, the closest one to your issue (avoiding downloading data) is this one:
https://github.com/facebookresearch/nevergrad/blob/ddcaec01126f5c3f28de2b6d1a69523d0cf05183/nevergrad/functions/mlda/datasets.py#L84-L88

Of course one could mock the whole ExperimentFunction, but then we would just test nothing so it would be pointless, the aim is to be figure out the deepest possible function that we can mock.


> What exactly do you mean independent tests for the function ? 

Currently your code is only called in the tests that tries to run parts of all the benchmarks, there is no test that just tries to call your function with some arguments and just check that it does not raise an exception. It may look stupid but actually that's really useful to make sure nothing is going completely wrong on the long term (I need this for maintenance)
Eg: something as simple as this:
https://github.com/facebookresearch/nevergrad/blob/ddcaec01126f5c3f28de2b6d1a69523d0cf05183/nevergrad/functions/rocket/test_core.py#L10-L14

> Hmm. I've never tried on Mac and this 'Value Error' for excel file I am sure I've never run into it. We need to investigate. Regarding the run time of the tests, they are actually very time consuming, and certainly depend on the number of repetitions. Currently, the budgets and optimizers evaluated are based on our paper submission regarding this benchmark, to GECCO '21. We kept the configuration the same as the one we used for our paper, for reproducibility. However, we can be flexible on this, if this is not very desirable for the purposes of this PR. Let's discuss further.

I think if it works for you, and as long as the unit tests don't get in the way of development, we'll merge what is the closest to your submission, and nevermind if it is not working on Mac (again, as long as it does not get in the way for all the unit tests etc). The issue tough is that if at some point it requires maintenance, I may just completely break it or need to remove it because I won't be able to test it.

nevergrad/benchmark/frozenexperiments.py
```diff
@@ -9,13 +9,53 @@
...
+    # optimizer_names = ["Shiwa","RandomSearch"]
+
+    total_experiment_count = len(budgets) * len(optimizer_names)
+    perfcap3d.Perfcap3DServerExecutor.execute_server(total_experiment_count)
+
+    # pylint: disable=stop-iteration-return
+    for budget in budgets:
+        for optim in optimizer_names:
+            yield Experiment(
+                function=perfcap3d.Perfcap3DFunction(experiment_filename),
```
jrapin, line 52: Implementation detail, please create the function in the outermost loop, it makes it much faster because the creation of the function can be costly compared to the rest of the loop.
In this case that means creating the function before the for loops, or if we merge the test cases into one big experiments, it means added `for experiment_filename in ....` as the outermost for loop, and create the corresponding function right after

nevergrad/benchmark/frozenexperiments.py
```diff
@@ -9,13 +9,53 @@
 from nevergrad.optimization.base import ConfiguredOptimizer
 from nevergrad.optimization import experimentalvariants  # pylint: disable=unused-import
 from nevergrad.functions import ArtificialFunction
+from nevergrad.functions.perfcap3d import core as perfcap3d
```
jrapin, line 12:
```suggestion
from nevergrad.functions import perfcap3d
```
then in `perfcap.__init__.py` do: 
```python
from core import Perfcap3DServerExecutor as Perfcap3DServerExecutor
from core import Perfcap3DFunction as Perfcap3DFunction
```

Issue #849, opened 2020-10-09
adding realistic pyomo models in nevergrad


teytaud:
Several interesting Pyomo models that we might want to integrate in Nevergrad:

PowNet-Cambodia – implementation of PowNet for Cambodian power system with data for 2016

PowNet-Laos – implementation of PowNet for Laotian power system with data for 2016

PowNet-Thailand – implementation of PowNet for Thai power system with data for 2016

Network flow optimization of California's water supply system: https://github.com/ucd-cws/calvin

teytaud:
All of them open source on internet.
thx @pacowong  for the discussion :-)

PR #850, opened 2020-10-10
[WIP] PowNet Pyomo Models

pacowong:
## Types of changes

To address the issue in https://github.com/facebookresearch/nevergrad/issues/849. I tried to implement apply Nevergrad to solve PowNet problem.

## How Has This Been Tested
The number of variables and constraints in the time-series model (https://github.com/kamal0013/PowNet) is huge. Based on the model, I have manually simplified it a bit and produce the `pownet_tiny` model. But still, I cannot test the implementation locally. And it seems to defeat our purpose for being realistic. Any suggestions? Which solvers should I use for this kind of problem?

## Checklist


jrapin:
I would rather we do not copy code from other repositories when possible (and if ever we do, we need to add their license). Also adding big files is really not practical, I'd like to keep nevergrad as small as possible in size (it's probably already a bit too big because of data files here and there :s).
So, is it possible to interface with the original code if it is installed (not adding it as a dependence) instead of duplicating all this code?

As for solvers, I have not enough understanding of the problem here, maybe @teytaud knows more?

pacowong:
Such a large amount of code in `pownet_model.py` is actually used for model definition. I will find ways to let the script automatically download `pownet_model.py` from the original repo when users invoke the function. Just like people download pre-trained models for deep neural networks. Therefore, it won't in the repo when the code is completed. No worry :)

Alternatively, may I just submit the pull request to a fork of Nevergrad for benchmarking purpose?

@teytaud any suggestion for solver?

requirements/bench.txt
```diff
@@ -14,3 +14,4 @@ torchvision
 pyomo>=5.7
 IOHexperimenter>=0.2.8
 mixsimulator>=0.2.9.5
+git+git://github.com/pacowong/pypownet@master#egg=pypownetr
```
jrapin, line 17: that ends up being a pain to install with the CI, this should not be added here. The best option would be to import inline and provide an explicit error message if not importabe.

nevergrad/functions/pyomo/core.py
```diff
@@ -148,6 +148,7 @@ def _pyomo_value_assignment(self, k_model_variables: tp.Dict[str, tp.Any]) -> No
         if self._value_assignment_code_obj == "":
             code_str = ""
             for k in k_model_variables:
+                #k_name = k.replace("'", "\\'")
```
jrapin, line 151:
```suggestion
```

nevergrad/functions/pownet/solver.py
```diff
@@ -0,0 +1,251 @@
...
+from datetime import datetime
+import nevergrad as ng
+from nevergrad.functions.pyomo import core
+
+import git
+import pypownetr.model
+
+def download_dataset(dataset_name="cambodian", force_update=False):
+    #git clone <repository> <path>
+    dataset_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'pypownet')
```
jrapin, line 16: please use `form pathlib import Path`  and use the Path object everywhere instead of `os.path`, this will make the code way more readable

nevergrad/functions/pownet/solver.py
```diff
@@ -0,0 +1,251 @@
+import os
```
jrapin, line 1: if this is all based on Pyomo, you should inclide it in the pyomo directory, that will help structure the code.

nevergrad/functions/pownet/solver.py
```diff
@@ -0,0 +1,251 @@
...
+    wind = []
+
+    hydro_import = []
+
+    srsv = []
+    nrsv = []
+
+    vlt_angle = []
+
+    system_cost = []
```
jrapin, line 60: this is a whole lot of variables :s sorry but I'll ask you to add type hints for them because knowing what they should be will be helpful for maintenance on the long run.

nevergrad/functions/pownet/solver.py
```diff
@@ -0,0 +1,251 @@
...
+            
+            for z in instance.h_imports:
+                # Load Hydropower time series data
+                for i in K:
+                    instance.HorizonHydroImport[z,i] = instance.SimHydroImport[z,(day-1)*24+i]     
+        
+        func = core.Pyomo(instance)
+        optimizer = ng.optimizers.CMA(parametrization=func.parametrization, budget=10)
+        recommendation = optimizer.minimize(func.function)
+        print(recommendation) #Low system_cost is better
```
jrapin, line 99: I would avoid prints in the code (and commented section, but then again it's WIP is that it?)

jrapin (commented):
I am not clear how this is supposed to be integrated in the benchmarks, can you clarify?

pacowong:
@jrapin I face a challenge when using Nevergrad to solve the problem. Since this model involves a large number of constraints, I don't know which optimizer can produce viable solutions during initialization. Should I implement an alternative Pyomo converter, which replaces `register_cheap_constraint` by penalization? Just like what I did in https://github.com/facebookresearch/nevergrad/pull/894?

jrapin:
> @jrapin I face a challenge when using Nevergrad to solve the problem. Since this model involves a large number of constraints, I don't know which optimizer can produce viable solutions during initialization. Should I implement an alternative Pyomo converter, which replaces `register_cheap_constraint` by penalization? Just like what I did in #894?

If you are not in a hurry I would wait for #807 to be merged, that should provide a more long term solution to your problem. I'll need to spend some time on #807 before though, probably next week.  This API will also need to evolve a bit even after the merge, so if you want to use it, the code will need to be easy to update.

pacowong:
Good to know. I can wait until Nevergrad resolves the issue.

teytaud:
> Good to know. I can wait until Nevergrad resolves the issue.

IMHO do not modify the problem! Maybe the current developments on constraints have solved it or will solve it (there is another PR in progress :-) ); and a problem that we do not solve correctly is a good piece of news, this means that we know how to make progress. We do need difficult important problems with constraints for benchmarking and improving our algorithms.

teytaud:
IMHO include the model as complete as possible -- even more if it is difficult. We will fix that by improving the plateform, not by simplifying the model. I'll play with it and use it for improving Nevergrad.

teytaud:
Hello! I come back to this.
Nevergrad has made significant progresses towards constraint solving.
I prefer, by far, that you include a problem that is too hard for Nevergrad and then I'll try to solve it, rather than simplifying it so that it is already solved. I can develop various tools for making constraint satisfaction better, but I need challenging problems --- if the current solution is ok for all problems currently integrated in Nevergrad I will not do more for the moment.

pacowong:
Thanks for the information. Is the improvement mainly due to the introduction of new optimizers or sampling methods? Any special settings I need to take care of?

PS: Recently, I'm a bit busy with my new job. I'll revisit the project later.

teytaud:
Hello! No idea of the current situation :-) our constraint management tools are better.
If that is not enough for the problem you are considering, I guess we should still merge it, and then I'll improve the constraint management tool for properly addressing it --- currently the main issue for constraints is that we have stopped improving it because it was good enough for solving every problem we have.
Yes, adding constraints with "register_cheap_constraints" should be ok, unless Jeremy recommends using a different API.

pacowong:
Finally have time to update this PR :s 
@teytaud I tried to create a test case `nevergrad/functions/pownet/test_solver.py`, yet the solver runs for a very long time and does not return results. Would you mind taking a look?

jrapin:
Hey,
please add pypownetr to mypy.ini so as to ignore the lack of type hints, it should solve  the static check issue in the CI
The solver indeed seems to take more then 10min, I cant check right now, but is it something that happens locally as well?
How many parameters are there? (CMA scales quadratically) how much time does a function evaluation last? are there constraints? (I couldnt see any specifically in the PR, but Pyomo uses some right?) The constraint solver may be the issue now

teytaud:
> Hey,
> please add pypownetr to mypy.ini so as to ignore the lack of type hints, it should solve the static check issue in the CI
> The solver indeed seems to take more then 10min, I cant check right now, but is it something that happens locally as well?
> How many parameters are there? (CMA scales quadratically) how much time does a function evaluation last? are there constraints? (I couldnt see any specifically in the PR, but Pyomo uses some right?) The constraint solver may be the issue now

Note that I am more than happy if we have a hard problem in particular if hardness comes from constraints. I'll use it for improving our constraint handling ---- which is not good.

pacowong:
The long-running time happens to me locally as well. This problem is interesting as it requires a solver to handle the constraints among the producer and consumers in a network.

Below are the statistics:
# of constraints: 28
# variables: 8 variables
# of objectives: 1

PR #875, opened 2020-10-21
Add hyperopt

### LINKED ISSUE #622: Interfacing Nevergrad with more Bayesian optimization libraries ?
We have an interface with a single Bayesian optimization library.
We have no parallel version, or no version for the noisy case.
We might either
- explore a better use of the existing bayesian optimization library
- implement an interface towards another bayesian optimization library

The example of SPSA is one of the simplest examples of implementation of an algorithm, so this could be used as a model regarding how to include a new algorithm: https://github.com/facebookresearch/nevergrad/blob/2ef9cb8cc3e1b769745f05dc1c8689386459e88c/nevergrad/optimization/optimizerlib.py#L686

(_internal_ask should provide the next point to be tested, this is the key part of implementing an optimization algorithm)

herilalaina:
## Types of changes

Fixes #622 

```bash
python -W ignore -m nevergrad.benchmark seqmltuning --seed=12 --repetitions=10 --plot --num_workers=100
```

fight_all

## Checklist


nevergrad/optimization/optimizerlib.py
```diff
@@ -1534,6 +1536,162 @@ def __init__(
...
+
+    def __init__(self, parametrization: IntOrParameter,
+                 budget: tp.Optional[int] = None,
+                 num_workers: int = 1,
+                 *,
+                 prior_weight=1.0,
+                 n_startup_jobs=20,
+                 n_EI_candidates=24,
+                 gamma=0.25,
+                 verbose=False) -> None:
```
jrapin, line 1581: If there are additional parameters, you should add a ConfiguredOptimizer class to wrap this. You can take ParametrizedOnePlusOne as an example, it is very short. You can move most of the docstring there as well, since the arguments should only be called through the ConfiguredOptimizer instantiation

I've also added 

```suggestion
                 prior_weight: float = 1.0,
                 n_startup_jobs: int = 20,
                 n_EI_candidates: int = 24,
                 gamma: float = 0.25,
                 verbose: bool = False) -> None:
```

jrapin (commented):
Hi,
Thanks for the PR. I've added some structural changes to do so that it fits better in the current code.

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -475,6 +475,62 @@ def test_bo_ordering() -> None:
...
 
+def test_hyperopt() -> None:
+    # Test parametrization
+    list_simple_parametrizations = [
+        ng.p.Choice(list(range(10))),
+        ng.p.Scalar(lower=0, upper=1),
+        ng.p.Scalar(lower=0, upper=1).set_integer_casting(),
+        ng.p.Log(lower=1e-3, upper=1e3),
+        ng.p.Array(init=np.zeros(10))
+    ]
```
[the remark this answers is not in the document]
herilalaina: ``@pytest.mark.parametrize`` is really cool :) thanks

teytaud:
Adding hyperopt would be fantastic :-)

nevergrad/optimization/optimizerlib.py
```diff
@@ -1534,6 +1540,220 @@ def __init__(
...
+                 *,
+                 prior_weight: float = 1.0,
+                 n_startup_jobs: int = 20,
+                 n_EI_candidates: int = 24,
+                 gamma: float = 0.25,
+                 verbose: bool = False) -> None:
+        super().__init__(parametrization, budget=budget, num_workers=num_workers)
+        try:
+            # try to convert parametrization to hyperopt search space
+            if not isinstance(parametrization, p.Instrumentation): raise NotImplementedError
```
jrapin, line 1557: please check `self.parametrization` instead for consistency. The call to `__init__` converts integer inputs to `Array`, so you should not need this special case as it should be handled the same way as `Array` (which if I understand well is not supported either if not bounded)

nevergrad/optimization/optimizerlib.py
```diff
@@ -1534,6 +1540,220 @@ def __init__(
 BO = ParametrizedBO().set_name("BO", register=True)
 
 
+class _HyperOpt(base.Optimizer):
```
jrapin, line 1543: since the code is quite big and an interface to another lib, which may require even more integrations later on, let's move this to a different file (example "nevergrad/optimization/nghyperopt.py", I would at least avoid calling the file "hyperopt.py" because that could make a mess with the imports!)

nevergrad/optimization/optimizerlib.py
```diff
@@ -1534,6 +1540,220 @@ def __init__(
...
+        raise NotImplementedError
+
+    def _dict_to_parametrization(self, x):
+        if isinstance(x, dict) and ("args" in x) and ("kwargs" in x):
+            x["args"] = tuple([self._dict_to_parametrization(x["args"][str(i)]) for i in range(len(x["args"]))])
+            x["kwargs"] = {k: self._dict_to_parametrization(v) for k, v in x["kwargs"].items()}
+            return (x["args"], x["kwargs"])
+        return x
+
+    def _parametrization_to_dict(self, x, **kwargs):
```
jrapin, line 1623: since self is not really used in `_dict_to_parametrization` and `_parametrization_to_dict` it would be better to move them out of the class (and keep them in the safe hyperopt-specific file as mentioned aboove), and test them independently of the rest

nevergrad/optimization/optimizerlib.py
```diff
@@ -1534,6 +1540,220 @@ def __init__(
...
+        if isinstance(x, p.Instrumentation):
+            x_dict = kwargs["default"] if "default" in kwargs else {}
```
jrapin, line 1625:
```suggestion
            x_dict = kwargs.get("default", {})
```

jrapin (commented):
Thanks for the work!
I'll need more time to understand everything in closer details. I'll need to understand further how the conversion works. Have you considered using `get_standardized_data` and map to a bounded space? If so, why did you choose against it?
An initial answer may be that in this case the bounds are not always explicit, which is definitely true, but I think we can circumvent this issue if need be.

herilalaina:
> ... Have you considered using `get_standardized_data` and map to a bounded space? If so, why did you choose against it?
> An initial answer may be that in this case the bounds are not always explicit, which is definitely true, but I think we can circumvent this issue if need be.

Current implementation adapts nevergrad parametrization into hyperopt search space if it succeeds else uses `get_standardized_data`. I agree that using standardized data interfaces (as in `BO`) would be much simpler. But in contrast to `BO`, hyperopt is able to efficiently handle categorical and conditional variables which I think is one of the main advantages of using hyperopt.

jrapin:
> > ... Have you considered using `get_standardized_data` and map to a bounded space? If so, why did you choose against it?
> > An initial answer may be that in this case the bounds are not always explicit, which is definitely true, but I think we can circumvent this issue if need be.
> 
> Current implementation adapts nevergrad parametrization into hyperopt search space if it succeeds else uses `get_standardized_data`. I agree that using standardized data interfaces (as in `BO`) would be much simpler. But in contrast to `BO`, hyperopt is able to efficiently handle categorical and conditional variables which I think is one of the main advantages of using hyperopt.

Looks like a good reason indeed. I'll need to figure out how this can be made easier in the future.
When you have addressed the other comments we can merge this thanks ;)

nevergrad/optimization/optimizerlib.py
```diff
@@ -1534,6 +1540,220 @@ def __init__(
...
+                    for idx_param in range(len(param[0].value))
+            }
+            space["kwargs"] = {
+                    param_name: self._get_search_space(param_name, param[1][param_name]) # type: ignore
+                    for param_name in param[1].value.keys()
+            }
+            return space
+        elif isinstance(param, (p.Log, p.Scalar)):
+            if (param.bounds[0][0] is None) or (param.bounds[1][0] is None):
+                raise ValueError(f"Scalar {param_name} not bounded.")
```
teytaud, line 1588: Maybe replace  << raise ValueError(f"Scalar {param_name} not bounded.") >> by
                << return hp.lognormal(label=param_name, 0., 1.) >> ?

herilalaina:
Sorry for the looonnng delay. I had difficulty to understant how `get/set_standardized_data` interfaces work. In some cases (when `data > 8`), it seems that  `get_standardized_data` is unable to revert back the change in `set_standardized_data`. Is this normal ?

```python
import numpy as np
import nevergrad as ng
parametrization = ng.p.Scalar(lower=0, upper=10)
for _ in range(10):
    data = np.random.uniform(0, 9, 1)
    cand = parametrization.spawn_child().set_standardized_data(data)
    np.testing.assert_allclose(data, cand.get_standardized_data(reference=parametrization), rtol=1e-1)

...
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  File "/home/tau/hrakotoa/miniconda3/envs/nevergrad/lib/python3.7/site-packages/numpy/testing/_private/utils.py", line 1528, in assert_allclose
    verbose=verbose, header=header, equal_nan=equal_nan)
  File "/home/tau/hrakotoa/miniconda3/envs/nevergrad/lib/python3.7/site-packages/numpy/testing/_private/utils.py", line 840, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=0.1, atol=0

Mismatched elements: 1 / 1 (100%)
Max absolute difference: 11.10761735
Max relative difference: 4.34943207
 x: array([8.553809])
 y: array([-2.553809])
```

jrapin:
I think it is, because of a sequence of behaviors:
- `set_standardized_data` takes a reference as input. If not provided, the reference (aka 0) is the current position
- in your case is 5 (see `parametrization.value`) since `Scalar` initializes in the middle of lower and upper if both or provided. - when you add a value (Eg. 8) which makes it go beyond the bounds, then the constraint kicks in and sets a different value (by default, the constraint is handled by "bouncing" on the bounds, so 5+8=13 would become 7)

There are good reasons for each of these different behaviors independently (I can detail if need be), but that can indeed be confusing, sorry :s I'm still looking for better patterns but not so sure it is actually feasible.

herilalaina:
Thank you for your explanations :) 
It means that values suggested by hyperopt ([here](https://github.com/facebookresearch/nevergrad/pull/875/files#diff-f25801d4d1b86c648b8c2d010664a596ec8fd79bf6e07086c9f2e63dd1bc0fe4R1654)) can be (slightly) different with the standardized data of the returned candidate; same thing for values computed with `_transform`. I will update the code to handle this behavior.

nevergrad/optimization/optimizerlib.py
herilalaina: I don't have strong opinion on that. I replaced it (but it will only make sense for real p.Scalar). 
Anyway, `hyperopt` is most suited to hyperparameter tuning and for such problems variables are usually bounded.

requirements/bench.txt
```diff
@@ -14,3 +14,4 @@ torchvision
 pyomo>=5.7
 IOHexperimenter>=0.2.8
 mixsimulator>=0.2.9.5
+hyperopt
```
jrapin, line 17: can you add a version number? That helps tracking what was a working version

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -484,6 +484,55 @@ def test_bo_ordering() -> None:
     optim.provide_recommendation()
 
 
+@pytest.mark.parametrize(  # type: ignore
```
jrapin, line 487: I think I would prefer seeing this in a test_nghyperopt.py file, because it's quite specific and complex, because more tests may be added eventually to test specifically the helper functions (_parametrization_to_dict etc), and because eventually I need to split this test file which is getting always bigger :D

jrapin (commented):
See couple of changes.
Overall the code is quite complex so I did not go into the details.
If you have ideas of tests that could be added to make sure everything works fine, it would be great because that will help refactor the code eventually if need be, otherwise i'm fine with merging it.

herilalaina:
I renamed `nghyperopt` to `ng_external_bo` (because I am planning to add other BO optimizers soon :)). Test failed on building docs and I can not figure why. Do you have an idea ?

jrapin:
> I renamed `nghyperopt` to `ng_external_bo` (because I am planning to add other BO optimizers soon :)).

Can you rename it to `externalbo` instead? 2 reasons for that:
- no need for an extra `ng` since we're already in the codebase :D
- following pep8 coding style, module/package names should not contain underscores

 > Test failed on building docs and I can not figure why. Do you have an idea ?

Can you try merging master into your branch? I fixed this build error in #890

jrapin:
Thanks for the update. The remaining issue is unrelated and should get fixed soon, I'll merge your PR right after, thanks for the work ;)

PR #876, opened 2020-10-21
Add an interface with mixsimulator

TokyAxel:
## Types of changes

## Checklist

### DIFF
--- a/nevergrad/functions/mixsimulator/core.py
+++ b/nevergrad/functions/mixsimulator/core.py
@@ -8,7 +8,7 @@
 import numpy as np
 #from nevergrad.parametrization import parameter as p
 from ..base import ExperimentFunction
-from mixsimulator.MixSimulator import MixSimulator
+from mixsimulator.MixSimulator import MixSimulator  #type: ignore
 
 
 class OptimizeMix(ExperimentFunction):
--- a/nevergrad/functions/mixsimulator/core.py
+++ b/nevergrad/functions/mixsimulator/core.py
@@ -6,22 +6,29 @@
 # Based on https://github.com/Foloso/MixSimulator/tree/nevergrad_experiment
 
 import numpy as np
-#from nevergrad.parametrization import parameter as p
 from ..base import ExperimentFunction
 from mixsimulator.MixSimulator import MixSimulator  #type: ignore
 
 
 class OptimizeMix(ExperimentFunction):
-
-    def __init__(self) -> None:
-        mix = MixSimulator()
-        mix.set_data_to("Toamasina")
+    """
+        MixSimulator is an application with an optimization model for calculating 
+        and simulating the least cost of an energy mix under certain constraints.
         
-        #If time == one_week --> dim = 672
-        params_one_week = mix.get_opt_params(672)
-        params_one_week.set_name("dim672")
+        For now, it uses a default dataset (more will be added soon).
         
-        super().__init__(self._simulate_mix,params_one_week)
+        For more information, visit : https://github.com/Foloso/MixSimulator     
+        
+        Parameters
+        ----------
+        time: int
+            total time over which it evaluates the mix (must in hour)
+                
+    """
+    def __init__(self,time: int = 168) -> None:
+        
+        parameters = self.parametrization(time = time)
+        super().__init__(self._simulate_mix , parameters)
         
         self.register_initialization()
 
@@ -31,3 +38,13 @@ class OptimizeMix(ExperimentFunction):
         mix.set_penalisation_cost(100)
         mix.set_carbon_cost(10)
         return mix.loss_function(x)
+        
+    def parametrization(self,time : int = 168):
+        mix = MixSimulator()
+        mix.set_data_to("Toamasina")
+        
+        #If time == 168h (one_week) --> dim = 672
+        params = mix.get_opt_params(time)
+        params.set_name("dims")
+        
+        return params
--- a/nevergrad/functions/mixsimulator/core.py
+++ b/nevergrad/functions/mixsimulator/core.py
@@ -26,16 +26,13 @@ class OptimizeMix(ExperimentFunction):
                 
     """
     def __init__(self,time: int = 168) -> None:
-        self.__mix = MixSimulator()
-        self.__mix.set_data_to("Toamasina")
-        self.__mix.set_penalisation_cost(100)
-        self.__mix.set_carbon_cost(10)
-        parameters = self.__mix.get_opt_params(time)
+        self._mix = MixSimulator()
+        self._mix.set_data_to("Toamasina")
+        self._mix.set_penalisation_cost(100)
+        self._mix.set_carbon_cost(10)
+        parameters = self._mix.get_opt_params(time)
         parameters.set_name("dims")
-        super().__init__(self._simulate_mix , parameters)
+        super().__init__(self._mix.loss_function, parameters)
         self.register_initialization(time=time)
         self.add_descriptors(time=time)
-
-    def _simulate_mix(self, x: np.ndarray) -> float:
-        return self.__mix.loss_function(x)
     
--- a/nevergrad/functions/mixsimulator/core.py
+++ b/nevergrad/functions/mixsimulator/core.py
@@ -30,4 +30,4 @@ class OptimizeMix(ExperimentFunction):
         mix.set_data_to("Toamasina")
         mix.set_penalisation_cost(100)
         mix.set_carbon_cost(10)
-        return mix.loss_func(x)
+        return mix.loss_function(x)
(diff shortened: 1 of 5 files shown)


TokyAxel:
There is an issue about type hints in our package but we already work on it :)

jrapin:
Hi,
Thanks for the PR!
Can you please add a simple unit test in mixsimulator folder?
Otherwise all seems good to me

Foloso:
> Hi,
> Thanks for the PR!
> Can you please add a simple unit test in mixsimulator folder?
> Otherwise all seems good to me

Hey 🙂 ok we'll add that

Foloso:
> Hi,
> Thanks for the PR!
> Can you please add a simple unit test in mixsimulator folder?
> Otherwise all seems good to me

you mean an example of code?

nevergrad/functions/mixsimulator/core.py
```diff
@@ -0,0 +1,33 @@
...
+
+# Based on https://github.com/Foloso/MixSimulator/tree/nevergrad_experiment
+
+import numpy as np
+#from nevergrad.parametrization import parameter as p
+from ..base import ExperimentFunction
+from mixsimulator.MixSimulator import MixSimulator  #type: ignore
+
+
+class OptimizeMix(ExperimentFunction):
```
jrapin, line 14: > you mean an example of code?

No that's not needed, but that actually reminds me that a quick description here as a docstring will be really helpful for other users, so that they understand in a few lines what this test case is about.

requirements/bench.txt
```diff
@@ -13,3 +13,4 @@ tqdm
 torchvision
 pyomo>=5.7
 IOHexperimenter>=0.2.8
+mixsimulator
```
jrapin, line 16: can you also add a version number please, that's helpful to identify which versions has been initially used.

jrapin (commented):
> you mean an example of code?

No, a unit test, aka a function which will be automatically run on each pull request to test that it does not break anything. Without unit test, your code could be broken and we won't know about it, so it's important.

A very basic example is here:
https://github.com/facebookresearch/nevergrad/blob/cb5472c0c597a8fd6156d6e85064dd734bf5fc5f/nevergrad/functions/stsp/test_core.py#L10-L14
It just runs the function, with appropriate inputs, and checks that the outputs is not completely stupid.

It's just a function with name starting with `test_` in a file with name starting with `test_` (just create mixsimulator/test_core.py), so that they can be identify by the unit test framework we use (`pytest`)

nevergrad/functions/mixsimulator/core.py
```diff
@@ -0,0 +1,50 @@
...
+        self.register_initialization()
+
+    def _simulate_mix(self, x: np.ndarray) -> float:
+        mix = MixSimulator()
+        mix.set_data_to("Toamasina")
+        mix.set_penalisation_cost(100)
+        mix.set_carbon_cost(10)
+        return mix.loss_function(x)
+        
+    def parametrization(self,time : int = 168):
```
jrapin, line 42:
```suggestion
    def _create_parametrization(self, time: int = 168):
```
The parametrization is recorded in the `parametrization` attribute, so the function needs to be renamed

nevergrad/functions/mixsimulator/core.py
```diff
@@ -0,0 +1,50 @@
...
+        
+        Parameters
+        ----------
+        time: int
+            total time over which it evaluates the mix (must in hour)
+                
+    """
+    def __init__(self,time: int = 168) -> None:
+        
+        parameters = self.parametrization(time = time)
```
jrapin, line 30:
```suggestion
    def __init__(self, time: int = 168) -> None:
        
        parameters = self._create_parametrization(time=time)
```

nevergrad/functions/mixsimulator/core.py
```diff
@@ -0,0 +1,50 @@
...
+        super().__init__(self._simulate_mix , parameters)
+        
+        self.register_initialization()
```
jrapin, line 33:
```suggestion
        self.register_initialization(time=time)
        self.add_descriptors(time=time)
```

The first line makes sure calling `copy()` on an instance creates a new instance with the same parameters. The second line makes sure the time is stored, for differentiating different cases when plotting results

nevergrad/functions/mixsimulator/core.py
TokyAxel: Hi :-)  we got it. thanks

nevergrad/functions/mixsimulator/core.py
```diff
@@ -0,0 +1,51 @@
...
+        super().__init__(self._simulate_mix , parameters)
+        
+        self.register_initialization(time=time)
+        self.add_descriptors(time=time)
+
+    def _simulate_mix(self, x: np.ndarray) -> float:
+        mix = MixSimulator()
+        mix.set_data_to("Toamasina")
+        mix.set_penalisation_cost(100)
+        mix.set_carbon_cost(10)
```
jrapin, line 40: it feels weird to have to recreate this each time, is there a reason? Wouldn't that be better to instantiate it in the init and keep it as an attribute? This way `_create_parametrization` will not be needed anymore either, since everything could be done from the init.

Also, it's better to avoid this code duplication between `simulate_mix` and `create_parametrization`, because if eventually something is changed (eg: Toamasina), then no need to change it twice.
Finally, it looks like even _simulate_mix could be removed, if you then use `super().__init__(self._mix.loss_function, parameters)`
Less code is always better :D

TokyAxel: You are right :D
Our first goal was to make everything work. But we already work on it to make it better now.

Foloso: Yes thank you for your help :-) and sorry for the delay. We work on it. Just wanted to make it work first :-)

nevergrad/functions/mixsimulator/core.py
```diff
@@ -0,0 +1,41 @@
...
+        For more information, visit : https://github.com/Foloso/MixSimulator     
+        
+        Parameters
+        ----------
+        time: int
+            total time over which it evaluates the mix (must be in hour)
+                
+    """
+    def __init__(self,time: int = 168) -> None:
+        self.__mix = MixSimulator()
```
jrapin, line 29: I would not recommend using a double underscore, this has complicated behaviors when it comes to subclassing (the name ends up being different). Up to you if you are confortable with it though, on my side I tend to go always with a single underscore

nevergrad/functions/mixsimulator/core.py
```diff
@@ -0,0 +1,41 @@
...
+        self.__mix.set_penalisation_cost(100)
+        self.__mix.set_carbon_cost(10)
+        parameters = self.__mix.get_opt_params(time)
+        parameters.set_name("dims")
+        super().__init__(self._simulate_mix , parameters)
+        self.register_initialization(time=time)
+        self.add_descriptors(time=time)
+
+    def _simulate_mix(self, x: np.ndarray) -> float:
+        return self.__mix.loss_function(x)
```
jrapin, line 40: do you plan on adding other stuff here later on? if not `_simulate_mix` is useless since it is only an alias, you can directly use `super().__init__(self.__mix.loss_function, parameters)` in

nevergrad/functions/mixsimulator/core.py
jrapin: no worries in term of delays, I am not always fast to answer, sorry :s

nevergrad/functions/mixsimulator/core.py
Foloso: nothing to add for now, so yes, gonna do this right now

jrapin (approved):
Let's factorize further if possible, otherwise this looks good to me ;)

nevergrad/functions/mixsimulator/core.py
Foloso: It's true that it can a problem for scalability.. Gonna fix that now too

nevergrad/functions/mixsimulator/core.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Based on https://github.com/Foloso/MixSimulator/tree/nevergrad_experiment
+
+import numpy as np
+from ..base import ExperimentFunction
+from mixsimulator.MixSimulator import MixSimulator  #type: ignore
```
jrapin, line 10:
```suggestion
from mixsimulator.MixSimulator import MixSimulator  #type: ignore
from ..base import ExperimentFunction
```
(standard order: from more general to more specific, so external first and internal afterwards)

jrapin:
Should I merge if all tests pass?

Foloso:
> Should I merge if all tests pass?

Yes!! :-)

Foloso:
> Should I merge if all tests pass?

I think it's OK now?

jrapin:
> > Should I merge if all tests pass?
> 
> I think it's OK now?

Yes! Thanks for the contribution!

TokyAxel:
Thanks also for your helps! :-)

PR #844, opened 2020-10-07
Add noise option for Mujoco

teytaud:
## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context / Related issue

Nobody believes that actions can be physically ok at multiplicative precision 1e-5 so having a counterpart
which takes into account such uncertainties would be great.

## Checklist

- [x] The documentation is up-to-date with the changes I made.
- [x] I have read the [**CONTRIBUTING**](https://facebookresearch.github.io/nevergrad/contributing.html) document and completed the CLA (see [**CLA**](https://facebookresearch.github.io/nevergrad/contributing.html#contributor-license-agreement-cla)).
- [x] All tests passed, and additional code has been covered with new tests.

PR #887, opened 2020-11-02
[draft, do not review now] Add an optional second layer in mujoco.

teytaud:
## Types of changes

## Checklist


nevergrad/functions/control/core.py
```diff
@@ -40,19 +40,22 @@ class BaseFunction(ExperimentFunction):
     """
 
     def __init__(self, num_rollouts: int, random_state: tp.Optional[int] = None) -> None:
-        super().__init__(self._simulate, p.Array(shape=self.policy_dim))
+        if self.policy_2ndlayer_dim is not None:
+            super().__init__(self._simulate, np.p.Instrumentation(p.Array(shape=self.policy_dim), p.Array(shape=self.policy_2ndlayer_dim).set_mutation(sigma=1.e-4)))
+        else:
+            super().__init__(self._simulate, p.Array(shape=self.policy_dim))
```
jrapin, line 46: avoid calling super in a if statement, it's always better to uncouple everything that can be uncoupled. In this case create the parametrization in the if, then call init afterwards

nevergrad/functions/control/core.py
```diff
@@ -152,3 +159,12 @@ class Humanoid(BaseFunction):
...
...197, 1.0435800890962725, -2.0834375475277884, 1.2667537566732037, -0.30076659164163855, -2.936598660670857, -4.701568329244183, 0.5482891196808485, -10.245820936980875, 7.420004642248826, 0.4255468058894504, 17.167740585723312, 16.175268727949796, 181.48592716060185, 0.11325750561423298, -0.06582083133392769, 0.0526578564471078, 0.2371962961790206, 0.3042510288306412, -0.12644331987411908, 2.523881380476937, -1.547423424434062, 0.33845447817037794, 3.7044012690867003, 5.77270922236946, -1.4549487373900374, 8.414220368253863, -4.740154811733693, 0.37906093062842394, -6.603153276131259, -14.672811393696326, 152.19174733969905, -9.948554129284157e-07, -3.1729733223247383e-07, 1.6278277777934387e-07, -3.6080215121179876e-05, -4.327504944517283e-05, 0.00011092777930590747, 0.0005553154635918675, -0.00021156350367889417, 7.217738550388757e-05, -0.0007368680078812343, -0.001578912969766702, 0.0007730670670347712, -7.128808629768361e-07, 4.5728191857655706e-07, 2.2865347644914397e-07, 9.111110122544389e-06, 1.160719855911783e-05, 9.381773874831127e-06, 5.569740868840155e-06, -1.4891833711621733e-06, 4.342784980667928e-06, 6.777033412499121e-05, 8.971821928675218e-05, 3.889718928290386e-05]
```
jrapin, line 169: copying something like this is never good for maintenance, and actually completely useless since there are multiple ways to use the original one

nevergrad/functions/control/core.py
```diff
@@ -40,19 +40,22 @@ class BaseFunction(ExperimentFunction):
     """
 
     def __init__(self, num_rollouts: int, random_state: tp.Optional[int] = None) -> None:
-        super().__init__(self._simulate, p.Array(shape=self.policy_dim))
+        if self.policy_2ndlayer_dim is not None:
```
jrapin, line 43: why is that not a parameter ?
Just make a parameter out of it, add it to register_initialization and descritpor, then you do not need a new class. It's the same as for adding noise #844 
also, I guess you don't need to provide a dim, since you don't have a single choice a for it, so make a bool out of it `second_layer: bool`. The simpler the better.

jrapin (changes requested):

nevergrad/functions/control/core.py
teytaud: We do have choice. Using 17x17 is arbitrary. We could have (N x 376) and (17 x N) for any N.

nevergrad/functions/control/core.py
teytaud: Done.

nevergrad/functions/control/core.py
teytaud: Created an inherited class.

nevergrad/functions/control/core.py
teytaud: Maybe I can have an optional int. None ==> single_layer. int ==> neural controller.

jrapin: if you do not plan for multiple hidden layers, then that's perfect indeed.

nevergrad/functions/control/core.py
jrapin: Please do not create another class, otherwise making combinations (with noise for instance) or propagating to other cases (not just Humanoids) will become a complete pain. If it's a parameter of the base class, then it becomes available for **all** Mujoco cases, which is much cleaner.

teytaud:
Changes in progress; closed for the moment.

nevergrad/functions/control/mujoco.py
```diff
@@ -41,6 +42,14 @@ def __call__(self, x):
             totalr = 0.
             while not done:
                 action = np.dot(x, (obs - self.mean) / self.std)
+                if y is not None:
+                    action = 0.1 * action
```
herilalaina, line 46: @teytaud Why do we need this ?

jrapin (changes requested):
As a reminder the Neuro* classes must be removed in favor of an init arguments

teytaud:
Deprecated.

Issue #761, opened 2020-07-12
coauthor requested -- tasks for one of our next papers :-)  


teytaud:
These tasks require a substantial effort, people who solve them are welcome coauthors for our next publications:

Finishing https://github.com/facebookresearch/nevergrad/pull/739 or 
finishing https://github.com/facebookresearch/nevergrad/pull/706 or
solving https://github.com/facebookresearch/nevergrad/issues/738 or 
solving https://github.com/facebookresearch/nevergrad/issues/622 or 
solving https://github.com/facebookresearch/nevergrad/issues/754

(external contributors first in the ordering of authors)

PR #894, opened 2020-11-09
Add Unit Commitment Problem

pacowong:
## Types of changes

Refer to https://github.com/facebookresearch/nevergrad/issues/761, I have added an unit commitment problem with reference to a Pyomo model in https://jckantor.github.io/ND-Pyomo-Cookbook/04.06-Unit-Commitment.html

In my implementation, the violation of the constraints is penalized, is it OK? 
I have tried `register_cheap_constraint`, but Nevergrad takes a very long time to produce a solution (for CMA). This is also the reason why I don't use Pyomo model converter in Nevergrad.

The upper bounds of the output are checked.

## Checklist


nevergrad/functions/unit_commitment/__init__.py
```diff
@@ -0,0 +1 @@
+from .core import UnitCommitmentProblem1 as UnitCommitmentProblem1
```
jrapin, line 1: for pep8 compliance, please rename the directory from `unit_commitment` to `unitcommitment`

nevergrad/functions/unit_commitment/core.py
```diff
@@ -0,0 +1,58 @@
+import numpy as np
+import nevergrad as ng
+from ..base import ExperimentFunction
+
+class UnitCommitmentProblem1(ExperimentFunction):
```
jrapin, line 5: is there a better name than `Problem1` ?
I feel that either the name should be more explicit, or that the `1` should be a parameter of the init so that you can instantiate all problems similarly.

nevergrad/functions/unit_commitment/core.py
```diff
@@ -0,0 +1,58 @@
...
+    ----------
+    T_points: int
+        number of time points
+    N_generators: int
+        number of generators
+    penalty_weight: float
+        weight to penalize for violation of constraints
+    """
+    def __init__(self, T_points: int = 13,
+                 N_generators: int = 3,
```
jrapin, line 21: for pep8 compliance please rename without capital letters.
I would recommend `num_timepoints` and `num_generators` for clarity, or even just `timepoints` and `generators` since fro; the typing it's clear that it's a number

nevergrad/functions/unit_commitment/core.py
```diff
@@ -0,0 +1,58 @@
...
+                 penalty_weight : float = 10000
+                 ) -> None:
+        params = {x: y for x, y in locals().items() if x not in ["self", "__class__"]}  # for copying
+        # Demand for certain time period
+        self.T_points = T_points
+        self.demands = np.array([100 + 100*np.random.uniform() for t in range(self.T_points)])
```
jrapin, line 27: please direcly use `np.random.uniform` parameters, it can take a lower bound, an upper bound and a size.
You may also consider using a possibly seeded random generator for reproducibility, but that really depends how you want your function to behave. The benchmarks should be reproducible without it anyway since the standard numpy random number generator is seeded before running

nevergrad/functions/unit_commitment/core.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # Generating units
+        self.N_generators = N_generators
+        self.p_max = 2*max(self.demands) / self.N_generators
+        self.p_min = 0.6*self.p_max
+
+        # Cost
+        self.cost_a = np.array([0.5 + 0.2*np.random.randn() for n in range(self.N_generators)])
```
jrapin, line 35: same as for self.demands

nevergrad/functions/unit_commitment/core.py
```diff
@@ -0,0 +1,58 @@
...
+        self.cost_b = np.array([10*np.random.uniform() for n in range(self.N_generators)])
+
+        self.penalty_weight = penalty_weight
+
+        param_x = ng.p.Array(shape=(self.N_generators, self.T_points)).set_bounds(0, self.p_max)
+        param_u = ng.p.Array(shape=(self.N_generators, self.T_points)).set_bounds(0, 1).set_integer_casting()
```
jrapin, line 41: is that either 1 or 0? I feel this will probably go very wrong with this parametrization because that does not leave a lot of options.
I would consider `Choice([0, 1], repetitions=timepoints * generators)` instead but that would require reshaping later on.
@teytaud how would you deal with this?

nevergrad/functions/unit_commitment/test_core.py
```diff
@@ -0,0 +1,20 @@
...
+    T = 10
+    N = 5
+    func = core.UnitCommitmentProblem1(T_points=T, N_generators=N)
+    x = np.random.rand(N, T)
+    u = np.random.rand(N, T)
+    value = func.function(x=x, u=u)
+    optimizer = ng.optimizers.NGO(parametrization=func.parametrization, budget=100)
+    recommendation = optimizer.minimize(func.function)
+    assert value < np.inf
+    assert func(**recommendation.kwargs) < np.inf
```
jrapin, line 20: this part does not seem necessary.
I would rather the test checks against the value at the initialization for instance:
func.functions(*parametrization.args, **parametrization.kwargs)
That would require seeding in the test.
The advantage with this is that if ever we modify the function while we thought we did not, we'll be notified by the test.

nevergrad/functions/unit_commitment/core.py
```diff
@@ -0,0 +1,58 @@
+import numpy as np
+import nevergrad as ng
+from ..base import ExperimentFunction
+
+class UnitCommitmentProblem1(ExperimentFunction):
+    """Model that uses conventional implementation for emi-continuous variables
+    The model is adopted from Pyomo model 1: Conventional implementation for emi-continuous variables
+    (https://jckantor.github.io/ND-Pyomo-Cookbook/04.06-Unit-Commitment.html)
+    The constraints are added to the objective with a heavy penalty for violation.
```
jrapin, line 9: can you develop what are x and u? if that makes sense, more explicit names would help.

jrapin (changes requested):

nevergrad/functions/unit_commitment/core.py
pacowong: Thanks for the suggestion. I have made the changes accordingly. But are these statement handled differently in the backend?-

nevergrad/functions/unitcommitment/core.py
```diff
@@ -0,0 +1,64 @@
...
+        param_operational_output = ng.p.Array(shape=(self.num_generators, self.num_timepoints)).set_bounds(0, self.p_max)
+        #param_operational_states = ng.p.Array(shape=(self.num_generators, self.num_timepoints)).set_bounds(0, 1).set_integer_casting() #
+        param_operational_states = ng.p.Choice([0, 1], repetitions=self.num_timepoints * self.num_generators)
+        instru = ng.p.Instrumentation(operational_output=param_operational_output, operational_states=param_operational_states).set_name("")
+        super().__init__(self.unit_commitment_obj_with_penalization, instru)
+        self.register_initialization(**params)
+        self._descriptors.update(num_timepoints=self.num_timepoints, num_generators=self.num_generators)
+
+
+    def unit_commitment_obj_with_penalization(self, operational_output, operational_states):
```
jrapin, line 52: should that wait for the penalization PR #807 ?

teytaud: Maybe now we would do better! That said the approach with penalization is ok to me.

nevergrad/functions/unit_commitment/core.py
teytaud: @pacowong  Yes they are handled differently in the backend.

teytaud:
The only missing thing is, I think, an addition in nevergrad/benchmark/experiments.py ? this should be just a few lines.

nevergrad/functions/unitcommitment/core.py
```diff
@@ -0,0 +1,80 @@
...
+        self,
+        problem_name="semi-continuous",
+        num_timepoints: int = 13,
+        num_generators: int = 3,
+        penalty_weight: float = 10000,
+    ) -> None:
+        if problem_name not in ["semi-continuous"]:
+            raise NotImplementedError
+
+        params = {x: y for x, y in locals().items() if x not in ["self", "__class__"]}  # for copying
```
jrapin, line 32:
```suggestion
```
The CI states that it's not used (and recent changes make it not needed anymore for copying

nevergrad/functions/unitcommitment/core.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
```
jrapin, line 1: please add the standard header to all new files

nevergrad/functions/unitcommitment/core.py
```diff
@@ -0,0 +1,80 @@
...
+        number of time points
+    num_generators: int
+        number of generators
+    penalty_weight: float
+        weight to penalize for violation of constraints
+    """
+
+    def __init__(
+        self,
+        problem_name="semi-continuous",
```
jrapin, line 24:
```suggestion
        problem_name: str = "semi-continuous",
```

nevergrad/functions/unitcommitment/core.py
```diff
@@ -0,0 +1,80 @@
...
+            0, self.p_max
+        )
+        # param_operational_states = ng.p.Array(shape=(self.num_generators, self.num_timepoints)).set_bounds(0, 1).set_integer_casting() #
+        param_operational_states = ng.p.Choice([0, 1], repetitions=self.num_timepoints * self.num_generators)
+        instru = ng.p.Instrumentation(
+            operational_output=param_operational_output,
+            operational_states=param_operational_states,
+        ).set_name("")
+        super().__init__(self.unit_commitment_obj_with_penalization, instru)
+        self._descriptors.update(num_timepoints=self.num_timepoints, num_generators=self.num_generators)
```
jrapin, line 58:
```suggestion
```

the init parameters are now automatically registered as descriptors if they are picklable

jrapin (commented):
please fix the minor comments and merge master within your branch and it should be good to merge if tests pass

pacowong:
@jrapin, the issues are resolved. Thanks.

jrapin:
merged, thanks @pacowong ;)

PR #903, opened 2020-11-21
Online mujoco

LaurentMnr95:
## Types of changes

## Checklist


nevergrad/functions/control/mujoco.py
```diff
@@ -23,10 +23,20 @@ class GenericMujocoEnv:
...
+
+    def __init__(self, env_name, num_rollouts,
+                 random_state, online_stats, state_mean=None, state_std=None):
+        self.online_stats = online_stats
+        self.s_n = 0.
+        self.n_obs = 0
+
+        if self.online_stats:
+            self.mean = 0.
+            self.std = 0.
```
jrapin, line 35: Why do you use either floats or arrays? having types that is not always the same is not a good idea :s it can lead to weird bugs, and the type checker will hate it because it makes it impossible to check

If those values are only used in the onlise_stats case, then you could just do something like this:
self.mean: np.ndarray
self.std: np.ndarray
if self.online_stats:
    self.mean = np.array(state_mean)
    self.std = np.array(state_std)

nevergrad/functions/control/mujoco.py
```diff
@@ -23,10 +23,20 @@ class GenericMujocoEnv:
...
+        else:
+            self.mean = np.array(state_mean)
+            self.std = np.array(state_std)
```
jrapin, line 38: preferably use `np.array(..., copy=True)`. I think this is the default, but I tend to prefer emphasizing it, especially since otherwise you could be modifying a global variable without noticing it

jrapin (changes requested):

PR #904, opened 2020-11-26
Reduce optimization methods in experiments

teytaud:
## Types of changes

## Checklist


jrapin (commented):
A better option would be to use optimizer groups!

teytaud:
> A better option would be to use optimizer groups!

I have seen that and I had our discussion in mind -- however here for many expensive benchmarks I've switched to something smaller than the corresponding optimizer group. That said I could define new, smaller optimizer groups. Let me check this.

jrapin:
> That said I could define new, smaller optimizer groups.

Exactly :D That's the whole aim of optimizer groups :D 
I like it muuuch better thanks ;)

nevergrad/optimization/optimizerlib.py
```diff
@@ -1631,6 +1660,10 @@ def __init__(
 
 chainCMAPowell = Chaining([CMA, Powell], ["half"]).set_name("chainCMAPowell", register=True)
 chainCMAPowell.no_parallelization = True
+chainMetaModelSQP = Chaining([MetaModel, SQP], ["half"]).set_name("chainMetaModelSQP", register=True)
+chainMetaModelSQP.no_parallelization = True
+chainMetaModelPowell = Chaining([MetaModel, Powell], ["half"]).set_name("chainMetaModelPowell", register=True)
+chainMetaModelPowell.no_parallelization = True
```
jrapin, line 1666: you should define those on the run in an opt group

nevergrad/benchmark/optgroups.py
```diff
@@ -46,6 +46,15 @@ def large() -> tp.Sequence[Optim]:
             "MultiCMA", "TripleCMA", "MultiScaleCMA", "RSQP", "RCobyla", "RPowell", "SQPCMA", "MetaModel", "PolyCMA", "ManyCMA"]
 
 
+def emna_variants() -> tp.Sequence[Optim]:
```
jrapin, line 49:
```suggestion
@registry.register
def emna_variants() -> tp.Sequence[Optim]:
```

nevergrad/benchmark/optgroups.py
```diff
@@ -46,6 +46,15 @@ def large() -> tp.Sequence[Optim]:
...
+    return ["IsoEMNA", "NaiveIsoEMNA", "AnisoEMNA", "NaiveAnisoEMNA", "CMA", "NaiveTBPSA",
+              "NaiveIsoEMNATBPSA", "IsoEMNATBPSA", "NaiveAnisoEMNATBPSA", "AnisoEMNATBPSA"]
+
+
+def basics() -> tp.Sequence[Optim]:
```
jrapin, line 54:
```suggestion
@registry.register
def basics() -> tp.Sequence[Optim]:
```

nevergrad/optimization/experimentalvariants.py
```diff
@@ -302,6 +302,9 @@
 ProgONOPO13 = ConfSplitOptimizer(num_optims=13, progressive=True, multivariate_optimizer=OptimisticNoisyOnePlusOne).set_name(
     "ProgONOPO13", register=True
 )
+ProgONOPOInf = ConfSplitOptimizer(num_optims=10000, progressive=True, multivariate_optimizer=OptimisticNoisyOnePlusOne).set_name(
```
jrapin, line 305: define in an optgroup?

teytaud: Done.

teytaud: Done.

jrapin: No, what I mean is that this should be directly defined in the opt group and not here.
In the optgruop you can just replace the string "ProgONOPOInf" by `ConfSplitOptimizer(num_optims=10000, progressive=True, multivariate_optimizer=OptimisticNoisyOnePlusOne).set_name("ProgONOPOInf")`

If it is used only once, then it is preferably not to define it in experimentalvariants

nevergrad/benchmark/optgroups.py
```diff
@@ -46,6 +47,22 @@ def large() -> tp.Sequence[Optim]:
...
 
+@registry.register
+def emna_variants() -> tp.Sequence[Optim]:
+    return ["IsoEMNA", "NaiveIsoEMNA", "AnisoEMNA", "NaiveAnisoEMNA", "CMA", "NaiveTBPSA",
+              "NaiveIsoEMNATBPSA", "IsoEMNATBPSA", "NaiveAnisoEMNATBPSA", "AnisoEMNATBPSA"]
+
+
+@registry.register
+def progressive() -> tp.Sequence[Optim]:
+    return ["ProgONOPOInf", "ProgONOPOAuto", "ProgONOPO13", "ProgODOPOInf", "ProgODOPOAuto", "ProgODOPO13"]
```
jrapin, line 56: Here is what I would prefer, creating all the combinations with for loops in optgroups:

```python
optims:tp.List[Optim] = []
for mutation in ["discrete", "gaussian"]:
    for num_optims in [None, 13, 10000]:
        name = "Prog" + ("Disc" if mutation == "discrete" else "") + ("Auto" if num_optims is None else str(num_optims)
        mv = ParametrizedOnePlusOne(noise_handling="optimistic", mutation=mutation)
        opt = ConfSplitOptimizer(
            num_optims=num_optims, progressive=True, multivariate_optimizer=mv
        ).set_name(name))
        optims.append(opt)
return optims
```
 Why?
- because it removes a whole lot of optim definition with unreadable names
- because it is easier to extend/adapt
- because it is less likely to be buggy

jrapin (approved):
See the unused imports that need to be removed, then I think it's nice ;)

jrapin:
btw, if you run the tests locally, it is supposed to update a file containing the name of the algorithms in each optgroup. It's not necessary, but it can be useful to have an overview if need be

nevergrad/benchmark/optgroups.py
teytaud: Done.

nevergrad/optimization/optimizerlib.py
teytaud: They are used inside other optimization methods so I think it's difficult. They will be more and more used in the future (see works in progress with significant improvements in particular for big budget). Memetism rules :-)

PR #905, opened 2020-11-27
Score images with iqa

brozi:
Adding an experiment generating images with pytorch_gan_zoo and scoring them with koncept512


nevergrad/functions/images/core.py
```diff
@@ -205,3 +200,51 @@ def _with_tag(
...
+
+        super().__init__(self._loss, array)
+        self.register_initialization(problem_name=problem_name)
+        self._descriptors.update(problem_name=problem_name)
+        self.loss_function = scorer()
+
+    def _loss(self, x: np.ndarray) -> float:
+        image = self._generate_images(x)
+        loss = self.loss_function.compute_loss(image)
+        print(f'k512 loss {loss}')
```
teytaud, line 245: Maybe remove ?
(also it's not necessarily K512, right ?)

nevergrad/optimization/differentialevolution.py
```diff
@@ -257,6 +257,7 @@ def __init__(
 
 
 DE = DifferentialEvolution().set_name("DE", register=True)
+DDE = DifferentialEvolution(crossover="dimension").with_name("DDE", register=True)
```
teytaud, line 260: Can you tell me a bit more about this line ?

teytaud: Ah this is "set_name", not "with_name".

teytaud:
Sounds good to me up to minor details! A second (not neural) IQA would complete the picture. @jrapin, ok for you ?

nevergrad/functions/images/core.py
brozi: Thanks I'll remove it in the next revision. I had forgotten about it.

nevergrad/optimization/differentialevolution.py
brozi: I'll remove this, it has nothing to do in this PR sorry

nevergrad/benchmark/experiments.py
```diff
@@ -1139,6 +1140,21 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
                         yield xp
 
 
+@registry.register
+def imagesUsingGan(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
```
jrapin, line 1144: Following pep8, this should be:
```suggestion
def images_using_gan(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
```

nevergrad/benchmark/experiments.py
```diff
@@ -1139,6 +1140,21 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+    """Optimizing an image using koncept512 and a GAN"""
+    seedg = create_seed_generator(seed)
+    optims = ["CMA", "Shiwa", "DE", "PSO", "RecES", "RecMixES", "RecMutDE", "NSGAIIES", "ParametrizationDE"]
+    if default_optims is not None:
+        optims = default_optims
+    for budget in [100 * 5 ** k for k in range(3)]:
+        for num_workers in [1]:
+            for algo in optims:
+                for func in [ImageFromPGAN()]:
```
jrapin, line 1153: please create the func outside all the for loops, and no need to iterate if there is only one of them (same for num_workers)

nevergrad/functions/images/image_losses.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
```
jrapin, line 1: following pep8, please rename to `imagelosses` (no underscore in module names)

nevergrad/functions/images/image_losses.py
```diff
@@ -0,0 +1,46 @@
...
+        pass
+
+    def compute_loss(self, img: np.ndarray) -> float:
+        assert False
+
+
+class _ImageLossUsingRef(_ImageLoss):
+    def __init__(self, reference: np.ndarray) -> None:
+        super().__init__()
+        self.reference = reference
```
jrapin, line 16: If there is no image loss without ref, then no need to define one
More generally, an image loss should just be a function `func(x: np.ndarray) -> float`: right?

In this case, it's more natural to use `__call__` instead of `compute_loss`

I would do something like this:
```python
class ImageLoss
    
    def __init__(self, reference: np.ndarray, method: str) -> None:
         ...

    def __call__(self, x: np.ndarray)-> float:
        ...
```
That will be way shorter and work just fine as well

nevergrad/functions/images/core.py
```diff
@@ -205,3 +200,50 @@ def _with_tag(
...
+
+    problem_name: the type of problem we are working on.
+    initial_noise: the initial noise of the GAN. It should be of dimension (1, 512). If None, it is defined randomly.
+    use_gpu: whether to use gpus to compute the images
+    scorer: which scorer to use for the images
+    mutable_sigma: whether the sigma should be mutable
+    n_mutations: number of mutations
+    """
+
+    def __init__(self, problem_name: str = "GAN_optimization", initial_noise: np.ndarray = None, use_gpu: bool = True, scorer=Koncept512Loss, mutable_sigma=True, n_mutations=35) -> None:
```
jrapin, line 217: problem name does not seem to be used for anything useful, remove it?

nevergrad/functions/images/core.py
```diff
@@ -205,3 +200,50 @@ def _with_tag(
...
+            initial_noise = np.random.normal(size=self.noise_shape)
+        assert initial_noise.shape == self.noise_shape, f'The shape of the initial noise vector was {initial_noise.shape}, it should be {self.noise_shape}'
+
+        array = ng.p.Array(init=initial_noise, mutable_sigma=mutable_sigma)
+        # parametrization
+        array.set_mutation(sigma=n_mutations)
+        array.set_recombination(ng.p.mutation.Crossover(axis=(0, 1))).set_name("")
+
+        super().__init__(self._loss, array)
+        self.register_initialization(problem_name=problem_name)
```
jrapin, line 238: this function is used to create a copy of the instance by reinitializing a new instance with the parameters you provide here. Since the useful parameters are all but the problem name, the new copy may be completely different from the initial version, so you'll need to update this

nevergrad/functions/images/core.py
```diff
@@ -205,3 +200,50 @@ def _with_tag(
...
+        self._descriptors.update(problem_name=problem_name)
```
jrapin, line 239: this is used to differetiate between different test cases in the benchmark, typically if you use different losses, so you need to provide the information there

jrapin (changes requested):

nevergrad/functions/images/core.py
jrapin: actually, if we wait for #914 then this won't be needed anymore :D

nevergrad/functions/images/image_losses.py
brozi: The Koncept512 Loss is an image loss without ref. 
I'll use __call__ instead of compute_loss.

nevergrad/functions/images/core.py
brozi: I changed "problem_name" to something that contains more information about the useful parameters.

nevergrad/functions/images/core.py
brozi: should I use a name that contains the initial noise ? For now I used the name using the other parameters

nevergrad/functions/images/core.py
brozi: I had just copied it from the Image class above. I guess it's not really useful.

nevergrad/functions/images/core.py
```diff
@@ -20,7 +21,7 @@
 
 
 class Image(base.ExperimentFunction):
-    def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
+    def __init__(self, problem_name: str = "recovering", index: int = 0, loss=SumAbsoluteDifferencesLoss) -> None:
```
jrapin, line 24: It's better to just use a loss name here because that's easier to store as an information for the benchmarks.
If we can store it, we can then easily identify which were run with what kind of loss, so that we can compare. So for ExperimentFunction I strongly recommend keeping to str/int/bool/float only whenever possible

nevergrad/functions/images/core.py
```diff
@@ -205,3 +200,49 @@ def _with_tag(
...
+
+    problem_name: the type of problem we are working on.
+    initial_noise: the initial noise of the GAN. It should be of dimension (1, 512). If None, it is defined randomly.
+    use_gpu: whether to use gpus to compute the images
+    scorer: which scorer to use for the images
+    mutable_sigma: whether the sigma should be mutable
+    n_mutations: number of mutations
+    """
+
+    def __init__(self, initial_noise: np.ndarray = None, use_gpu: bool = True, scorer=Koncept512Loss, mutable_sigma=True, n_mutations=35) -> None:
```
jrapin, line 217: same thing here, better describe it with a string if possible. Unless you plan to have many of those or you want to experiment easily offline with different ones, but in this case we'll need to store some information about them

nevergrad/functions/images/imageLosses.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
+from koncept.models import Koncept512
+
+
+class _ImageLoss:
+    def __init__(self) -> None:
+        pass
+
+    def __call__(self, img: np.ndarray) -> float:
+        assert False
```
jrapin, line 10: A reason you don't need this is that this is basically `tp.Callable[[np.ndarray], float]`

nevergrad/functions/images/core.py
jrapin: where? I think it should just be removed, it's not used at all :s

nevergrad/functions/images/core.py
jrapin: I don't understand what you mean. In any case for now I guess you can just remove it

nevergrad/functions/images/core.py
jrapin: yeah I figured afterwards, it should not have been in the class above in the first class, I had not noticed it :s tiny mistakes tend to propagate a lot :D

nevergrad/functions/images/image_losses.py
jrapin: All depends in how you want to use this I guess. For benchmarking with nevergrad, having a instance of a custom class as parameter makes things a bit hard because we can't store this object as a descriptor for the benchmark, so if you change it nevergrad won't notice it and you may end up summing xps with different losses without noticing. So for this it's better to avoid such parameters

If you plan on using the ExperimentFunction on your own, out of nevergrad benchmark system, and with different custom losses, then it makes to leave the option open, but we'll need to improve this a bit so that we don't have issues when running benchmarks

nevergrad/functions/images/core.py
teytaud: You can just remove register_initialization, given jeremy's PR (provided that you git pull origin master).

nevergrad/functions/images/core.py
teytaud: Given Jeremy's PR, you can just remove the descriptor I guess (provided that you git pull origin master).

nevergrad/functions/images/imageLosses.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class _ImageLossUsingRef(_ImageLoss):
+    def __init__(self, reference: np.ndarray) -> None:
+        super().__init__()
+        self.reference = reference
```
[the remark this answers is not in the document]
teytaud: Tens of them, you mean tens of reference ? Then yes we might have tens of references, e.g. 50 looks reasonable.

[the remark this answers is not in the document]
brozi: I think it's about the number of loss functions using a reference. I can think of a few: sum of absolute differences, sum of squares, SSIM, PPSN, combinations of these... It probably won't be tens of them. I feel like it's nicer to have a class to ensure that these losses are easy to reuse and are using the same format.

nevergrad/functions/images/imageLosses.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class SumAbsoluteDifferencesLoss(_ImageLossUsingRef):
+    def __init__(self, reference: np.ndarray) -> None:
+        super().__init__(reference)
+        self.domain_shape = self.reference.shape
+
+    def __call__(self, x: np.ndarray) -> float:
+        x = np.array(x, copy=False).ravel()
+        x = x.reshape(self.domain_shape)
```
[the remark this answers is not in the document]
brozi: to be honest I just copied the loss in the Image class to put it in a class instead... I assumed that everything was done for a good reason but maybe I should check that this is actually useful.

nevergrad/functions/images/core.py
brozi: Do you mean I should define loss='SumAbsoluteDifferences' and then have a series of if and elif to define the loss? Wouldn't that lead to code duplication and also require to update this function when adding a new loss making adding losses more tedious ?

nevergrad/functions/images/core.py
```diff
@@ -205,3 +200,49 @@ def _with_tag(
...
+
+        array = ng.p.Array(init=initial_noise, mutable_sigma=mutable_sigma)
+        # parametrization
+        array.set_mutation(sigma=n_mutations)
+        array.set_recombination(ng.p.mutation.Crossover(axis=(0, 1))).set_name("")
+
+        super().__init__(self._loss, array)
+        self.loss_function = scorer()
+        self.register_initialization(initial_noise=initial_noise, use_gpu=use_gpu, scorer=scorer, mutable_sigma=mutable_sigma, n_mutations=n_mutations)
+        self._descriptors.update(initial_noise=initial_noise, use_gpu=use_gpu, scorer=scorer, mutable_sigma=mutable_sigma, n_mutations=n_mutations)
```
jrapin, line 239: Merge master and delete as follow, this is new to simplify the code :D 
```suggestion
```

nevergrad/functions/images/core.py
jrapin: that's why it should be avoided if there are many losses, but used if there are not so many. Actually with only the current 2 you have now, I think you could remove 30 lines of code and it would work just as well + be more robust.
But if you plan on adding some as you seem to want, then let's keep something like this. Although I need to figure out how.

First thing I don't get is that is taking a reference and the other no, so both are incompatible. Are both losses supposed to work in both cases? I would be in favor of enforcing that an `ImageLoss` can accept a reference as input even if it does not use it. That removes some code. Then we type that `loss: ImageLoss` and we `self.add_descriptors(loss=loss.__name___)` since we need a string to describe it.

nevergrad/functions/images/imageLosses.py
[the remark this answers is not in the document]
jrapin: yeah, never assume some code is done for a good reason :D

nevergrad/functions/images/imageLosses.py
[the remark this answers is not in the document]
jrapin: yes about the number of loss functions using a reference (or not)
discussion continues in https://github.com/facebookresearch/nevergrad/pull/905#discussion_r536268764

nevergrad/functions/images/core.py
jrapin: yes :D definitely merge master, those two lines will disappear :D

nevergrad/functions/images/core.py
brozi: I think you're right, it would make it better, the previous thing wasn't a good interface for all image lossses. I'll do what you suggest.

nevergrad/functions/images/imagelosses.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+from koncept.models import Koncept512
+
+
+class _ImageLoss:
```
jrapin, line 5:
```suggestion
class ImageLoss:
```
(and propagate)

this will be used for typing

nevergrad/functions/images/imagelosses.py
```diff
@@ -0,0 +1,37 @@
...
+    def __init__(self, reference=None) -> None:
+        self.reference = reference
+
+    def __call__(self, img: np.ndarray) -> float:
+        pass
```
jrapin, line 10:
```suggestion
        raise NotImplementedError
```

nevergrad/functions/images/core.py
```diff
@@ -205,3 +198,49 @@ def _with_tag(
...
+
+    problem_name: the type of problem we are working on.
+    initial_noise: the initial noise of the GAN. It should be of dimension (1, 512). If None, it is defined randomly.
+    use_gpu: whether to use gpus to compute the images
+    scorer: which scorer to use for the images
+    mutable_sigma: whether the sigma should be mutable
+    n_mutations: number of mutations
+    """
+
+    def __init__(self, initial_noise: np.ndarray = None, use_gpu: bool = True, scorer=Koncept512Loss, mutable_sigma=True, n_mutations=35) -> None:
```
jrapin, line 215:
```suggestion
    def __init__(self, initial_noise: np.ndarray = None, use_gpu: bool = True,
                 loss: imagelosses.ImageLoss = Koncept512,
                 mutable_sigma=True, n_mutations=35) -> None:
```

nevergrad/functions/images/core.py
jrapin, line 24:
```suggestion
    def __init__(self, problem_name: str = "recovering", index: int = 0,
                 loss=imagelosses.SumAbsoluteDifferences) -> None:
```

nevergrad/functions/images/core.py
```diff
@@ -10,6 +10,7 @@
 import torch.nn as nn
 import torch
 import torchvision
+from nevergrad.functions.images.imageLosses import SumAbsoluteDifferencesLoss, Koncept512Loss
```
jrapin, line 13:
```suggestion
from . import imagelosses
```

nevergrad/functions/images/imagelosses.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class SumAbsoluteDifferencesLoss(_ImageLoss):
```
jrapin, line 13: Given the module and the class, it's clear they are losses. If we import the module we get `imagelosses.SumAbsoluteDifferencesLoss` which would be redundant long and not helping, so better remove it for all but the base class
```suggestion
class SumAbsoluteDifferences(ImageLoss):
```

nevergrad/functions/images/imagelosses.py
```diff
@@ -0,0 +1,37 @@
...
+        super().__init__(reference)
+        self.domain_shape = self.reference.shape
+
+    def __call__(self, x: np.ndarray) -> float:
+        assert x.shape == self.domain_shape, f"Shape = {x.shape} vs {self.domain_shape}"
+        value = float(np.linalg.norm(x - self.reference, 1))
+        return value
+
+
+class Koncept512Loss(_ImageLoss):
```
jrapin, line 24:
```suggestion
class Koncept512(ImageLoss):
```

nevergrad/functions/images/core.py
```diff
@@ -205,3 +198,49 @@ def _with_tag(
...
+
+
+class ImageFromPGAN(base.ExperimentFunction):
+    """
+    Creates face images using a GAN from pytorch GAN zoo trained on celebAHQ and optimizes the noise vector of the GAN
+
+    problem_name: the type of problem we are working on.
+    initial_noise: the initial noise of the GAN. It should be of dimension (1, 512). If None, it is defined randomly.
+    use_gpu: whether to use gpus to compute the images
+    scorer: which scorer to use for the images
```
jrapin, line 210:
```suggestion
    loss: which loss to use for the images
```

nevergrad/functions/images/core.py
```diff
@@ -49,18 +50,10 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
         array.set_recombination(ng.p.mutation.Crossover(axis=(0, 1), max_size=max_size)).set_name("")  # type: ignore
 
         super().__init__(self._loss, array)
-        self.register_initialization(problem_name=problem_name, index=index)
-        self._descriptors.update(problem_name=problem_name, index=index)
+        self.loss_function = loss(reference=self.data)
```
jrapin, line 53: just do
`super().__init__(loss(reference=self.data), array)`
Defining `_loss` does not bring anything

jrapin (changes requested):
please merge master within this PR, so that the new updates I mentioned are available (otherwise the PR won't work)

nevergrad/functions/images/core.py
brozi: We actually also need to import Koncept512Loss but I'll do the relative import

nevergrad/functions/images/imagelosses.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class SumAbsoluteDifferences(ImageLoss):
+    def __init__(self, reference: np.ndarray) -> None:
+        super().__init__(reference)
+        self.domain_shape = self.reference.shape
+
+    def __call__(self, x: np.ndarray) -> float:
+        assert x.shape == self.domain_shape, f"Shape = {x.shape} vs {self.domain_shape}"
+        value = float(np.sum(np.fabs(x - self.reference)))
```
brozi, line 20: I reverted it since np.linalg.norm does max(sum(abs(x), axis=0)), which is different.

nevergrad/functions/images/imagelosses.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+from koncept.models import Koncept512 as K512_model
```
teytaud, line 2:
```suggestion
from koncept.models import Koncept512 as K512_model  # type: ignore
```

requirements/bench.txt
```diff
@@ -15,3 +15,5 @@ pyomo>=5.7
 mixsimulator>=0.2.9.5
 hyperopt>=0.2.5
 IOHexperimenter==0.2.8.7
+koncept>=0.2.2
+tensorflow>=2.3.1
```
jrapin, line 19: is tensorflow necessary? I don't see it anywhere. At some point we decided not to include it in bench.txt because it's big, then again, bench.txt is already big and not installed by default anyway

nevergrad/functions/images/imagelosses.py
jrapin: waouh that's sad, I just read the doc 3 times now and i don't understand why they did this, especially since for numbers different from 1 and 2 they do the expected computation: `sum(abs(x)**ord)**(1./ord)`
Sorry for the mistake

nevergrad/functions/images/core.py
```diff
@@ -184,3 +180,49 @@ def make_folder_functions(
...
+class ImageFromPGAN(base.ExperimentFunction):
+    """
+    Creates face images using a GAN from pytorch GAN zoo trained on celebAHQ and optimizes the noise vector of the GAN
+
+    problem_name: the type of problem we are working on.
+    initial_noise: the initial noise of the GAN. It should be of dimension (1, 512). If None, it is defined randomly.
+    use_gpu: whether to use gpus to compute the images
+    loss: which loss to use for the images
+    mutable_sigma: whether the sigma should be mutable
+    n_mutations: number of mutations
```
jrapin, line 194: FYI here is how to format the docstring so that it can be interpreted automatically by sphinx, which creates the documentation
```suggestion
    Parameters
    ----------
    problem_name: str
        the type of problem we are working on.
    initial_noise: np.ndarray
        the initial noise of the GAN. It should be of dimension (1, 512). If None, it is defined randomly.
    use_gpu: bool
        whether to use gpus to compute the images
    loss: ImageLoss
        which loss to use for the images
    mutable_sigma: bool
        whether the sigma should be mutable
    n_mutations: int
        number of mutations
```

jrapin (commented):
It looks good now, a few more details to sort out and we can merge
pytest seems to have had an issue on windows but this is unrelated (or then again, it may be due to the gpu setting, which could be concerning)

Can you add a very simple test as well that just tests that the function returns the same value under the same conditions ?

requirements/bench.txt
brozi: It's a dependency of koncept and we need to add this line to make sure that the tensorflow version is recent enough to work (at least on my setup on my devfair).

nevergrad/functions/images/core.py
```diff
@@ -47,17 +50,12 @@ def __init__(self, problem_name: str = "recovering", index: int = 0) -> None:
...
     def _loss(self, x: np.ndarray) -> float:
-        assert self.problem_name == "recovering"
-        x = np.array(x, copy=False).ravel()
-        x = x.reshape(self.domain_shape)
-        assert x.shape == self.domain_shape, f"Shape = {x.shape} vs {self.domain_shape}"
-        # Define the loss, in case of recovering: the goal is to find the target image.
-        assert self.index == 0
-        value = float(np.sum(np.fabs(np.subtract(x, self.data))))
-        return value
+        return self.loss_function(x)
```
jrapin, line 58: please delete the `_loss` function which is not used anymore

jrapin (commented):
If the windows problem is still present, then this is concerning because it may not be unrelated :s not sure what to do about it, it may come from tensorflow/pytorch interactions, or issues with the requirement to use gpu :s

nevergrad/functions/images/core.py
brozi: sure. Sorry I thought it was already done.

nevergrad/functions/images/core.py
```diff
@@ -184,3 +178,59 @@ def make_folder_functions(
...
+        whether the sigma should be mutable
+    n_mutations: int
+        number of mutations
+    """
+
+    def __init__(self, initial_noise: np.ndarray = None, use_gpu: bool = False,
+                 loss: imagelosses.ImageLoss = imagelosses.Koncept512(),
+                 mutable_sigma: bool = True, n_mutations: int = 35) -> None:
+        if torch.cuda.is_available():
+            use_gpu = False
```
jrapin, line 207: Isnt it:
```suggestion
        if not torch.cuda.is_available():
            use_gpu = False
```

jrapin (commented):
Given the glitch in Windows I see only one way to move forward for now: do the koncept512 import in the `Koncept512` init, so that it is only imported when needed. I think tensorflow could be messing up with torch, as it usually does.

teytaud:
> If the windows problem is still present, then this is concerning because it may not be unrelated :s not sure what to do about it, it may come from tensorflow/pytorch interactions, or issues with the requirement to use gpu :s

Is it bad if we <<  import os      if os.name != 'nt:'     >>   for this part of the code ?

jrapin:
> > If the windows problem is still present, then this is concerning because it may not be unrelated :s not sure what to do about it, it may come from tensorflow/pytorch interactions, or issues with the requirement to use gpu :s
> 
> Is it bad if we << import os if os.name != 'nt:' >> for this part of the code ?

I'd rather use anything else, if there is an issue with Windows, it may end up being an issue with Mac (there is no CI for Mac) and for Linux later on at a version change. Usually the import order tends to change a lot of things with tensorflow/pytorch so let's investigate this first.

jrapin:
just a reminder: any test which you can add on your class would help **a lot** in the future to make sure nothing is broken. Given the requirements of the function, I expect it could be just instantiating the function (without even calling it) and with the L1 loss. This would be sufficient for basic checks. Without it the code could end up completely broken eventually without us noticing.

brozi:
> just a reminder: any test which you can add on your class would help **a lot** in the future to make sure nothing is broken. Given the requirements of the function, I expect it could be just instantiating the function (without even calling it) and with the L1 loss. This would be sufficient for basic checks. Without it the code could end up completely broken eventually without us noticing.

I added a few unit tests to check that it can be instantiated with both losses and gives coherent losses. I also added a test for the l1 loss. I could have added a test to check the output of K512 for a few values but I don't like this type of tests as they break when the model is updated (which would happen if they update k512).

jrapin (commented):
still a few comments sorry, and I'll try to find a solution today for koncept on windows

jrapin:
>  I could have added a test to check the output of K512 for a few values but I don't like this type of tests as they break when the model is updated (which would happen if they update k512).

If that's a part that may get updated often, potentially by the external package, and does not really matter so sure. If that's something that *should not* change unnoticed, I can assure you it is reaaaaally useful :D it prevents a lot of unforeseen bugs :D 
anyway in this case, no need

nevergrad/functions/images/imagelosses.py
```diff
@@ -0,0 +1,41 @@
...
+        super().__init__()
+        import os
+        if os.name != 'nt':
+            from koncept.models import Koncept512 as K512_model  # type: ignore
+            self.koncept = K512_model()
+        else:
+            self.koncept = None
+
+    def __call__(self, img: np.ndarray) -> float:
+        loss = self.koncept.assess(img) if self.koncept else np.zeros(1)
```
jrapin, line 38: i don't think it's a good idea to have a "silent" weird behavior, better explicitely raise an exception during the `__init__`
I'll edit the code myself though because it requires restructuring other parts of the code

nevergrad/functions/images/core.py
```diff
@@ -184,3 +178,59 @@ def make_folder_functions(
...
+                               pretrained=True, useGPU=use_gpu)
+
+        self.noise_shape = (1, 512)
+        if initial_noise is None:
+            initial_noise = np.random.normal(size=self.noise_shape)
+        assert initial_noise.shape == self.noise_shape, f'The shape of the initial noise vector was {initial_noise.shape}, it should be {self.noise_shape}'
+
+        array = ng.p.Array(init=initial_noise, mutable_sigma=mutable_sigma)
+        # parametrization
+        array.set_mutation(sigma=n_mutations)
```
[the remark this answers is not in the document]
brozi: oh good catch. That's a very bad name.

nevergrad/functions/images/core.py
```diff
@@ -184,3 +177,64 @@ def make_folder_functions(
...
+    initial_noise: np.ndarray
+        the initial noise of the GAN. It should be of dimension (1, 512). If None, it is defined randomly.
+    use_gpu: bool
+        whether to use gpus to compute the images
+    loss: ImageLoss
+        which loss to use for the images
+    mutable_sigma: bool
+        whether the sigma should be mutable
+    sigma: int
+        number of mutations
```
jrapin, line 199:
```suggestion
        standard deviation of the initial mutations
```

nevergrad/functions/images/core.py
```diff
@@ -184,3 +177,64 @@ def make_folder_functions(
...
+        the type of problem we are working on.
+    initial_noise: np.ndarray
+        the initial noise of the GAN. It should be of dimension (1, 512). If None, it is defined randomly.
+    use_gpu: bool
+        whether to use gpus to compute the images
+    loss: ImageLoss
+        which loss to use for the images
+    mutable_sigma: bool
+        whether the sigma should be mutable
+    sigma: int
```
jrapin, line 198:
```suggestion
    sigma: float
```

jrapin (approved):

PR #915, opened 2020-12-01
NGOpt8 uses DE if multiobjective

teytaud:
## Types of changes

## Checklist


nevergrad/optimization/optimizerlib.py
```diff
@@ -2102,6 +2102,8 @@ def _select_optimizer_cls(self) -> base.OptCls:
         # Extracting info as far as possible.
         assert self.budget is not None
         optimClass: base.OptCls
+        if not self.parametrization.descriptors.monoobjective and (self.noise_from_instrumentation or not self.has_noise):
+            optimClass = DE
```
jrapin, line 2106: That's gonna be more complicated than this, monoobjective descriptor is not reliable, we should wait for the first `tell` and modify the optimizer if need be :s

jrapin (changes requested):

nevergrad/optimization/optimizerlib.py
teytaud: I add a warning in "tell" in case monoobjective has not been set properly.

I think it's still much better than ignore the multiobjective nature of the problem when we have the info.
Agree that it would be possible, inside tell, to switch back to reinstantiating an optimizer... looks complicated though,
and the current hack is already quite useful.

jrapin: the descriptors were here for the time when it was impossible to use tell with a multiobjective loss. Now that this is possible, they should disappear and we definitely should not rely on them anymore. I am not in favor of telling people to set the tag since they are bound to disappear asap :s

jrapin:
Replaced by #922 if ok with you

teytaud:
Deprecated

PR #929, opened 2020-12-05
sequential or fully parallel ML.

teytaud:
This benchmark has not been much used for the moment, let us make it shorter and clearer.

## Types of changes

## Checklist


jrapin (commented):
Why do you remove all middle ground between one shot and sequential?
I feel this is not representative at all of mltuning cases

teytaud:
> Why do you remove all middle ground between one shot and sequential?
> I feel this is not representative at all of mltuning cases

Just because I feel these two extreme cases are the priority these days.

jrapin:
> > Why do you remove all middle ground between one shot and sequential?
> > I feel this is not representative at all of mltuning cases
> 
> Just because I feel these two extreme cases are the priority these days.

why so? typical workflows in my understanding would have something like 4 passes

teytaud:
> > > Why do you remove all middle ground between one shot and sequential?
> > > I feel this is not representative at all of mltuning cases
> > 
> > 
> > Just because I feel these two extreme cases are the priority these days.
> 
> why so? typical workflows in my understanding would have something like 4 passes

Ok then in a perfect world I'd like to have three cases so that we can see what happens but ok it's slow enough, I appreciate the recent speed, so let's have [1, budget//4] and [budget]; after all the subplots will distinguishes the different cases in the dashboard.

PR #930, opened 2020-12-05
Add a constrained counterpart of YABBOB

teytaud:
## Types of changes

## Checklist

- [x ] I have read the [**CONTRIBUTING**](https://facebookresearch.github.io/nevergrad/contributing.html) document and completed the CLA (see [**CLA**](https://facebookresearch.github.io/nevergrad/contributing.html#contributor-license-agreement-cla)).
- [x ] All tests passed, and additional code has been covered with new tests.


nevergrad/benchmark/experiments.py
```diff
@@ -570,6 +618,24 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
...
+        if constraints > 3 and constraints <= 7:
+            func.parametrization.register_cheap_constraint(_ball_float)
+        if constraints > 4:
+            func.parametrization.register_cheap_constraint(_positive_sum)
+        if constraints > 5:
+            func.parametrization.register_cheap_constraint(_positive_diff)
+        if constraints > 6:
+            func.parametrization.register_cheap_constraint(_positive_second_diff)
+        if constraints > 7:
+            func.parametrization.register_cheap_constraint(_ball)
```
jrapin, line 638: please restructurate using some kind of for loop, this is too harad to read.
using something like constraints = [_positive_sum_float, ...]
and then selecting the ones to register should help.

nevergrad/benchmark/experiments.py
```diff
@@ -43,6 +43,54 @@
 # fmt: off
 
 
+def _positive_sum_float(data: np.ndarray) -> float:  # This one is Boolean.
```
jrapin, line 46: it's not
```suggestion
def _positive_sum_float(data: np.ndarray) -> float:
```

nevergrad/benchmark/experiments.py
```diff
@@ -43,6 +43,54 @@
...
+def _positive_second_diff_float(data: np.ndarray) -> float:
+    if not isinstance(data, np.ndarray):
+        raise ValueError(f"Unexpected inputs as np.ndarray, got {data}")
+    return float(2 * np.sum(data[1::2]) - 3 * np.sum(data[::2]))
+
+
+def _ball_float(data: np.ndarray) -> float:
+    if not isinstance(data, np.ndarray):
+        raise ValueError(f"Unexpected inputs as np.ndarray, got {data}")
+    return float(np.sum(np.square(data))) - float(len(data)) - float(np.sqrt(len(data)))  # Most points violate the constraint.
```
jrapin, line 67: that's way too much repeated code.

You should go with something like
```python
class _Constraint

    def __init__(self, name: str) -> None:
        ...

    def __call__(self, data: np.ndarray) -> float:
        ...
```
it will still be a lot of ifs, but that would be 2 lines per case instead of 6

jrapin (changes requested):

nevergrad/benchmark/experiments.py
teytaud: Done.

nevergrad/benchmark/experiments.py
teytaud: Done.

nevergrad/benchmark/experiments.py
```diff
@@ -43,6 +43,57 @@
...
+
+    def __call__(self, data: np.ndarray) -> float:
+        if not isinstance(data, np.ndarray):
+            raise ValueError(f"Unexpected inputs as np.ndarray, got {data}")
+        return self._function(data)
+
+
+class _positive_sum_float(_Constraint):
+    def _function(self, data: np.ndarray) -> float:
+        return float(np.sum(data))
```
jrapin, line 59: This is not what I expressed, since this is just as long. Please do this instead, which factorizes a lot of code:

```suggestion
class _Constraint:

    def __init__(self, name: str, as_bool: bool) -> None:
         self.name = name
         self.as_bool = as_bool

    def __call__(self, data: np.ndarray) -> tp.Union[bool, float]:
        if not isinstance(data, np.ndarray):
            raise ValueError(f"Unexpected inputs as np.ndarray, got {data}")
        if self.name == "positive_sum_float":
            value = float(np.sum(data))
        elif self.name ==...:
             ...
         else:
             raise NotImplentedError(f"Unknown function {self.name}")
         return value > 0 if self.as_bool else value
```

Also I would consider renaming the function, positive_sum_float is not positive, so probably just `sum`, and since summing would get a float, no need for `_float`.  And the `as_bool` (or `positive_indicator`?) flag can provides the equivalent in boolean form without more code.

nevergrad/benchmark/experiments.py
```diff
@@ -570,6 +621,13 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
         for num_blocks in [1]
         for d in ([100, 1000, 3000] if hd else [2, 10, 50])
     ]
+    assert constraints < 8, "We have only four possible constraints."
+    for func in functions:
+        constraints_list = [_positive_sum_float(), _positive_diff_float(), _positive_second_diff_float(), _ball_float(),
+            _positive_sum(), _positive_diff(), _positive_second_diff(), _ball()]
+        for u in range(8):
+            if constraints > u and constraints <= u+4:
+                func.parametrization.register_cheap_constraint(constraints_list[u])
```
jrapin, line 630: It's much better, but I still don't get why there is this weird u, u+4 case, can you explain so that we find a clearer option?

nevergrad/benchmark/experiments.py
```diff
@@ -586,6 +644,18 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
...
 
+@registry.register
+def yaconstrainedbbob(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Counterpart of yabbob with higher dimensions."""
+    for i in range(8):
+        internal_generator = yabbob(seed, constraints=i)
+        j = 0
+        for xp in internal_generator:
+            j = j + 1
+            if j % 8 == i:  # We use only one test case out of 4, due to computational cost.
```
jrapin, line 655: I'm not sure I would hardcode 8 everywhere, don't you plan on adding more cases?

jrapin (changes requested):

nevergrad/benchmark/experiments.py
teytaud: We just want to have various combinations of boolean and float constraints. With this simple if we get 8 different combinations and all cases: all boolean, all float, and mixed.

nevergrad/benchmark/experiments.py
teytaud: No plan for now, but using len(...) is simple, let me do this.

teytaud: +the constant 4 does not have to be equal to len(...)/2, that's just an arbitrary choice.

nevergrad/benchmark/experiments.py
jrapin: I'm still unclear on what that does, especially since the order in the nested for loops creating the constraints has never been intuitive to me (is a  sequence of bool = True then False and so on? or all True, then all False?)
Wouldn't that be more explicit to use the name and the boolean directly, and maybe two for loops, instead of the objects themselves ? not sure, since I still don't get it
Also, is it expected that u == 0 has no constraint?

teytaud: I have checked the order of the nested "for for" in the list for being sure that it does what I want.
I have, on purpose, modified the code for not multiplying the number of functions - for the moment we can still afford modifications (I'd like the problems to be somehow stable over 2021).

Please note that for lists, list comprehension are deterministic and the order matters, as explained in https://stackoverflow.com/questions/1198777/double-iteration-in-list-comprehension, "nested" for are in some sense horizontal.

teytaud: My last modification was not very good for diversity of cases, I've modified back while keeping the new style of that code.

nevergrad/benchmark/experiments.py
```diff
@@ -43,6 +43,28 @@
...
+
+    def __call__(self, data: np.ndarray) -> tp.Union[bool, float]:
+        if not isinstance(data, np.ndarray):
+            raise ValueError(f"Unexpected inputs as np.ndarray, got {data}")
+        if self.name == "positive_sum":
+            value = float(np.sum(data))
+        elif self.name == "positive_diff":
+            value = float(np.sum(data[::2]) - np.sum(data[1::2]))
+        elif self.name == "positive_second_diff":
+            value = float(2 * np.sum(data[1::2]) - 3 * np.sum(data[::2]))
```
jrapin, line 60:
```suggestion
        if self.name == "sum":
            value = float(np.sum(data))
        elif self.name == "diff":
            value = float(np.sum(data[::2]) - np.sum(data[1::2]))
        elif self.name == "second_diff":
            value = float(2 * np.sum(data[1::2]) - 3 * np.sum(data[::2]))
```
it's definitely not positive so the naming is confusing

nevergrad/benchmark/experiments.py
```diff
@@ -573,6 +595,17 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
         for num_blocks in [1]
         for d in ([100, 1000, 3000] if hd else [2, 10, 50])
     ]
+    max_num_constraints = 4
+    constraints_list: tp.List[tp.Any] = [_Constraint(name, as_bool)
+        for as_bool in [False, True]
+        for name in ["positive_sum", "positive_diff", "positive_second_diff", "ball"]
```
jrapin, line 601:
```suggestion
        for name in ["sum", "diff", "second_diff", "ball"]
```

nevergrad/benchmark/experiments.py
```diff
@@ -573,6 +595,17 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
...
+        ]
+    assert constraints < len(constraints_list) + max_num_constraints, (
+        "constraints should be in 0, 1, ..., {len(constraints_list) + max_num_constraints - 1} (0 = no constraint).")
+    for func in functions:
+        for u in range(len(constraints_list)):
+            if u < constraints and constraints - max_num_constraints <= u:  # we add the constraints with index u in {constraints-max, ..., constraints-1}
+                func.parametrization.register_cheap_constraint(constraints_list[u % len(constraints_list)])
```
jrapin, line 608: this is definitely still very confusing for me
`constraint_list` should probably just be `constraints`
then `constraints` should be something else but I can't get a hold of what it would mean `constraint_case` maybe?

Also do you need to add a descriptor for the cases? not sure

nevergrad/benchmark/experiments.py
```diff
@@ -715,7 +754,7 @@ def constrained_illconditioned_parallel(seed: tp.Optional[int] = None) -> tp.Ite
         in [True, False]
     ]
     for func in functions:
-        func.parametrization.register_cheap_constraint(_positive_sum)
+        func.parametrization.register_cheap_constraint(_Constraint("positive_sum", as_bool=False))
```
jrapin, line 757:
```suggestion
        func.parametrization.register_cheap_constraint(_Constraint("sum", as_bool=False))
```

nevergrad/benchmark/experiments.py
teytaud: The constraint is checking if this is positive, hence the naming -- but ok for changing.

nevergrad/benchmark/experiments.py
```diff
@@ -585,6 +622,18 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
...
+@registry.register
+def yaconstrainedbbob(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Counterpart of yabbob with higher dimensions."""
+    for i in range(8):
+        internal_generator = yabbob(seed, constraints=i)
+        j = 0
+        for xp in internal_generator:
+            j = j + 1
+            if j % 8 == i:  # We use only one test case out of 4, due to computational cost.
+                yield xp
```
jrapin, line 634:
```suggestion
    step = 8 # only one test case out of 8, due to computational cost.
    slices = [itertools.islice(yabbob(seed, constraints=i), 0, None, step) for i in range(8)]
    return itertools.chain.from_iterables(slices)
```

nevergrad/benchmark/experiments.py
```diff
@@ -43,6 +43,32 @@
...
+            value = float(2 * np.sum(data[1::2]) - 3 * np.sum(data[::2]))
+        elif self.name == "ball":
+            value = float(np.sum(np.square(data))) - float(len(data)) - float(np.sqrt(len(data)))  # Most points violate the constraint.
+        else:
+            raise NotImplementedError(f"Unknown function {self.name}")
+        return value > 0 if self.as_bool else value
+
+
+class MissingBenchmarkPackageError(ModuleNotFoundError):
+    pass
```
jrapin, line 69:
```suggestion
```

nevergrad/benchmark/experiments.py
teytaud: There is already a variable termed constraints.

teytaud: I try to refactor a little bit.

jrapin: yes exactly, my point is "why". I don't understand *what* `constraints` currently represent, so the name should maybe be changed to clarify

nevergrad/benchmark/experiments.py
```diff
@@ -585,6 +618,14 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
 
 
 
+@registry.register
+def yaconstrainedbbob(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Counterpart of yabbob with higher dimensions."""
+    step = 8 # only one test case out of 8, due to computational cost.
+    slices = [itertools.islice(yabbob(seed, constraints=i), 0, None, step) for i in range(8)]
+    return itertools.chain.from_iterables(slices)
```
jrapin, line 626:
```suggestion
    return itertools.chain.from_iterable(slices)
```

nevergrad/benchmark/experiments.py
teytaud: Yo! Looks better.

nevergrad/benchmark/experiments.py
```diff
@@ -572,6 +610,14 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
                     yield xp
 
 
+@registry.register
+def yaconstrainedbbob(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Counterpart of yabbob with higher dimensions."""
+    step = 8 # only one test case out of 8, due to computational cost.
+    slices = [itertools.islice(yabbob(seed, constraints=i), 0, None, step) for i in range(step)]
+    return itertools.chain.from_iterable(slices)
```
jrapin, line 618:
```suggestion
    return itertools.chain(*slices)
```

jrapin (approved):

nevergrad/benchmark/experiments.py
```diff
@@ -572,6 +610,14 @@ def yabbob(seed: tp.Optional[int] = None, parallel: bool = False, big: bool = Fa
                     yield xp
 
 
+@registry.register
+def yaconstrainedbbob(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Counterpart of yabbob with higher dimensions."""
+    step = 8 # only one test case out of 8, due to computational cost.
+    slices = [yabbob(seed, constraint_case=i) for i in range(step)]
```
jrapin, line 617:
```suggestion
    cases = 8 # only one test case out of 8, due to computational cost.
    slices = [yabbob(seed, constraint_case=i) for i in range(cases)]
```

jrapin (approved):

PR #934, opened 2020-12-07
Use an auxiliary solver for constraints

teytaud:
## Types of changes

In case the optimization method did not suggest anything satisfactory after half the constraint budget, we switch to an auxiliary constraint optimization method. 

## Checklist


teytaud:
In a perfect world we would like to detect when we should do this.
For example, I expect this to be very necessary for Powell, not for the 1+1, maybe not for DE...

nevergrad/parametrization/core.py
```diff
@@ -263,6 +263,23 @@ def set_name(self: P, name: str) -> P:
 
     # %% Constraint management
 
+    def constraint_violation(self) -> float:
+        """Returns maximum of constraint values.u
+        Uses 1 for True, constraint violation for float, negative numbers ok
+        when inside the satisfaction domain.
```
jrapin, line 269: why do you use a max, a wouldnt a sum provide more information?
Also I don't understand the docstring at all, can you clarify ? negative values are violation as far as I understand, so not OK

jrapin (commented):
I'll need more time to understand this, I'll be back on it later, please ping me if need be

nevergrad/parametrization/core.py
teytaud: Yes a sum provides more information. However, here, it would not work.

We want something which solves all the constraints, not something which solves the sum of the constraints.
This is just not equivalent.

We should use the exact value of each constraint for learning a model of constraints, and use it: for the moment this is not included in Nevergrad. This is the main next step. This is not for the current PR, and using the sum as a substitute would be a bad idea.

I improve the docstring.

teytaud: I also add an explanation of the critical "tell".

jrapin:
why would it need 990?

teytaud:
> why would it need 990?

Because if num_ask is close to the budget limit we should immediately switch to the auxiliary solver.

nevergrad/optimization/base.py
```diff
@@ -430,28 +430,37 @@ def ask(self) -> p.Parameter:
         current_num_ask = self.num_ask
         # tentatives if a cheap constraint is available
         # TODO: this should be replaced by an optimization algorithm.
-        max_trials = self._constraints_manager.max_trials
-        for k in range(max_trials):
+        max_trials = max(
+            1, self._constraints_manager.max_trials // 2  if self.budget is None else 0
```
teytaud, line 434: This could be better using PR 990 but let's do this for now.

teytaud:
deprecated, 1047 is the current version.
1047 might be merged or not later, depending on compatibility with other works in progress.

PR #940, opened 2020-12-09
Enable multiple layers for mujoco

herilalaina:
## Types of changes

@teytaud @LaurentMnr95 

## Checklist


jrapin:
Hi!
Thanks for taking this over.
I can't find the initial PR, but as I had stated there, this must not be separate `Neuro-` classes, but only a parameter `policy_dim` parameter to provide to the base classes. This will avoid an exponential explosion of classes when adding more options (eg: noise #844 which will need to implement it as a parameter as well).
Do you understand how to do this? Or what help would you need?

nevergrad/functions/control/core.py
```diff
@@ -39,19 +39,29 @@ class BaseFunction(ExperimentFunction):
...
+        self.add_descriptors(num_rollouts=num_rollouts)
+        if not hasattr(self, "layer_rescaling_coef"): self.layer_rescaling_coef = np.ones(len(self.policy_dim) - 1)
         self._descriptors.pop("random_state", None)  # remove it from automatically added descriptors
 
-    def _simulate(self, x: np.ndarray) -> float:
+    def _simulate(self, x: tp.Tuple) -> float:
         env = GenericMujocoEnv(env_name=self.env_name,
-                               state_mean=self.state_mean,
-                               state_std=self.state_std,
+                               state_mean=self.state_mean if len(self.policy_dim) == 2 else None,
```
teytaud, line 56: This I don't understand... why should the presence of a neural net interfere with the use of this normalization ?

nevergrad/functions/control/core.py
```diff
@@ -39,19 +39,29 @@ class BaseFunction(ExperimentFunction):
         random state for reproducibility in Gym environment.
     """
 
-    def __init__(self, num_rollouts: int, random_state: tp.Optional[int] = None) -> None:
-        super().__init__(self._simulate, p.Array(shape=self.policy_dim))
+    def __init__(self, num_rollouts: int, activation: str = "tanh", random_state: tp.Optional[int] = None) -> None:
+        list_parametrizations = [p.Array(shape=(a, b)).set_name(r"layer_{a}_{b}") for a, b in
+                                 zip(self.policy_dim[:-1], self.policy_dim[1:])]
+        parametrization = p.Instrumentation(*list_parametrizations).set_name(self.env_name)
```
teytaud, line 45: Excellent for our work on structured optimization --- cool!

nevergrad/functions/control/core.py
herilalaina: You are right, it shouldn't

herilalaina:
> Hi!
> Thanks for taking this over.
> I can't find the initial PR, but as I had stated there, this must not be separate `Neuro-` classes, but only a parameter `policy_dim` parameter to provide to the base classes. This will avoid an exponential explosion of classes when adding more options (eg: noise #844 which will need to implement it as a parameter as well).
> Do you understand how to do this? Or what help would you need?

Yes clear enough. Fixes coming :)

herilalaina:
@jrapin that error seems to be unrelated to the PR (seg fault in `imagesxp.ImageFromPGAN()`)

jrapin:
> @jrapin that error seems to be unrelated to the PR (seg fault in `imagesxp.ImageFromPGAN()`)

@brozi any idea where that comes from?  `ImportError: cannot import name 'multi_gpu_model' from 'tensorflow.keras.utils'`

brozi:
> > @jrapin that error seems to be unrelated to the PR (seg fault in `imagesxp.ImageFromPGAN()`)
> 
> @brozi any idea where that comes from? `ImportError: cannot import name 'multi_gpu_model' from 'tensorflow.keras.utils'`

Not sure it fixes it but you can try adding this at the end of bench.txt
tensorflow==2.3.1
tensorflow-estimator==2.3.0

teytaud:
> > > @jrapin that error seems to be unrelated to the PR (seg fault in `imagesxp.ImageFromPGAN()`)
> > 
> > 
> > @brozi any idea where that comes from? `ImportError: cannot import name 'multi_gpu_model' from 'tensorflow.keras.utils'`
> 
> Not sure it fixes it but you can try adding this at the end of bench.txt (and remove the previous tensorflow line). It passes in this PR which is built on top of the current master branch. #964
> tensorflow==2.3.1
> tensorflow-estimator==2.3.0

yes that seems to fix the issue for me

herilalaina:
@brozi Thanks :)
@teytaud @jrapin I refractored mujoco variants (rescaling state values, noise, deterministic evaluation, layer dimension) in `BaseFunction` class. Let me know if I have missed any options

nevergrad/functions/control/mujoco.py
```diff
@@ -20,27 +20,49 @@ class GenericMujocoEnv:
...
 
-    def __call__(self, x):
-        """Compute average cummulative reward of a given policy.
+    def __call__(self, layers):
+        """Compute loss (average cumulative negative reward) of a given policy.
         """
         returns = []
         for _ in range(self.num_rollouts):
             obs = self.env.reset()
+            print(obs)
```
jrapin, line 56:
```suggestion
```

nevergrad/functions/control/test_mujoco.py
```diff
@@ -3,24 +3,49 @@
...
     state_mean = [0, 0]
     state_std = [1, 1]
 
+
 def test_gym() -> None:
-    func = MountainCarContinuous(num_rollouts=5)
+    for intermediate_layer_dim in [None, (10,), (5, 5)]:
+        for noise_level in [0.0, 0.9]:
+            for deterministic_sim in [True, False]:
+                for states_normalization in [True, False]:
```
jrapin, line 22: this tests now takes 10s which is a bit long for a test :s . Sometimes there is nothing we can do about it (unfortunately :s), but here can we either or both:
- use fewer rollout
- not test the whole grid of parameters combinations, picking a few test cases should be enough to test each elements if they are uncoupled enough

Also, please use [pytest.mark.parametrize(https://docs.pytest.org/en/stable/parametrize.html) for providing the test cases, since this splits the test and allows to figure out more easily which cases break (+ it will be more readable, with fewer indented for loops)

nevergrad/functions/control/test_mujoco.py
```diff
@@ -3,24 +3,49 @@
...
 
-    x = func.parametrization.sample().value
-    assert not np.array_equal(x, [[0., 0.]])
-    value = func(func.parametrization.sample().value)
-    assert value != 0
+    for module in ["Ant", "Swimmer", "HalfCheetah", "Hopper", "Walker2d", "Humanoid"]:
+        for intermediate_layer_dim in [None, (50,)]:
+            for noise_level in [0.0, 0.9]:
+                for deterministic_sim in [True, False]:
+                    for states_normalization in [True, False]:
```
jrapin, line 46: same as above, however since this test will only run for mujoco users which will be interested in making sure all the combinations work, you can keep all combinations.
The simplest way would be to stack multiple pytest.mark.parametrize in this way:
```python
@pytest.mark.parametrize("x", [0, 1])
@pytest.mark.parametrize("y", [2, 3])
def test_foo(x, y):
    pass
```

nevergrad/benchmark/experiments.py
```diff
@@ -1040,6 +1040,31 @@ def control_problem(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+    """MuJoCo testbed. Learn neural policies."""
+    seedg = create_seed_generator(seed)
+    num_rollouts = 1
+    funcs = [control.Swimmer(num_rollouts=num_rollouts, intermediate_layer_dim=(50,), random_state=seed),
+             control.HalfCheetah(num_rollouts=num_rollouts, intermediate_layer_dim=(50,), random_state=seed),
+             control.Hopper(num_rollouts=num_rollouts, intermediate_layer_dim=(50,), random_state=seed),
+             control.Walker2d(num_rollouts=num_rollouts, intermediate_layer_dim=(50,), random_state=seed),
+             control.Ant(num_rollouts=num_rollouts, intermediate_layer_dim=(50,), random_state=seed),
+             control.Humanoid(num_rollouts=num_rollouts, intermediate_layer_dim=(50,), random_state=seed)
+             ]
```
jrapin, line 1054: always benefinicial on the long run to avoid duplicated code ;) (easier to maintain)

```suggestion
    funcs = [
        Env(num_rollouts=num_rollouts, intermediate_layer_dim=(50,), random_state=seed) for env in [
            control.Swimmer, control.HalfCheetah, control.Hopper, control.Walker2d, control.Ant, control.Humanoid
        ]
    ]
```

nevergrad/benchmark/experiments.py
```diff
@@ -1040,6 +1040,31 @@ def control_problem(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+
+    optims = ["CMA", "NGOpt4", "DiagonalCMA", "NGOpt8", "MetaModel", "chainCMAPowell"]
+
+    for budget in [50, 500, 5000, 10000, 20000, 35000, 50000, 100000, 200000]:
+        for num_workers in [1]:
+            if num_workers < budget:
```
jrapin, line 1060: you can remove those two lines

nevergrad/benchmark/experiments.py
```diff
@@ -1023,7 +1023,7 @@ def control_problem(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
     for sigma, func in zip(sigmas, funcs):
         f = func.copy()
         param: ng.p.Array = f.parametrization.copy()  # type: ignore
-        param.set_mutation(sigma=sigma).set_name(f"sigma={sigma}")
+        param[0][0].set_mutation(sigma=sigma).set_name(f"sigma={sigma}") # type: ignore
```
jrapin, line 1026: @teytaud @LaurentMnr95 
Do we want to set the same sigma to each of the layers?
If so we need something like this (a bit hacky :s)
```suggestion
        for array in param[0]._content.values():
            array.set_mutation(sigma=sigma)
        param.set_name(f"sigma={sigma}")
```
@herilalaina the name of the mutation should be set on the full parametrization (param)
Also, since this is not an Array anymore, please update the typing (to ng.p.Instrumentation?) just above

nevergrad/functions/control/core.py
```diff
@@ -88,25 +121,63 @@ def evaluation_function(self, *recommendations: p.Parameter) -> float:
 
 class Ant(BaseFunction):
     env_name = 'Ant-v2'
-    policy_dim = (8, 111)
+    policy_dim: tp.Tuple[int, ...] = (111, 8)
```
jrapin, line 124: how come this has changed shape? I'm afraid this could actually change how state mean is processed, but I am not very familiar with this so I may be wrong
@teytaud maybe this needs cross-checking

nevergrad/functions/control/core.py
```diff
@@ -35,25 +35,58 @@ class BaseFunction(ExperimentFunction):
...
-    def _simulate(self, x: np.ndarray) -> float:
-        env = GenericMujocoEnv(
-            env_name=self.env_name,
-            state_mean=self.state_mean,
-            state_std=self.state_std,
-            num_rollouts=self.num_rollouts,
-            random_state=self.random_state,
-        )
-        return env(x)
+    def _simulate(self, x: tp.Tuple) -> float:
```
jrapin, line 77: There is something weird here
Given your instrumentation, you should have
```suggestion
    def _simulate(self, *x: np.ndarray) -> float:
```

nevergrad/functions/control/core.py
```diff
@@ -35,25 +35,58 @@ class BaseFunction(ExperimentFunction):
...
+        env = GenericMujocoEnv(env_name=self.env_name,
+                               state_mean=self.state_mean if self.states_normalization else None,
+                               state_std=self.state_std if self.states_normalization else None,
+                               num_rollouts=self.num_rollouts,
+                               activation=self.activation,
+                               layer_rescaling_coef=self.layer_rescaling_coef,
+                               noise_level=self.noise_level,
+                               random_state=self.parametrization.random_state)
+        env.env.seed(self.random_state if self.deterministic_sim else self.parametrization.random_state.randint(10000))
+        loss = env(x[0])
```
jrapin, line 87: I may be incorrect, but I expect this should be:
```suggestion
        loss = env(*x)
```
otherwise that would be 1-layer only all the time

nevergrad/functions/control/mujoco.py
```diff
@@ -20,27 +20,49 @@ class GenericMujocoEnv:
...
+        if self.activation == 'tanh':
+            return np.tanh(x)
+        elif self.activation == "sigmoid":
+            return 1. / (1 + np.exp(-x))
+        else:
+            raise NotImplementedError(r"Activation {self.activation} not implemented.")
 
-    def __call__(self, x):
-        """Compute average cummulative reward of a given policy.
+    def __call__(self, layers):
```
jrapin, line 50: Again, there is some weird mismatch about how parameters are handled, can you please crosscheck?

```suggestion
    def __call__(self, *layers: np.ndarray):
```

jrapin (changes requested):
Hi
Thanks for the work.
I have some concerns with regards to how parameters are handled, I am afraid it may actually be running on 1 layer only, even when other layers are provided (actually I would expect the code to bug, so it does not, so maybe I am wrong, but I'd like to understand why)

nevergrad/functions/control/core.py
herilalaina: I changed the order to match (input layer, ..., output layer) for the sake of clarity. It shouldn't interfere with state_mean because we normalize only the input of the first layer. But cross-checking always helps :)

nevergrad/functions/control/core.py
herilalaina: What I have understood is that, for p.Instrumentation, parameter `x` is a tuple of (args, kwargs) ? thus, x[0] is a list of np.ndarray. I changed parametrization from p.Intrumentation to p.Tuple to be less confusing.

nevergrad/functions/control/mujoco.py
herilalaina: As described above, `layers` is now a tuple of np.ndarray.

herilalaina:
Happy new year :-) 
I addressed your comments. Tests are now failing on mixsimulator (after merging with master).

nevergrad/functions/control/core.py
jrapin: @teytaud how can we make sure this works as expected?

jrapin (commented):
Hey,
Looks like the current bug is unrelated (an issue with a dependency which makes our test fail :s) I'll try to silence it.
Overall that looks good now, I'd just like to validate the reordering of the shapes with @teytaud to be extra-sure

nevergrad/functions/control/core.py
teytaud: Can you define "as expected" ? <<parametrization: p.Tuple = p.Tuple(*list_parametrizations).set_name(self.env_name)>> is perfect to me, in the sense "this is a tuple with one array per layer". 

In terms of impact with the state mean: this is unrelated, the state mean and std are in the domain of the control, not in the domain of the weights.

teytaud: (we might consider other parametrizations, in particular for progressive methods, but this will be for another PR...)

jrapin (approved):

nevergrad/functions/control/mujoco.py
```diff
@@ -20,26 +20,49 @@ class GenericMujocoEnv:
...
-    def __init__(self, env_name, state_mean, state_std, num_rollouts, random_state):
-        self.mean = np.array(state_mean)
-        self.std = np.array(state_std)
+    def __init__(self, env_name, state_mean, state_std, num_rollouts,
+                 activation, layer_rescaling_coef, noise_level, random_state):
+        self.mean = state_mean
+        self.std = state_std
         self.env = gym.make(env_name)
         self.num_rollouts = num_rollouts
-        self.env.seed(random_state)
```
jrapin, line 32: I was ready to merge, however I just noticed that the `random_state` has a very different meaning than before, @herilalaina is that expected?

herilalaina: Yes, this is expected. Seeding of gym environment is now handled in `_simulate` (according to `deterministic_sim`). But I still have to pass random generator to `GenericMujocoEnv` for noise generation.

jrapin:
@teytaud as the heaviest user of benchmarks, I let you merge if you are fine with this, on my side all seems good ;)

jrapin:
Merged, thanks for all the hard work @herilalaina

PR #948, opened 2020-12-12
More image problems, including MOO

teytaud:
## Types of changes

We need real MOO problems.
A second modification is coming, with a specifically MOO evaluation function.

## Checklist


nevergrad/functions/images/imagelosses.py
```diff
@@ -45,3 +86,12 @@ def __init__(self, reference: tp.Optional[np.ndarray] = None) -> None:
     def __call__(self, img: np.ndarray) -> float:
         loss = self.koncept.assess(img)
```
brozi, line 87: Koncept 512 takes images of dimension (x, y, 3) and containing float between 0 and 255

teytaud: R, G, B ?

brozi (commented):
generate_images returns something of size (n_images, 3, 512, 512). It is permuted to match k512's format with the rgb dim at the end.

nevergrad/functions/images/imagelosses.py
```diff
@@ -1,35 +1,102 @@
...
 
 
-class SumAbsoluteDifferences(ImageLoss):
+class ImageLossWithReference(ImageLoss):
     def __init__(self, reference: np.ndarray) -> None:
         if reference is None:
             raise ValueError("A reference is required")
         self.reference = reference
         super().__init__(reference)
+        assert len(self.reference.shape) == 3, self.reference.shape
```
brozi, line 29: If you want to make sure the format of the reference is the same for every loss, maybe add something about the values. 
```suggestion
        assert len(self.reference.shape) == 3, self.reference.shape
        assert len(self.reference.min()) >= 0 and len(self.reference.max()) <= 255, f"The reference should contain values between 0 and 255, they were between {self.reference.min()} and {self.reference.max()}"
```

nevergrad/functions/images/imagelosses.py
```diff
@@ -1,35 +1,102 @@
...
+        img1 = torch.clamp(torch.Tensor(img1).permute(0, 3, 1, 2) / 255.0, 0, 1) * 2.0 - 1.0
+        return self.loss_fn(img0, img1)
+
+
+@registry.register
+class LpipsVgg(LpipsAlex):
+    def __init__(self, reference: np.ndarray) -> None:
+        super().__init__(reference)
+        self.loss_fn = lpips.LPIPS(net="vgg")
+
```
brozi, line 71: I think it would make more sense to have a class Lpips and two classes inheriting from it for Alexnet and VGG
```suggestion
class Lpips(ImageLossWithReference):
    def __init__(self, reference: np.ndarray, net: str) -> None:
        super().__init__(reference)
        self.loss_fn = lpips.LPIPS(net=net)

    def __call__(self, img: np.ndarray) -> float:
        assert img.shape[2] == 3
        assert len(img.shape) == 3
        img0 = np.expand_dims(img, 0)
        assert img0.shape[3] == 3
        assert img0.shape[0] == 1
        assert len(img0.shape) == 4
        img1 = np.expand_dims(self.reference, 0)
        img0 = torch.clamp(torch.Tensor(img0).permute(0, 3, 1, 2) / 255.0, 0, 1) * 2.0 - 1.0
        img1 = torch.clamp(torch.Tensor(img1).permute(0, 3, 1, 2) / 255.0, 0, 1) * 2.0 - 1.0
        return self.loss_fn(img0, img1)


@registry.register
class LpipsVgg(Lpips):
    def __init__(self, reference: np.ndarray) -> None:
        super().__init__(reference, net="vgg")


@registry.register
class LpipsAlex(Lpips):
    def __init__(self, reference: np.ndarray) -> None:
        super().__init__(reference, net="alex")
```

brozi (changes requested):
I suggested a few changes. Most are just to make the code clearer, one is about a bug (lack of - sign for the blur loss)

nevergrad/functions/images/imagelosses.py
teytaud: I guess the "len" here is a typo, but yes that makes sense.

teytaud: Done.

nevergrad/functions/images/imagelosses.py
```diff
@@ -43,5 +110,33 @@ def __init__(self, reference: tp.Optional[np.ndarray] = None) -> None:
...
+    This estimates bluriness
+    """
+
+    def __init__(self, reference: tp.Optional[np.ndarray] = None) -> None:
+        super().__init__()  # No reference needed! Just blur estimation.
+
+    def __call__(self, img: np.ndarray) -> float:
+        assert img.shape[2] == 3
+        assert len(img.shape) == 3
+        return cv2.Laplacian(img, cv2.CV_64F).var()
```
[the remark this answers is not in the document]
teytaud: Done.

nevergrad/functions/images/imagelosses.py
teytaud: Right: done.

docs/conf.py
```diff
@@ -17,60 +17,64 @@
 
 import os
 import sys
-sys.path.insert(0, os.path.abspath('..'))
+
+sys.path.insert(0, os.path.abspath(".."))
```
jrapin, line 21: I guess this happened because you ran black on the docs, which is not supposed to happen.
That's fine by me for this *but* this means you did not use the specs for `black` which are in the repository. To run black you need to be inside the repository `nevergrad` but outside the package `nevergrad`.

Please revert all changes that happened to mujoco and photonics, since they are in the exclude list for black (PRs in progress, I would rather not make it more difficult)

nevergrad/benchmark/experiments.py
```diff
@@ -1197,11 +1197,140 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+    # We optimize func_blur or func_brisque and check performance on func_iqa.
+    func_iqa = imagesxp.Image(loss=imagesxp.imagelosses.Koncept512)
+    func_blur = imagesxp.Image(loss=imagesxp.imagelosses.Blur)
+    func_brisque = imagesxp.Image(loss=imagesxp.imagelosses.Brisque)
+    
+    # TODO: add the proxy info in the parametrization.
+    for budget in [100 * 5 ** k for k in range(3)]:
+        for num_workers in [1]:
+            for algo in optims:
+                for func in [func_blur.copy(), func_brisque.copy()]:
```
jrapin, line 1256: why do you copy ? please avoid this, this would take time, especially within so many loops!

nevergrad/benchmark/experiments.py
```diff
@@ -1197,11 +1197,140 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+                    func.special_evaluation_function = func_iqa.evaluation_function
```
jrapin, line 1257: I guess I'll learn more of it later, this looks weird for now

nevergrad/benchmark/experiments.py
```diff
@@ -1197,11 +1197,140 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+
+@registry.register
+def image_quality(seed: tp.Optional[int] = None, cross_val: bool=False) -> tp.Iterator[Experiment]:
+    """Optimizing images for quality:
+    TODO
+    """
+    seedg = create_seed_generator(seed)
+    optims: tp.List[tp.Any] = get_optimizers("structured_moo")
+    if default_optims is not None:
+        optims = default_optims
```
jrapin, line 1270: I still do not like this, we can have a switch that overrides get_optimizers instead (for later)

nevergrad/benchmark/experiments.py
```diff
@@ -1197,11 +1197,140 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+            imagesxp.Image(loss=imagesxp.imagelosses.Blur),
+            imagesxp.Image(loss=imagesxp.imagelosses.Brisque),
+            ]
+    upper_bounds: tp.List[tp.Any] = [func(func.parametrization.sample().value) for func in funcs]
+    # TODO: add the proxy info in the parametrization.
+    for budget in [100 * 5 ** k for k in range(3)]:
+        for num_workers in [1]:
+            for algo in optims:
+                mofuncs = fbase.multi_experiments(funcs, upper_bounds=upper_bounds, no_crossval=[1, 2], pareto_size=16) if cross_val else [fbase.MultiExperiment(funcs, upper_bounds=upper_bounds)]
+                for func in mofuncs:
```
jrapin, line 1284: as usual, please create the functions at the lower level, because that more time that all the rest

nevergrad/benchmark/experiments.py
```diff
@@ -1197,11 +1197,140 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+    
+        # Creating a reference value.
+        base_value = func(func.parametrization.sample().value)
+        mofuncs = fbase.multi_experiments([func, func_blur, func_iqa], upper_bounds=[base_value, base_blur_value, 100.], no_crossval=[1, 2], pareto_size=16) if cross_val else [
+                fbase.MultiExperiment([func, func_blur, func_iqa], upper_bounds=[base_value, base_blur_value, 100.])]
+        for budget in [100 * 5 ** k for k in range(3)]:
+            for num_workers in [1]:
+                for algo in optims:
+                    for mofunc in mofuncs:
+                        xp = Experiment(mofunc, algo, budget, num_workers=num_workers, seed=next(seedg))
```
jrapin, line 1318: can this be factorized with `image_quality` ? There is a lot of code that looks really similar, and factorizing will help avoiding spending time modifying the same thing everywhere when we will need to (that always happen...)

nevergrad/functions/base.py
```diff
@@ -65,7 +65,12 @@ def __new__(cls: tp.Type[EF], *args: tp.Any, **kwargs: tp.Any) -> EF:
         inst._descriptors["function_class"] = cls.__name__
         return inst  # type: ignore
 
-    def __init__(self: EF, function: tp.Callable[..., tp.Loss], parametrization: p.Parameter) -> None:
+    def __init__(
+        self: EF,
+        function: tp.Callable[..., tp.Loss],
+        parametrization: p.Parameter,
+        special_evaluation_function: tp.Optional[tp.Callable[..., tp.Loss]] = None,
```
jrapin, line 72: `evaluation_function` is already the way to provide a special evaluation function, so in my point of view this is a duplicate, except to shortcut the need of subclassing. But I am not sure that it is that much more convenient, so I would like to avoid this extra parameter.

let's discuss this further offline, ad I'll try to provide another way

nevergrad/functions/base.py
```diff
@@ -254,6 +264,43 @@ def update_leaderboard(identifier: str, loss: float, array: np.ndarray, verbose:
         pass  # better avoir bugs for this
 
 
+def multi_experiments(
+    xps: tp.List[ExperimentFunction],
+    upper_bounds: tp.ArrayLike,
+    pareto_size: int,
+    no_crossval: tp.List[tp.Any] = [],
+) -> tp.List[ExperimentFunction]:
```
jrapin, line 272: I'll still need to understand this fully but here are a few things for now:
- instead of `xps` being xps and `no_crossval`  being indices, a more robust API would be to have the crossval xps on one side and the actual xps on another one.

Also, I would make this a `classmethod` of `MultiExperiment` with something like:

```python
@classmethod
def create_moo_crossvalidation_experiments(...)
```

nevergrad/functions/images/test_imagelosses.py
```diff
@@ -3,10 +3,22 @@
...
-    assert loss(np.ones((3, 2))) == 6
+    loss = imagelosses.SumAbsoluteDifferences(reference=np.zeros((3, 4, 3)))
+    assert loss(np.ones((3, 4, 3))) == 36.0
+
+
+def test_consistency_losses_with_oteytaud() -> None:
+    path = Path(__file__).with_name("headrgb_olivier.png")
+    image = PIL.Image.open(path).resize((256, 256), PIL.Image.ANTIALIAS)
+    data = np.asarray(image)[:, :, :3]  # 4th Channel is pointless here, only 255.
+    for loss_name, loss_class in imagelosses.registry.items():
```
jrapin, line 21: please use a pytest.mark.parametrized instead, so that we get independent tests

nevergrad/functions/images/imagelosses.py
```diff
@@ -1,35 +1,110 @@
...
         pass
 
     def __call__(self, img: np.ndarray) -> float:
-        raise NotImplementedError
+        raise NotImplementedError(f"__call__ undefined in class {type(self)}")
 
 
-class SumAbsoluteDifferences(ImageLoss):
-    def __init__(self, reference: np.ndarray) -> None:
+class ImageLossWithReference(ImageLoss):
```
jrapin, line 23: i'm still uneasy with keeping these two classes for not much, another option would be to include this code in `ImageLoss` under a condition `self.REQUIRES_REFERENCE` (a class attribute, which defaults to True since there are fewer cases which do not need it)

Not sure it is simpler though. I probably just do not like the name at all, `ReferenceImageLoss` would be more appropriate for a class name in my opinion

nevergrad/functions/images/imagelosses.py
```diff
@@ -1,35 +1,110 @@
...
+
     def __call__(self, x: np.ndarray) -> float:
         assert x.shape == self.domain_shape, f"Shape = {x.shape} vs {self.domain_shape}"
         value = float(np.sum(np.fabs(x - self.reference)))
         return value
 
 
+@registry.register
+class Lpips(ImageLossWithReference):
+    def __init__(self, reference: np.ndarray, net: str) -> None:
```
jrapin, line 49: without default value, this breaks the API and probably the tests which goes through all the cases

nevergrad/functions/images/imagelosses.py
```diff
@@ -43,5 +118,33 @@ def __init__(self, reference: tp.Optional[np.ndarray] = None) -> None:
...
+
+
+@registry.register
+class Brisque(ImageLoss):
+    """
+    This estimates the Brisque score (lower is better).
+    """
+
+    def __init__(self, reference: tp.Optional[np.ndarray] = None) -> None:
+        super().__init__()  # No reference needed! Just Brisque estimation.
```
jrapin, line 147:
```suggestion
```

nevergrad/functions/images/imagelosses.py
```diff
@@ -43,5 +118,33 @@ def __init__(self, reference: tp.Optional[np.ndarray] = None) -> None:
...
+
+
+@registry.register
+class Blur(ImageLoss):
+    """
+    This estimates bluriness
+    """
+
+    def __init__(self, reference: tp.Optional[np.ndarray] = None) -> None:
+        super().__init__()  # No reference needed! Just blur estimation.
```
jrapin, line 132: This is not needed
```suggestion
```

nevergrad/optimization/base.py
```diff
@@ -149,6 +150,14 @@ def _rng(self) -> np.random.RandomState:
...
         return self.parametrization.random_state
 
+    @property
+    def pareto_front_extractor(self) -> str:
+        return self.__pareto_front_extractor
+
+    @pareto_front_extractor.setter
+    def pareto_front_extractor(self, pareto_front_extractor: str):
+        self.__pareto_front_extractor = pareto_front_extractor
+
```
jrapin, line 160: why do we need this? this seems very weird to me because:
- it adds an API
- it uses setter and getter but without any side effect, so it should only be an instance attribute instead
- it uses double `_` which I never ever use because it usually does not do what you expect when subclassing :)

nevergrad/optimization/optimizerlib.py
```diff
@@ -995,8 +998,11 @@ def __init__(
         num_workers: int = 1,
         base_optimizer: base.OptCls = CMA,
         scale: tp.Optional[float] = None,
+        pareto_front_extractor: tp.Optional[str] = None,
     ) -> None:
         super().__init__(parametrization, budget=budget, num_workers=num_workers)
+        if pareto_front_extractor is not None:
+            self.__pareto_front_extractor = pareto_front_extractor
```
jrapin, line 1005: typically, I am not completely sure this will work because of the `__`

docs/conf.py
teytaud: Done.

nevergrad/benchmark/experiments.py
teytaud: Right, no good reason, done.

nevergrad/benchmark/experiments.py
teytaud: Right, this looks like a good idea shortening everything. Can do after the present PR.

nevergrad/benchmark/experiments.py
teytaud: Yes this was plain stupid -- fixed.

nevergrad/functions/base.py
teytaud: Done for the class method; genre of parametrization coming soon.

teytaud: Hum, I'm not sure for the parametrization... the solutions I see make the code significantly more complicated.

nevergrad/functions/images/imagelosses.py
teytaud: This class should not be registered, it makes no sense.
Your comment applies to other classes - I modify this.

nevergrad/functions/images/imagelosses.py
teytaud: Done.

nevergrad/functions/images/imagelosses.py
teytaud: Done.

nevergrad/optimization/optimizerlib.py
teytaud: __ --> _ done

nevergrad/optimization/base.py
teytaud: ok I simplify thx.

teytaud:
I'm doing soooo dirty things for the OOM issues in the ci...

nevergrad/functions/helpers.py
```diff
@@ -0,0 +1,123 @@
...
+            if provided, selects a subset of the full pareto front with the given maximum size
+        subset: str
+            method for selecting the subset (see optimizer.pareto_front)
+
+        """
+        funcs: tp.List["SpecialEvaluationExperiment"] = []
+        if pareto_subset_methods is None:
+            pareto_subset_methods = ["random", "loss-covering", "EPS", "domain-covering", "hypervolume"]
+            if "home/circleci/repo/venv" in str(pathlib.Path().absolute()):
+                pareto_subset_methods = ["random"]  # faster test in circleci.
```
jrapin, line 107: I don't really like this kind of hack in the code, especially something that depends on something external that is not guaranteed to be true (and which is probably not under Windows) :s also if it's slow on circleci, it's slow locally, which is not good.
one way would be to locally set a environment variable, but I am not a big fan either :s If this is the only way to deal with it I'll do it

nevergrad/functions/helpers.py
```diff
@@ -0,0 +1,123 @@
...
+            iterable of experiment functions, used only as training functions in the crossvalidation and never for test..
+        pareto_size: int
+            if provided, selects a subset of the full pareto front with the given maximum size
+        subset: str
+            method for selecting the subset (see optimizer.pareto_front)
+
+        """
+        funcs: tp.List["SpecialEvaluationExperiment"] = []
+        if pareto_subset_methods is None:
+            pareto_subset_methods = ["random", "loss-covering", "EPS", "domain-covering", "hypervolume"]
```
jrapin, line 105: if we are bound to test all these as defaults, maybe we'll need to be able to return several outputs for one run (one for each subset). I feel doing different experiments altogether may be a waste of resources

nevergrad/functions/images/imagelosses.py
```diff
@@ -3,38 +3,104 @@
...
-    def __init__(self, reference: np.ndarray) -> None:
+class ImageLossWithReference(ImageLoss):
+    def __init__(self, reference: tp.Optional[np.ndarray] = None) -> None:
         if reference is None:
             raise ValueError("A reference is required")
         self.reference = reference
         super().__init__(reference)
+        assert len(self.reference.shape) == 3, self.reference.shape
+        assert self.reference.min() >= 0.0
+        assert self.reference.max() <= 256.0, f"Image max = {self.reference.max()}"
```
jrapin, line 36: isnt it 255?

requirements/bench.txt
```diff
@@ -1,6 +1,8 @@
 requests>=2.21.0
 xlwt>=1.3.0
 xlrd>=1.2.0
+lpips==0.1.3
+opencv-python
```
jrapin, line 5: please add a version

jrapin (changes requested):

nevergrad/benchmark/experiments.py
```diff
@@ -1197,11 +1198,143 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+    # We optimize func_blur or func_brisque and check performance on func_iqa.
+    funcs: tp.List[ExperimentFunction] = [
+        imagesxp.Image(loss=loss) for loss in (
+            imagesxp.imagelosses.Koncept512,
+            imagesxp.imagelosses.Blur,
+            imagesxp.imagelosses.Brisque,
+        )
+    ]
+    # TODO: add the proxy info in the parametrization.
+    if cross_val:
```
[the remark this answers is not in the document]
teytaud: yes

nevergrad/benchmark/xpbase.py
```diff
@@ -166,6 +166,7 @@ def __init__(
         # make sure the random_state of the base function is created, so that spawning copy does not
         # trigger a seed for the base function, but only for the copied function
         self.function.parametrization.random_state  # pylint: disable=pointless-statement
+        self.evaluation_by_best_of_pareto_front = 0
```
jrapin, line 169:
```suggestion
```

nevergrad/functions/images/imagelosses.py
```diff
@@ -3,38 +3,104 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import cv2  # type: ignore
+import imquality.brisque as brisque  # type: ignore
```
jrapin, line 7: instead of adding type ignore here, please append to the list of ignored imports:
https://github.com/facebookresearch/nevergrad/blob/493b526ebe0c210cdc709184b95a35e66b73a3ec/mypy.ini#L4

nevergrad/benchmark/experiments.py
```diff
@@ -1197,11 +1198,143 @@ def images(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+        optims = default_optims
+    funcs:tp.List[ExperimentFunction] = [imagesxp.Image(loss=loss) for loss in [
+                imagesxp.imagelosses.SumAbsoluteDifferences,
+                imagesxp.imagelosses.LpipsAlex,
+                imagesxp.imagelosses.LpipsVgg,
+                imagesxp.imagelosses.SumSquareDifferences,
+                imagesxp.imagelosses.HistogramDifference]]
+    base_values: tp.List[tp.Any] = [func(func.parametrization.sample().value) for func in funcs]
+    mofuncs: tp.List[tp.Any] = helpers.SpecialEvaluationExperiment.create_crossvalidation_experiments(funcs, pareto_size=25) if cross_valid else [
+            fbase.MultiExperiment(funcs, upper_bounds=base_values)]  # type: ignore
```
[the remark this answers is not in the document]
teytaud: done.

nevergrad/functions/images/imagelosses.py
teytaud: we meet values between 255 and 256, hence the change.

jrapin:
please merge master, there are unrelated changes in it

nevergrad/functions/images/imagelosses.py
teytaud: done.

requirements/bench.txt
teytaud: done.

nevergrad/functions/helpers.py
teytaud: True.... our structure is not adapted to this.

Maybe this is something which should be moved to scripts.

I love the idea of a scripts directory for adding examples -- that would help users and disconnect applications from the core.

nevergrad/functions/images/imagelosses.py
teytaud: Done in the PR on PR.

Issue #947, opened 2020-12-12
Minimum Score Not Recorded


pacowong:
## Steps to reproduce
This is an issue discovered in my pull request (https://github.com/facebookresearch/nevergrad/pull/902). To isolate the issue, I have created a smaller test case.

```python
import nevergrad as ng
min_score = None

def fake_training(score: int) -> float:
    global min_score
    if min_score is None or min_score > score:
        min_score = score
    return score

def test(choice_size=1000, budget=300):
    global min_score
    min_score = None
    parametrization = ng.p.Instrumentation(score=ng.p.Choice(list(range(0, choice_size))))
    optimizer = ng.optimizers.OnePlusOne(parametrization=parametrization, budget=budget, num_workers=1)
    recommendation = optimizer.minimize(fake_training)
    print((min_score, recommendation.kwargs['score']))
    return min_score==recommendation.kwargs['score']
```

To run the test:
```python
for i in range(0, 100):
    if not test(1000, 300):
         break
```

## Observed Results
```python
(0, 0)
(1, 1)
(0, 0)
(2, 2)
(10, 10)
(9, 21)
```

## Expected Results
Print 100 tuples and the elements in each tuple should be the same.

## Investigation
The `_update_archive_and_bests` function in `nevergrad/optimization/base.py` should be responsible for incorrect behaviour.
I have checked that the minimum loss has been passed into this function but the archiving process may have some problems:
1. the value of `mvalue` can be `None` in `if mvalue is self.current_bests[name]`
2. Assertion file if `assert mvalue.parameter.loss == candidate.loss` is added before `if mvalue.parameter.loss > candidate.loss`
Not sure why this is the case because `fake_training` function is a deterministic function.

jrapin:
Hi @pacowong 
Thanks for pointing this out. This was indeed an expected behavior which aim was to deal with noisy data. In this case however it was confusing and definitely suboptimal so we have updated the optimizers code. It should be way better indeed.

pacowong:
Thanks.

PR #951, opened 2020-12-14
Fix unexpected recommendation

jrapin:
## Types of changes

#947 
What happens:
- 1+1 converges to very small sigma/steps very quickly
- when reaching machine precision, it considers it is reestimating the same point
- points which have several estimations have lower pessimistic loss, even though they are not as good
- the best pessimistic point is returned, which is not good in this case.

## Checklist


jrapin:
@teytaud not sure how to solve this, maybe use the average as default to return the value ? this seems to be the expected behavior. Returning the pessimistic best as a default even for non-noisy optimization is confusing

teytaud:
> @teytaud not sure how to solve this, maybe use the average as default to return the value ? this seems to be the expected behavior. Returning the pessimistic best as a default even for non-noisy optimization is confusing

In the deterministic setting yes we should return the average.

nevergrad/optimization/base.py
```diff
@@ -488,7 +488,8 @@ def recommend(self) -> p.Parameter:
         """
         recom_data = self._internal_provide_recommendation()  # pylint: disable=assignment-from-none
         if recom_data is None:
-            return self.current_bests["pessimistic"].parameter
+            name = "minimum" if self.parametrization.descriptors.deterministic_function else "pessimistic"
+            return self.current_bests[name].parameter
```
jrapin, line 492: @teytaud this is a drastic change in my opinion, you may want to crosscheck it in the xps

nevergrad/optimization/base.py
```diff
@@ -395,7 +395,7 @@ def _update_archive_and_bests(self, candidate: p.Parameter, loss: tp.FloatLoss)
         # update current best records
         # this may have to be improved if we want to keep more kinds of best losss
 
-        for name in ["optimistic", "pessimistic", "average"]:
+        for name in self.current_bests:
```
teytaud, line 398: ah yes cool as a shortcut :-)

nevergrad/optimization/base.py
teytaud: I think it is safe but yes I'll check xps... complicated though are the absolute indicators are just new.

nevergrad/optimization/test_base.py
```diff
@@ -152,3 +155,25 @@ def test_naming() -> None:
...
+
+    def __init__(self) -> None:
+        self.min_loss = float("inf")
+
+    def __call__(self, score: int) -> float:
+        self.min_loss = min(score, self.min_loss)
+        return score
+
+
+def test_recommendation_correct() -> None:
```
teytaud, line 171: A possible test would be the example in the issue.
That said, I find the fix quite cool and convincing.

nevergrad/optimization/base.py
jrapin: I guess we can wait a bit to merge if it is safer

jrapin: or I provide a fix for the noisy variants

nevergrad/optimization/test_base.py
jrapin: Actually that is the same test, just simplified a bit, but it does exactly the same

teytaud (approved):

PR #970, opened 2020-12-18
Test optimizers on nan and inf

teytaud:
## Types of changes

## Checklist


teytaud:
Tested with --count 10 several times, looks like this test is ok.

nevergrad/optimization/base.py
```diff
@@ -487,7 +489,7 @@ def recommend(self) -> p.Parameter:
             on the function (:code:`objective_function(*candidate.args, **candidate.kwargs)`).
         """
         recom_data = self._internal_provide_recommendation()  # pylint: disable=assignment-from-none
-        if recom_data is None:
+        if recom_data is None or any(np.isnan(recom_data)):
```
teytaud, line 492: if the optimization method returns nan then let us use our own recommendation.

nevergrad/optimization/optimizerlib.py
```diff
@@ -2539,6 +2539,9 @@ def _num_objectives_set_callback(self) -> None:
                 # override at runtime
                 self._optim = DE(self.parametrization, self.budget, self.num_workers)
 
+    def recommend(self) -> p.Parameter:
+        return super().recommend()
```
teytaud, line 2543: We do not want to use an algorithm-specific recommendation, it is often not that good...

teytaud: This modifies NGOpt8 and NGO. I would say it's worth it -- we are already different from any existing publication, so it's good to do the right thing rather than to focus on stability. I would rather focus on stabilization on Dec. 31 and then do a new variant.

nevergrad/optimization/base.py
```diff
@@ -314,6 +314,8 @@ def tell(self, candidate: p.Parameter, loss: tp.Loss) -> None:
         if isinstance(loss, (Real, float)):
             # using "float" along "Real" because mypy does not understand "Real" for now Issue #3186
             loss = float(loss)
+            if not (loss < 5.0e20):
+                loss = 5.0e20  # sys.float_info.max leads to numerical problems so let us do this.
```
jrapin, line 318: What happens beyond?
I really don't like that, this is a hard constant and can lead to unexpected behaviors. It does make a hard assumption on the cost function.
If there is no way out of it, at the very least:
1. add an explanation as a comment
2. use warnings.warn to send a warning that there is something not going well, this must not go unnoticed.

nevergrad/optimization/base.py
jrapin: safer indeed

nevergrad/optimization/optimizerlib.py
```diff
@@ -2539,6 +2539,18 @@ def _num_objectives_set_callback(self) -> None:
                 # override at runtime
                 self._optim = DE(self.parametrization, self.budget, self.num_workers)
 
+    def recommend(self) -> p.Parameter:
```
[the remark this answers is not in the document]
jrapin: please don't copy paste code, otherwise we'll forget to update in both locations if we have to update this function.
here you can use `Optimizer.recommend(self)`
Also, don't you want to integrate this in `NGOptBase` instead? That way all variants can benefit it.
Finally, remove the docstring, the one that matters is the one in the base class

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -137,6 +137,44 @@ def check_optimizer(
...
+        and "TBPSA" not in name  # Dedicated to noisy optimization
+        and "SPSA" != name  # Dedicated to noisy optimization
+        and "BO" not in name  # Our BO does weird things sometimes
+        and "Noisy" not in name  # Dedicated to noisy optimization
+        and "chain" not in name  # Sometimes chaining is not adapted, let us remove all of them
+        and "NGOptBase" != name
+        and "Shiwa" != name
+        and "NGOpt2" != name
+        and "NGO" != name
+    ):
```
jrapin, line 171: please factorize, I'm pretty sure this is going to be worse and worse so better do it early
you could just have one list of names with and then do `if any(x in exclude)`
or two lists if you want to have exact matches for some.

Then again, that would still be hard to maintain, so whenever possible it should actually test with `isinstance(optim, excluded_classes)`
Why is it better?
- because you get all the child classes with it, whatever their name
- because if names change, then you'll have to modify the code here (with strings... we may change name, and leave some strings here for eternity and beyond for no reasons)
I understand it's not always possible though (for some algorithms, you only know it uses BO under the hood because of the name...)

jrapin (changes requested):

nevergrad/optimization/optimizerlib.py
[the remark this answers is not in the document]
teytaud: I do another PR with this change alone first.

teytaud:
... will probably fail as long as other PRs are not merged.

nevergrad/optimization/optimizerlib.py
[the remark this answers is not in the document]
teytaud: Moved to another PR.

nevergrad/optimization/base.py
teytaud: Warning added.
Explanation added.

nevergrad/optimization/test_optimizerlib.py
teytaud: Simplified a lot.
Actually we should remove a lot of old optimizers.
This is impossible for now due to the "mixsimulator" import which crashes if the list of optimization methods is reduced --- this should be fixed soon.

teytaud:
Tested with a lot of replicas --- sounds good.
My guess is that with a better CMA the second chance can be removed.

nevergrad/optimization/base.py
```diff
@@ -314,6 +314,11 @@ def tell(self, candidate: p.Parameter, loss: tp.Loss) -> None:
         if isinstance(loss, (Real, float)):
             # using "float" along "Real" because mypy does not understand "Real" for now Issue #3186
             loss = float(loss)
+            # Non-sense values including NaNs should not be accepted.
+            # We do not use max-float as various later transformations could lead to greater values.
+            if not (loss < 5.0e20):
+                warnings.warn(f"Pathological value {loss} in tell.")
```
jrapin, line 320: More explicit message for the user:
```suggestion
                warnings.warn(f"Clipping very high value {loss} in tell (rescale the cost function?).")
```

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -137,6 +137,51 @@ def check_optimizer(
...
+        if result < 2.0:
+            return
+        assert (  # The "bad" algorithms, most of them originating in CMA's recommendation rule.
+            any(x == name for x in ["WidePSO", "SPSA", "NGOptBase", "Shiwa", "NGO"])
+            or isinstance(optim, optlib.Portfolio)
+            or isinstance(optim, optlib._CMA)
+            or isinstance(optim, optlib._Rescaled)
+            or isinstance(optim, optlib.recaster.SequentialRecastOptimizer)
+            or "NGOpt" in name
+        )  # Second chance!
```
jrapin, line 179: if ever this test ends up needing to get more complicated, it would be better to use a for loop and aviod duplicated code (this is how it's done in `test_optimizers`)

nevergrad/optimization/optimizerlib.py
```diff
@@ -2322,85 +2322,6 @@ class NGO(NGOptBase):  # compatibility
     pass
 
 
-@registry.register
-class NGOpt2(NGOptBase):
```
jrapin, line 2326: why is it disappearing? happy that it does anyway :D

nevergrad/optimization/base.py
teytaud: Thx!

nevergrad/optimization/optimizerlib.py
teytaud: Not used anymore.
The main issue for removing old stuff is the interference with mixsimulator; there are many things to be removed.

jrapin: it's probably preferable to remove it anyway, if it should not be used.
But breaking other people's code is always a bit tricky :s that's why I prefer avoiding creating too many different registered optimizers :D

jrapin (approved):

Issue #966, opened 2020-12-17
Investigate emerging reproducibility issue


jrapin:
#965 fixes tests that have started being flaky recently (Dec 2020) without any related change.
This is probably a dependency update having a wide impact on random states, but not sure.

In 17 December 2020, running this on master was flaky:
pytest nevergrad/benchmark/test_xpbase.py::test_noisy_artificial_function_loss --count=100 --exitfirst
However this was stable before (never had any failure for months/years on this test. Still, using branches dating back to June 2020, I could see the same error. This is why I expect a dependency update has been at play.

From a recent, seemingly stable build in December:
```
absl-py==0.11.0
alabaster==0.7.12
appdirs==1.4.4
astroid==2.4.2
astunparse==1.6.3
attrs==20.3.0
autodocsumm==0.2.2
Babel==2.9.0
backcall==0.2.0
bayesian-optimization==1.2.0
beautifulsoup4==4.9.3
black==20.8b1
bleach==3.2.1
Brotli==1.0.9
cachetools==4.2.0
certifi==2020.12.5
cffi==1.14.4
chardet==3.0.4
click==7.1.2
cloudpickle==1.6.0
cma==3.0.3
colorama==0.4.4
commonmark==0.9.1
coverage==5.3
cryptography==3.3.1
cycler==0.10.0
decorator==4.4.2
docutils==0.16
fcmaes==1.2.2
Flask==1.1.2
Flask-Compress==1.8.0
future==0.18.2
gast==0.3.3
google-auth==1.23.0
google-auth-oauthlib==0.4.2
google-pasta==0.2.0
grpcio==1.34.0
gym==0.17.3
h5py==2.10.0
hiplot==0.1.21
hyperopt==0.2.5
idna==2.10
imageio==2.9.0
imagesize==1.2.0
importlib-metadata==3.1.1
iniconfig==1.1.1
IOHexperimenter==0.2.8.7
ipython==7.19.0
ipython-genutils==0.2.0
isort==5.6.4
itsdangerous==1.1.0
jedi==0.17.2
jeepney==0.6.0
Jinja2==2.11.2
joblib==0.17.0
Keras-Preprocessing==1.1.2
keyring==21.5.0
kiwisolver==1.3.1
koncept==0.2.2
kuti==0.9.6
lazy-object-proxy==1.4.3
Markdown==3.3.3
MarkupSafe==1.1.1
matplotlib==3.3.3
mccabe==0.6.1
[...]

PR #977, opened 2020-12-24
Fix random state for test_noisy_artificial_function_loss

pacowong:
## Types of changes

Related issue: https://github.com/facebookresearch/nevergrad/issues/966
The goal is to resolve the bugs in `random_state` management for `test_noisy_artificial_function_loss` which causes reproducibility issues.
If you want to patch the current system immediately, I can add back `np.random.seed(seed)` to the test case.

## Checklist


teytaud:
Thanks! For the bug in "static" CI, just run "black nevergrad" and this should be solved, this is a formatting issue.

pacowong:
Merry Christmas! It is necessary to decide if `ArtificialVariable` is stochastic, which pseudorandom generator should be used. Because it uses `utils.Transform` which is stochastic. while test cases are set to fixed values. In other words, any changes in using the pseudorandom generator will affect the results. 

Until the decision is made, the following statement will be fixed:
```
if noise_level > 0:
    parametrization.descriptors.deterministic_function = False
```

teytaud:
Actually there is deterministic_function which means that the *function* is deterministic and deterministic which means that the function *and* the transformation are deterministic.

teytaud:
Can you show which bug you are fixing ? This is still unclear to me.

teytaud:
Merry Christmas to all people who have a Christmas or whichever day off you prefer during this time of the year :-)

pacowong:
> 
> 
> Actually there is deterministic_function which means that the _function_ is deterministic and deterministic which means that the function _and_ the transformation are deterministic.

This involves a series of several function calls.
First of all, `ArtificialFunction` implements `evaluation_function` method that uses `_transform` method, in which,
```python
def _transform(self, x: tp.ArrayLike) -> np.ndarray:
        data = self.transform_var.process(x) # <<< self.transform_var is a ArtificialVariable object
        return np.array(data)
```

I found that the `process` method of ```ArtificialVariable``` class is stochastic because 
```python
    def process(  # pylint: disable=unused-argument
        self, data: tp.ArrayLike, deterministic: bool = True
    ) -> np.ndarray:
        if not self._transforms: #This is true as self._transforms is []
            self._initialize() # <<< Trace this method
            ....
```

From the `_initalize()` method, we can see that it creates an object `utils.Transform`
```python
    def _initialize(self) -> None:
        for transform_inds in tools.grouper(indices, n=self.block_dimension):
            self._transforms.append(
                utils.Transform( #<<< This line calls __init__ in utils.Transform
                    transform_inds,
                    translation_factor=self.translation_factor,
                    rotation=self.rotation,
                    random_state=self.random_state,
                )
            )
```

Finally, the class initializer of `Transform` is stochastic because
```python
class Transform:
    """Defines a unique random transformation (index selection, translation, and optionally rotation)
    which can be applied to a point
    """

    def __init__(...):
       ...
       self.translation = ... #<<< This line is stochastic

```

Therefore, my conclusion is `evaluation_function` of `ArtificialFunction` is stochastic. This is the bug I discovered.

nevergrad/optimization/optimizerlib.py
```diff
@@ -110,7 +110,7 @@ def __init__(
         if mutation == "adaptive":
             self._adaptive_mr = 0.5
         if mutation == "coordinatewise_adaptive":
-            self._velocity = np.random.uniform(size=self.dimension) * arity / 4.0
+            self._velocity = self._rng.uniform(size=self.dimension) * arity / 4.0
```
jrapin, line 113: good catch!
I would have expected this to get caught by the CI though :s

jrapin: same for the 3 changes below, nice catches!

nevergrad/benchmark/test_xpbase.py
```diff
@@ -48,21 +48,25 @@ def test_run_packed_artificial_function() -> None:
...
 
 
 def test_noisy_artificial_function_loss() -> None:
     func = ArtificialFunction(name="sphere", block_dimension=5, noise_level=0.3)
     seed = np.random.randint(99999)
     xp = xpbase.Experiment(func, optimizer="OnePlusOne", budget=5, seed=seed)
+    # Because copy() can have different random initialization for the parameters
+    # The function should be copied early.
+    np.random.seed(seed)
+    pfunc = func.copy()
```
jrapin, line 63: can you elaborate? I don't get it, I would have expected to work the same from anywhere.

jrapin (commented):
That's great thanks!
More comments coming, I'm trying to get a better understanding of it little my little.

Also, I am considering adding a test which counts the number of calls to np.random and fails if we add one, with a message explaining the options :D Might be safer on the long run^^

pacowong:
@jrapin Concerning your question about the comment
```python
    # Because copy() can have different random initialization for the parameters
    # The function should be copied early.
    np.random.seed(seed)
    pfunc = func.copy()
    # assert False #Paco: terminate early for debugging
```
Please trace the following code:
1. ExperimentFunction::copy()
   Trace this line, `output = self._internal_copy()`
1. ExperimentFunction::_internal_copy(self: EF) -> EF
   Trace this line, `output: EF = self.__class__(...)`
1. ExperimentFunction::__init__
   Trace this line, `self.parametrization = parametrization`
1. ExperimentFunction::parametrization(self, parametrization: p.Parameter) -> None
```python
@parametrization.setter
def parametrization(self, parametrization: p.Parameter) -> None:
    self._parametrization = parametrization
    self._parametrization.freeze()
    # pylint: disable=pointless-statement
    # print(self._parameterization._random_state) #Paco: you will see None is output twice for each run in the console
    self._parametrization.random_state  # force initialization for synchronization of random state
    # # TODO investigate why this synchronization is needed
```
5. Parameter::random_state(self) -> np.random.RandomState
Somewhere in the program, the random_state of a Parameter variable will be read and this will invoke
```python
    @property
    def random_state(self) -> np.random.RandomState:
        """Random state the instrumentation and the optimizers pull from.
        It can be seeded/replaced.
        """
        if self._random_state is None:
            # use the setter, to make sure the random state is propagated to the variables
            seed = np.random.randint(2 ** 32, dtype=np.uint32)
            self._set_random_state(np.random.RandomState(seed))
        assert self._random_state is not None
        return self._random_state
```
Because `self._random_state is None` satisfies, `seed = np.random.randint(2 ** 32, dtype=np.uint32)` will be invoked.

jrapin:
Sorry I took so long, so many things happening solving this has slipped my mind.
In the flow you are describing, it looks you are missing that 5 is actually called in 4. The last line of `parametrization` setter actually calls the `random_state` getter and the random state is therefore initialized from the very beginning.

In both cases anyway, I don't think changing the the seeding and copy location to a few lines above is supposed to matter.

pacowong:
When I added 
```python
if self.TEMP_DEBUG:
    assert False
```
in the block of `self._random_state is None:` in 5, the assertion is triggered. The seed needs to be reset before `func.copy()`.

In addition, I moved the `func.copy()` forward because it helps isolate the problem. For example, the random states of `func` may change after it is used by the optimizer by the code below:
```python
    xp.run()
    loss_ref = xp.result["loss"]
    # now with copy
    assert xp._optimizer is not None
    reco = xp._optimizer.provide_recommendation()
    assert reco is not None
```

jrapin:
> When I added
> 
> ```python
> if self.TEMP_DEBUG:
>     assert False
> ```
> 
> in the block of `self._random_state is None:` in 5, the assertion is triggered. The seed needs to be reset before `func.copy()`.

Not sure what you mean by this, but I definitely agree that the seed needs to be set before a new function is initialized.

> In addition, I moved the `func.copy()` forward because it helps isolate the problem. For example, the random states of `func` may change after it is used by the optimizer by the code below:

I see your point. My initial aim with the test was to prove that I could reinstantiate a function later on, in an isolated part of the code, and that it would provide the same results. When you move it, you show that the initial function does not change the copy. I'm not sure one claim is stronger than the other, since in the first case you don't really know if there are no intereferences, and in the second, the copy could have modified the initial function (I would expect it shouldn't have... although maybe sometimes it could, so not sure), I guess both should be true^^

jrapin:
Is there anything blocking you from continuing? I see the CI has some changed values. I guess the values can be updated if that's only due to different random generators.

pacowong:
@jrapin The complaints from `mypy` are valid. Please check why `facebook:master` branch passed the tests previously.

jrapin:
@pacowong the typing issue was due to a mypy version update which now seems to parse non-package files. It's now fixed in master so you can merge master within your branch, and if things are still OK I'll merge this, thanks for the fixes ;)

jrapin (approved):

jrapin:
@pacowong thanks for the fix!

pacowong:
@jrapin You're welcome. ;)

PR #990, opened 2021-01-01
Add qualifier for algorithms which need a budget

teytaud:
## Types of changes

## Checklist


nevergrad/benchmark/experiments.py
```diff
@@ -960,26 +960,26 @@ def rocket(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
-    seq = np.arange(0, 1601, 20)
-    for budget in seq:
-        for num_workers in [1, 30]:
-            if num_workers < budget:
-                for algo in optims:
-                    for fu in funcs:
+try:  # Dirty try/except because MixSimulator is not up to date.
+    OptimizeMix()
+    @registry.register
+    def mixsimulator(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
```
jrapin, line 966: don't do this at the base level, there is a Exception you can send from within the experiment instead:
https://github.com/facebookresearch/nevergrad/blob/f3e6085b368fb31a0501d327951560a9afc287ac/nevergrad/benchmark/experiments.py#L1452-L1455
Or better yet, in this case, you should send it from inside the function, example:
https://github.com/facebookresearch/nevergrad/blob/master/nevergrad/functions/images/imagelosses.py#L113-L120

nevergrad/benchmark/experiments.py
```diff
@@ -960,26 +960,26 @@ def rocket(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
+        for budget in seq:
+            for num_workers in [1, 30]:
+                if num_workers < budget:
+                    for algo in optims:
                         xp = Experiment(fu, algo, budget, num_workers=num_workers, seed=next(seedg))
                         if not xp.is_incoherent:
                             yield xp
-
+except AttributeError as e:
+    assert "optimizerlib" in str(e), e
```
jrapin, line 982: what is the message? a small comment would be helpful, right now I don't understand what optimizerlib has to do with it.

nevergrad/optimization/base.py
```diff
@@ -98,6 +98,7 @@ def __init__(
         # you can also replace or reinitialize this random state
         self.num_workers = int(num_workers)
         self.budget = budget
+        self.restricted_budget = False  # False if we can extend the number of ask.
```
jrapin, line 101: isnt that only true for one shots? maybe we could merge this descriptions?
also, I would tend to make it a class attribute instead, so that we can identify which classes work or not without instantiating it, and that would mean adding it to ConfiguredOptimizer as well (and dealing with different cases there for DiscreteBSO

nevergrad/optimization/base.py
```diff
@@ -621,6 +622,12 @@ def minimize(
 
 
 # Adding a comparison-only functionality to an optimizer.
+# Can be used as follows:
+# SQPOptimizer = SQP(50)   # Creates a SQP optimizer in a continuous domain for dimension 50 without constraints.
+# addCompare(SQPOptimizer)
+#
+# Then, SQPOptimizer can be used with SQPOptimize.compare([c1,c2,c3], [bc1, bc2, bc3]) instead of using "tell":
+# it will asssume that candidates c1, c2, c3 are better than candidates bc1, bc2, bc3.
 def addCompare(optimizer: Optimizer) -> None:
```
jrapin, line 631: If you plan on using it further, please:
- rename it with a correct case.
- use a docstring instead of comments

I still really don't like it though, because it is adding a method dynamically, and because this method is defined inline so would probably not be picklable, which will cause issues in the experiments. I suggest we talk through it a bit more, maybe we can come up with something better eventually.

nevergrad/optimization/optimizerlib.py
```diff
@@ -1557,6 +1565,7 @@ def __init__(
     ) -> None:
         super().__init__(parametrization, budget=budget, num_workers=num_workers)
         assert budget is not None
+        self.restricted_budget = True
```
jrapin, line 1568: don't add it when it is already added by the base class.

jrapin: this is also true for `budget_before_choosing` we should not be copied everywhere.

jrapin (changes requested):

nevergrad/optimization/base.py
teytaud: Not all one shot actually.
Some of them, but not all.

There is a complicated issue as you point out with ConfiguredOptimizer, which is the reason for which I did not use a class attribute.

This will remain draft for some time --- other priorities for now.

nevergrad/optimization/optimizerlib.py
teytaud: Done x2 :-)

nevergrad/optimization/base.py
jrapin: > There is a complicated issue as you point out with ConfiguredOptimizer, which is the reason for which I did not use a class attribute.

the reason is that we handle two different objects when dealing with optimizers: classes and Configured optimizer instances, so yes, this requires changes in two places in that case, but I think that is the way you would need to go if I get the aim of what you want to do, or else I'd need more inputs.

Another completely different option would be to make sure that restricted budgets optimizer revert to something else when they are over, but that can be messy as well :s

PR #992, opened 2021-01-05
Specify instrum in photonics/leaderboard

teytaud:
## Types of changes

## Checklist

### DIFF
--- a/nevergrad/functions/photonics/core.py
+++ b/nevergrad/functions/photonics/core.py
@@ -148,7 +148,7 @@ class Photonics(base.ExperimentFunction):
         as_tuple: bool = False,
     ) -> None:
         assert name in ["bragg", "morpho", "chirped"]
-        self.name = name
+        self.name = name + ("as_tuple" if as_tuple else "")
         self._as_tuple = as_tuple
         self._base_func = {
             "morpho": photonics.morpho,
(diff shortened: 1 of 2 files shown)


nevergrad/functions/photonics/core.py
```diff
@@ -148,7 +148,7 @@ def __init__(
         as_tuple: bool = False,
     ) -> None:
         assert name in ["bragg", "morpho", "chirped"]
-        self.name = name
+        self.name = name + ("as_tuple" if as_tuple else "")
```
jrapin, line 151: maybe at least add a coma between the name and "as_tuple"?
as far as I am concerned a better option would be to use `f"{name},{self.parametrization.name}"` although in that case we would need to set a proper name to the parametrization when it's an array, that's not backward compatible (we'd need to update the names in the leaderboard manually)

also, I just figured that in the tuple case, the sigma is not adapted as it is in the array case, that may be a problem.

teytaud: I added a _

I changed the version of mixsimulator, currently master is not ok, this is an issue.

teytaud: Ok even with the current version of mixsimulator this does not work.

jrapin (approved):

Issue #976, opened 2020-12-23
Ask-and-tell interface outputting repeated points


LucasCampos:
I was simulating a optimization which has four workers (GPUs), which runs a function (`fake_training`) asynchronously over 20 days. Each new day, the four workers would get four new points to study. However, after a few days, the optimizer starts outputting the same point for two workers. 

There is no noise in the target-function. I am using nevergrad version 0.4.2_post5.

## Steps to reproduce

  1. Run the code below
  2. Check 1st and 3rd lines of day 9

## Observed Results

On day 9, the results for worker 0 and 2 are the same, as seen below (full log attached)

```
Day: 9
Point for worker 0: {'learning_rate': 0.05191107915156722, 'batch_size': 2, 'learning_decay': 0.8634180716999259, 'learning_decay_steps': 4500} . Simulated loss: 0.2328946307866975
Point for worker 1: {'learning_rate': 0.05191107915156722, 'batch_size': 2, 'learning_decay': 0.8634180716999259, 'learning_decay_steps': 4242} . Simulated loss: 66564.2328946308
Point for worker 2: {'learning_rate': 0.05191107915156722, 'batch_size': 2, 'learning_decay': 0.8634180716999259, 'learning_decay_steps': 4500} . Simulated loss: 0.2328946307866975
Point for worker 3: {'learning_rate': 0.05191107915156722, 'batch_size': 1, 'learning_decay': 0.8634180716999259, 'learning_decay_steps': 4500} . Simulated loss: 1.2328946307866975
```

*Full log:* [log.txt](https://github.com/facebookresearch/nevergrad/files/5733843/log.txt)

## Expected Results

I would have expected that each worker receives a completely independent phase-space point.

## Relevant Code

```python
#! /usr/bin/env python

import nevergrad as ng
import numpy as np

def fake_training(learning_rate: float, batch_size: int, learning_decay: float, learning_decay_steps: int) -> float:
    return 10*(learning_rate - 0.2)**2 + (batch_size - 2)**2 + (learning_decay_steps - 4500)**2 + (learning_decay-0.98)**2


np.random.seed(320)
parametrization = ng.p.Instrumentation(
    learning_rate=ng.p.Log(lower=0.0001, upper=1.0),
    batch_size=ng.p.Scalar(lower=1, upper=2).set_integer_casting(),
    learning_decay = ng.p.Scalar(lower=0.5, upper=1),
[...]

PR #1001, opened 2021-01-07
Avoid two candidates with same value.

teytaud:
## Types of changes

## Checklist


teytaud:
Supposed to fix https://github.com/facebookresearch/nevergrad/issues/976

nevergrad/optimization/base.py
```diff
@@ -439,6 +440,17 @@ def ask(self) -> p.Parameter:
             else:
                 candidate = self._internal_ask_candidate()
                 # only register actual asked points
+
+            # In the deterministic case, possibly mutate.
+            num_determinism_failures = 0
+            while candidate.get_value_hash() in self._forbidden_value:
+                num_determinism_failures += 1
+                if num_determinism_failures >= 10:
```
teytaud, line 448: with this limit we get rid of the risk of infinite / super-long loop.

Still, it would be better to have the auxiliary optimizer (which ensures that we do not go too far) - but Roma was not built in a single day.

nevergrad/functions/pyomo/test_core.py
```diff
@@ -78,7 +78,7 @@ def square2(m: tp.Any) -> float:
     model.constraint1 = pyomo.Constraint(rule=lambda m: m.x >= 2)
 
     func = core.Pyomo(model)
-    optimizer = ng.optimizers.OnePlusOne(parametrization=func.parametrization, budget=100)
+    optimizer = ng.optimizers.OnePlusOne(parametrization=func.parametrization, budget=20)
```
teytaud, line 81: This shows that we are faster than previously.

nevergrad/parametrization/mutation.py
```diff
@@ -146,7 +146,7 @@ def axis(self) -> tp.Optional[tp.Tuple[int, ...]]:
     def _apply_array(self, arrays: tp.Sequence[np.ndarray]) -> np.ndarray:
         assert len(arrays) == 1
         data = arrays[0]
-        axis = tuple(range(data.dim)) if self.axis is None else self.axis
+        axis = tuple(range(len(data))) if self.axis is None else self.axis
```
teytaud, line 149: I don't understand why this was not caught earlier...

jrapin (changes requested):
there are too many side effects, I'll need time to analyze this:
- it's a breaking change (people with noisy function, including using noisy optimizers will by default get deterministic behavior)
- it can explode memory for both long runs and huge parametrization
- it's using the value hash which probably needs a lot of revamp because it is hard to implement (leading to issues when adding new parametrizations), and probably too inefficient.

Using request changes to make sure we do not merge it yet.

PR #1002, opened 2021-01-08
Minor photonics reformatting

jrapin:
## Types of changes

Add typing, black and rename with pep8 compatible variable names

## Checklist

### DIFF
--- a/nevergrad/functions/photonics/photonics.py
+++ b/nevergrad/functions/photonics/photonics.py
@@ -268,8 +268,8 @@ def absorption(
     mu: np.ndarray,
     type_: np.ndarray,
     hauteur: np.ndarray,
-    pol: float,
-    theta: int,
+    pol: int,
+    theta: float,
 ) -> np.ndarray:
     f = mu if not pol else epsilon
     k0 = 2 * np.pi / lam
@@ -347,6 +347,9 @@ def absorption(
 
 
 def cf_photosic_reference(X: np.ndarray) -> float:
+    """vector X is only the thicknesses of each layers, because the materials (so the epislon)
+    are imposed by the function. This is similar in the chirped function.
+    """
     lam_min = 375
     lam_max = 750
     n_lam = 100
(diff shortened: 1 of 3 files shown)


nevergrad/functions/photonics/photonics.py
```diff
@@ -257,42 +255,51 @@ def cascade2(A: np.ndarray, B: np.ndarray) -> np.ndarray:
...
-        f = Epsilon
+def absorption(
+    lam: float,
+    epsilon: np.ndarray,
+    mu: np.ndarray,
+    type_: np.ndarray,
+    hauteur: np.ndarray,
+    pol: float,
+    theta: int,
+) -> np.ndarray:
```
jrapin, line 273: @Ellawin @AnMoreau I've added type hints, and slightly renamed the variables (renaming capitalized letters) since this is more "standard" this way.

nevergrad/functions/photonics/test_core.py
```diff
@@ -171,8 +171,29 @@ def test_photosic_reference() -> None:
...
+            99.9577,
+            82.7487,
+            62.7583,
+            104.1682,
+            139.9002,
+            93.3356,
+            75.6039,
+        ]
+    )
+    cf_test = photonics.cf_photosic_realistic(eps_and_d)
```
jrapin, line 198: @Ellawin @AnMoreau if I get it well, this is structured with the first part being epsilon and the second some kind of thikness? like bragg? Any number of layers work? what are the ones you tend to use?
I'll need to know this so we can explicitely define how this function must be called in the benchmarks

jrapin (commented):
@Ellawin @AnMoreau the changes here are either renaming some variables or reformatting, so no difference. I'll probably merge this before starting deeper changes if need be.

Are there any performance issue currently with the function? If need be I expect this can be rewritten a bit for better efficiency

nevergrad/functions/photonics/photonics.py
Ellawin: pol is an int, and theta a float

nevergrad/functions/photonics/test_core.py
Ellawin: It's exactly that for _cf_photosic_realistic_, _eps_and_d_ is a vector composed in a first part with the epsilon values (the material used in each one of the layers), and in a second part with the thicknesses of each one of the layers, like in Bragg. 
Any number of layers can work. Basically I used between 4 and 50 layers, and the best results are generally obtained when the structure has between 10 and 20 layers. The epsilon values are generally comprised between 1.00 and 9.00.

For _cf_photosic_reference_, the vector _X_ is only the thicknesses of each layers, because the materials (so the epislon) are imposed by the function. This is similar in the chirped function.

jrapin: I added these information in the docstring of the functions (the small texts under the "def" line, which is what appears if you look for help on a function)

jrapin (commented):
I'll be merging this before making actual changes (since this is only unrelated reformatting)

PR #1067, opened 2021-03-03
benchmarking contest track idea

microprediction:
benchmarking contest track idea


teytaud:
Hello!
For the moment I maintain a leaderboard that I rerun entirely periodically. That leaderboard is cool, but I appreciate some advantages of your ELO system.
The key issue is how to "merge" a ELO system like yours and our vast library of benchmarks (we do have an enormous list of benchmarks).
Instead of running everything a given number of time, we might randomly choose (i) a benchmark (ii) an algorithm and update the ELO rank of this algorithm on that benchmark. This is a significant work though.

microprediction:
Yes, I do exactly that. The script is [eloratings.py](https://github.com/microprediction/humpday/blob/main/humpday/comparison/eloratings.py) which you are welcome to.
There's not a lot to Elo per see, though convergence is slow and, well, it doesn't really converge. 

Aside: I'm developing an Elo-like rating system based on [winning](https://github.com/microprediction/winning) which is an algo I just had accepted which will, I think, be much better as it allows for n-way random comparisons. 

I should at the very least use more of your benchmarks.

PR #1089, opened 2021-04-03
Adding non-progressive NoisySplit

teytaud:
## Types of changes

They perform very well in many experiments (incl. important benchmarks added recently).

Good perf on cool new benchmarks.

## Checklist


nevergrad/optimization/optimizerlib.py
```diff
@@ -2473,3 +2473,42 @@ def __init__(
...
         super().__init__(_MSR, locals())
+
+
+# Adding noisy splitters.
+Split_ONOPO3 = ConfSplitOptimizer(
+    num_optims=3, progressive=False, multivariate_optimizer=OptimisticNoisyOnePlusOne
+).set_name("Split_ONOPO3", register=True)
+Split_ONOPO5 = ConfSplitOptimizer(
+    num_optims=5, progressive=False, multivariate_optimizer=OptimisticNoisyOnePlusOne
+).set_name("Split_ONOPO5", register=True)
```
jrapin, line 2484: That's too many,
Please instead create a specific ConfOptimizer (deriving from ConfSplitOptimizer) and just add a couple of examples in experimental variants + make the new configurable optimizer accessible through ng.families (with a proper docstring), that should document it automatically.

teytaud: I'm not in favor of moving to experimentalvariants, because those optimizers are becoming our main tool for large scale RL.
I'm trying to move this to a separate file, but I meet some difficulties:
- I'd like to include that file (say splitters.py) inside optimizerlib.py so that it's in the default lib
- splitters.py needs ConfSplitOptimizer: ok I can move it there.
- ConfSplitOptimizer needs CMA: so I need a dependency to optimizerlib.py ==> circular dependency.

Not sure what to do.

jrapin: What I am saying is that the main tool for large scale optimizer should not be a `Split_ONOPO5` or some other number, but directly something like `NoisySplit(num=5)` which is way more flexible, removes a lot of lines, and is easier to read.

teytaud: Let me simplify then. What do you think ?

teytaud: Our users are sometimes lost when an optimization method is not optimizerlib.xxxx, I try to take that into account.

jrapin: > Our users are sometimes lost when an optimization method is not optimizerlib.xxxx, I try to take that into account.

They will have to get used to it because this will *have to* be the way forward to be able to scale. This will give them flexibility, and robustness, so I keep being definitely against keeping many different optimizers this way.

jrapin: With this it's still:
- not following the naming pattern for optimizers (= class names, without `_`)
- not readable because the name is full of acronyms
- not flexible because not using the configured optimizer interface directly does not let you configure it
So we really need to teach people how to use these options
if the issue is that it's not in the registry as a string, I can probably make it work someway without registration.

jrapin:
See PR on this PR: #1104 
Differences:
- uses a configured optimizer because if we want flexibility and robustness, that's the only way to go (if people want to use a string instead, they can use the registry, after important experimental variants)
- renamed with (hopefully) clearer names.
- see comment, I think the "inf" versions are equivalent

PR #1145, opened 2021-06-14
Adding yahdlbbbob and simplifying stuff

teytaud:
## Types of changes

## Checklist


nevergrad/benchmark/experiments.py
```diff
@@ -613,6 +613,7 @@ def yabbob(
     split: bool = False,
     tiny: bool = False,
     tuning: bool = False,
+    reduced: bool = False,
```
jrapin, line 616: reduction_factor? because otherzise tiny and reduced are similar

teytaud: i've used benchmark_reduction because reduction_factor would suggest an int (which would be a bad idea as we need a number which is coprime with the sizes of loops...).

jrapin: I'd rather have an int and a check that the int is correct then, for now these multiple parameters do not make sense to me :s

teytaud: You mean:
- extracting all lists (one per loop)
- checking coprimality 
Feasible, but the code will not be compact and I'm not sure this will clarify anything.

jrapin: no, just use `reduction_factor: int` and then:
```python
assert reduction_factor in [1, 13, 17]  # needs to be a cofactor
functions = functions[::reduction_factor]
```
I guess that's enough and much simpler and do not add parameters which should be mutually exclusive

teytaud: Yes right this simplies, and this also simplifies yatinybbob and future stuff.

nevergrad/optimization/optimizerlib.py
```diff
@@ -1234,6 +1234,8 @@ def __init__(
         self.max_num_vars = max_num_vars
         self.multivariate_optimizer = multivariate_optimizer
         self.monovariate_optimizer = monovariate_optimizer
+        if self.max_num_vars is not None and self.max_num_vars == 1:  # If we decide to focus on 1, we want 1.
+            self.monovariate_optimizer = multivariate_optimizer
```
jrapin, line 1238: @teytaud I find this really unexpected, if max_num_var is 1 I would definitely expect the monovariate optimizer to be used!

teytaud: All the mess is because CMA does not like dimension one. Maybe I should just switch from monovariate + multivariate to
multivariate + monovariate_for_cases_in_which_the_multivariate_optimizer_does_not_like_dim_1 and switch to the latter only if the multivariate one does not work in dim 1 ?

Ok the name "monovariate_for_cases_in_which_the_multivariate_optimizer_does_not_like_dim_1" is slightly long.

jrapin: maybe don't make it a parameter and just switch to something else for CMA?

teytaud: I just removed that blocking part and we'll see later.

PR #1166, opened 2021-06-23
Export compiler gym data

teytaud:
## Types of changes

## Checklist

### DIFF
--- a/nevergrad/functions/gym/multigym.py
+++ b/nevergrad/functions/gym/multigym.py
@@ -597,7 +597,7 @@ class GymMulti(ExperimentFunction):
                 / 23.0  # This is not compiler_gym but we keep this 23 constant.
             )
         assert self.uses_compiler_gym
-        compilergym_storage = {}
+        compilergym_storage: tp.Dict[tp.Any, tp.Any] = {}
         rewards = [
             np.log(
                 max(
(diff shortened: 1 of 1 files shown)


nevergrad/functions/gym/multigym.py
```diff
@@ -595,17 +597,25 @@ def evaluation_function(self, *recommendations) -> float:
...
+                        compilergym_storage=compilergym_storage,
                     ),
                 )
             )
             for compiler_gym_pb_index in range(23)
         ]
+        print(
+            f"<<<compilergym:{[locals()[k] for k in sorted(locals().keys())]}:{[(k, compilergym_storage[k]) for k in sorted(compilergym_storage.keys())]}>>>",
+            file=sys.stderr,
+        )
```
jrapin, line 618: @bottler should we keep it this way for now and move it to use probably logger.info directly?

jrapin: @teytaud what is the aim here? I feel this is overly complicated

teytaud: We need info on the final points.
But not only the candidate: the list of actions that are obtained from the candidate on each test case.
I'm not sure how much this will prove useful for other problems. What is sure is that these details are very useful for the current compilergym project. We might wait a bit before deciding if we want to merge something like that.

jrapin: I'm fine with some prints, not sure it has to go to stderr though, probably better to just use stdout.
and to make it less hacky, it's better to not use locals... can you print an example here? is everything used?

teytaud:
To be implemented better.

PR #1177, opened 2021-07-05
Adding Olympus benchmark

ncarraz:
## Types of changes

## Checklist

### DIFF
--- a/nevergrad/benchmark/experiments.py
+++ b/nevergrad/benchmark/experiments.py
@@ -31,6 +31,7 @@ from nevergrad.functions.games import game
 from nevergrad.functions.causaldiscovery import CausalDiscovery
 from nevergrad.functions import iohprofiler
 from nevergrad.functions import helpers
+from nevergrad.functions import Olympus
 from .xpbase import Experiment as Experiment
 from .xpbase import create_seed_generator
 from .xpbase import registry as registry  # noqa
@@ -1231,6 +1232,26 @@ def neuro_control_problem(seed: tp.Optional[int] = None) -> tp.Iterator[Experime
                 xp = Experiment(fu, algo, budget, num_workers=1, seed=next(seedg))
                 if not xp.is_incoherent:
                     yield xp
+
+
+@registry.register
+def olympus(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
+    """Olympus surfaces """
+    funcs = [
+        Olympus(kind) 
+        for kind in ["AckleyPath", "Branin", "Dejong", "HyperEllipsoid","Levy", "Michalewicz", 
+            "Rastrigin", "Rosenbrock", "Schwefel", "StyblinskiTang", "Zakharov"]
+    ]
+    seedg = create_seed_generator(seed)
+    optims = get_optimizers("basics", "noisy", seed=next(seedg))
+    for budget in [25, 50, 100, 200, 400, 800, 1600, 3200, 6400, 12800, 25600]:
+        for num_workers in [1]:  # , 10, 100]:
+            if num_workers < budget:
+                for algo in optims:
+                    for fu in funcs:
+                        xp = Experiment(fu, algo, budget, num_workers=num_workers, seed=next(seedg))
+                        if not xp.is_incoherent:
+                            yield xp
 
 
 @registry.register
--- a/nevergrad/benchmark/experiments.py
+++ b/nevergrad/benchmark/experiments.py
@@ -31,7 +31,7 @@ from nevergrad.functions.games import game
 from nevergrad.functions.causaldiscovery import CausalDiscovery
 from nevergrad.functions import iohprofiler
 from nevergrad.functions import helpers
-from nevergrad.functions import Olympus
+from nevergrad.functions.olympus import Olympus
 from .xpbase import Experiment as Experiment
 from .xpbase import create_seed_generator
 from .xpbase import registry as registry  # noqa
@@ -1237,11 +1237,12 @@ def neuro_control_problem(seed: tp.Optional[int] = None) -> tp.Iterator[Experime
 @registry.register
 def olympus(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
     """Olympus surfaces """
-    funcs = [
-        Olympus(kind) 
-        for kind in ["AckleyPath", "Branin", "Dejong", "HyperEllipsoid","Levy", "Michalewicz", 
-            "Rastrigin", "Rosenbrock", "Schwefel", "StyblinskiTang", "Zakharov"]
-    ]
+    funcs = []
+    for kind in ["AckleyPath", "Dejong", "HyperEllipsoid","Levy", "Michalewicz", 
+                "Rastrigin", "Rosenbrock", "Schwefel", "StyblinskiTang", "Zakharov"]:
+        for k in range(2, 6):
+            funcs.append(Olympus(kind, 10 ** k)) 
+
     seedg = create_seed_generator(seed)
     optims = get_optimizers("basics", "noisy", seed=next(seedg))
     for budget in [25, 50, 100, 200, 400, 800, 1600, 3200, 6400, 12800, 25600]:
--- a/nevergrad/functions/olympus/core.py
+++ b/nevergrad/functions/olympus/core.py
@@ -1,0 +1,41 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Based on https://github.com/aspuru-guzik-group/olympus
+
+import matplotlib.pyplot as plt
+import numpy as np
+from nevergrad.parametrization import parameter as p
+from ..base import ExperimentFunction
+import olympus
+
+
+class Olympus(ExperimentFunction):
+    def __init__(self, kind : str, dimension: int = 10) -> None:
+        self.kind = kind
+        self.param_dim = dimension
+        super().__init__(self._simulate_traditional_surface, p.Array(shape=(dimension,)))
+    
+    def _simulate_traditional_surface(self, x: np.ndarray) -> float:
+        assert kind in ["AckleyPath", "Branin", "Dejong", "HyperEllipsoid",
+                        "Levy", "Michalewicz", "Rastrigin", "Rosenbrock", 
+                        "Schwefel", "StyblinskiTang", "Zakharov"]
+
+        traditional_surfaces = {"Michalewicz":olympus.surfaces.Michalewicz,
+        "AckleyPath":olympus.surfaces.AckleyPath,
+        "Branin":olympus.surfaces.Branin,
+        "Dejong":olympus.surfaces.Dejong,
+        "HyperEllipsoid":olympus.surfaces.HyperEllipsoid,
+        "Levy":olympus.surfaces.Levy,
+        "Michalewicz":olympus.surfaces.Michalewicz,
+        "Rastrigin":olympus.surfaces.Rastrigin,
+        "Rosenbrock":olympus.surfaces.Rosenbrock,
+        "Schwefel":olympus.surfaces.Schwefel,
+        "StyblinskiTang":olympus.surfaces.StyblinskiTang,
+        "Zakharov":olympus.surfaces.Zakharov,
+        }
+        
+        surface = traditional_surfaces[self.kind](param_dim=self.param_dim)
+        return surface.run(x)[0][0]
--- a/nevergrad/functions/olympus/core.py
+++ b/nevergrad/functions/olympus/core.py
@@ -19,22 +19,22 @@ class Olympus(ExperimentFunction):
         super().__init__(self._simulate_traditional_surface, p.Array(shape=(dimension,)))
     
     def _simulate_traditional_surface(self, x: np.ndarray) -> float:
-        assert kind in ["AckleyPath", "Branin", "Dejong", "HyperEllipsoid",
+        assert self.kind in ["AckleyPath", "Dejong", "HyperEllipsoid",
                         "Levy", "Michalewicz", "Rastrigin", "Rosenbrock", 
                         "Schwefel", "StyblinskiTang", "Zakharov"]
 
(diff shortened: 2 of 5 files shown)


ncarraz:
Benchmark on the traditional surfaces only. I did not include Branin because there was an internal Olympus error when trying to instatiate it. No noise yet but to be included later.

nevergrad/functions/olympus/core.py
```diff
@@ -0,0 +1,41 @@
...
+import numpy as np
+from nevergrad.parametrization import parameter as p
+from ..base import ExperimentFunction
+import olympus
+
+
+class Olympus(ExperimentFunction):
+    def __init__(self, kind : str, dimension: int = 10) -> None:
+        self.kind = kind
+        self.param_dim = dimension
```
teytaud, line 18: This is correct but does not inform the optimization algorithm that the objective function is noisy, I would do something like

parametrization = p.Array(shape=(dimension,))
parametrization.function.deterministic = False

(@jrapin , big boss of parametrization stuff, might know more than i do :-) )

nevergrad/functions/olympus/core.py
```diff
@@ -0,0 +1,41 @@
...
+        super().__init__(self._simulate_traditional_surface, p.Array(shape=(dimension,)))
+    
+    def _simulate_traditional_surface(self, x: np.ndarray) -> float:
+        assert self.kind in ["AckleyPath", "Dejong", "HyperEllipsoid",
+                        "Levy", "Michalewicz", "Rastrigin", "Rosenbrock", 
+                        "Schwefel", "StyblinskiTang", "Zakharov"]
+
```
teytaud, line 25: Sounds correct to me, but for making assessment faster we can average noise by specifying that for a posteriori evaluation we might average 42 runs as follows (this does not impact the run, just the evaluation of the final solution at the end...):

def evaluation_function(self, *recommendations) -> float:
        """Averages multiple evaluations if necessary."""
        x = recommendations[0].value
        losses = [surface.run(x)[0][0] for _ in range(42)]
        return sum(losses) / len(losses)

teytaud:
> Benchmark on the traditional surfaces only. I did not include Branin because there was an internal Olympus error when trying to instatiate it. No noise yet but to be included later.

ah sorry for having commented without seeing that you had already mentioned that noise would come up later :-)

nevergrad/functions/olympus/core.py
teytaud: Even better (if possible) would be to have "def evaluation_function" return the value without noise (while still using the noisy one in super.__init__(....., ...) ).

ncarraz:
test_olympus in test_core.py does not test everything yet. I managed to use the surface without the noise in evaluation_function but you should check.

ncarraz:
I think it passed the tests but I run out of credits in circkeci

nevergrad/benchmark/experiments.py
```diff
@@ -1233,6 +1234,39 @@ def neuro_control_problem(seed: tp.Optional[int] = None) -> tp.Iterator[Experime
...
+        "HyperEllipsoid",
+        "Levy",
+        "Michalewicz",
+        "Rastrigin",
+        "Rosenbrock",
+        "Schwefel",
+        "StyblinskiTang",
+        "Zakharov",
+    ]:
+        for k in range(2, 6):
```
teytaud, line 1253: Maybe 5 is enough because dim = 1 000 000 in a noisy context will be a lot.

ncarraz: okay I will fix that. what about the noise scale ?

nevergrad/functions/olympussurfaces/core.py
```diff
@@ -0,0 +1,82 @@
...
+            "Schwefel": surfaces.Schwefel,
+            "StyblinskiTang": surfaces.StyblinskiTang,
+            "Zakharov": surfaces.Zakharov,
+            "DiscreteAckley": surfaces.DiscreteAckley,
+            "DiscreteDoubleWell": surfaces.DiscreteDoubleWell,
+            "DiscreteMichalewicz": surfaces.DiscreteMichalewicz,
+            "LinearFunnel": surfaces.LinearFunnel,
+            "NarrowFunnel": surfaces.NarrowFunnel,
+            "GaussianMixture": surfaces.GaussianMixture,
+        }
```
jrapin, line 70: Please move this to a global variable, then you can create a function `get_kinds()` which returns `list(TRADITIONAL_SURFACES)`
This will remove a lot of duplicated items, and therefore will be simpler to maintain

nevergrad/functions/olympussurfaces/core.py
```diff
@@ -0,0 +1,82 @@
...
+            surface = traditional_surfaces[self.kind](param_dim=self.param_dim, noise=noise)
+        else:
+            surface = traditional_surfaces[self.kind](param_dim=self.param_dim)
+        return surface.run(x)[0][0]
+
+    def evaluation_function(self, *recommendations) -> float:
+        """Averages multiple evaluations if necessary."""
+        x = recommendations[0].value
+        losses = [self.surface_without_noise(x) for _ in range(42)]
+        return sum(losses) / len(losses)
```
jrapin, line 82: would that work to use the non-noisy case instead?
```python
traditional_surfaces[self.kind](param_dim=self.param_dim)
return surface.run(x)[0][0]
```

ncarraz: Using the non-noisy surface in the evaluation function was proposed by Olivier.

teytaud: Yes! But now we can remove the loop.

ncarraz: ok got it

nevergrad/functions/olympussurfaces/core.py
```diff
@@ -0,0 +1,121 @@
...
+
+import nevergrad as ng
+
+try:
+    from olympus import surfaces  # type: ignore
+    from olympus import noises  # type: ignore
+    from olympus import Emulator  # type: ignore
+    from olympus.datasets import Dataset  # type: ignore
+except ImportError as e:
+    raise ng.errors.UnsupportedExperiment("Please install olympus for Olympus experiments") from e
```
jrapin, line 21: This needs to be done inline in the function, otherwise it will break all other experiments.

Also, instead of all the type ignore, add `olympus` to mypy.ini:
https://github.com/facebookresearch/nevergrad/blob/603221a535dfaebfd7d76b56732cbd9b716292eb/mypy.ini#L3

nevergrad/functions/olympussurfaces/core.py
```diff
@@ -0,0 +1,121 @@
...
+        "Schwefel": surfaces.Schwefel,
+        "StyblinskiTang": surfaces.StyblinskiTang,
+        "Zakharov": surfaces.Zakharov,
+        "DiscreteAckley": surfaces.DiscreteAckley,
+        "DiscreteDoubleWell": surfaces.DiscreteDoubleWell,
+        "DiscreteMichalewicz": surfaces.DiscreteMichalewicz,
+        "LinearFunnel": surfaces.LinearFunnel,
+        "NarrowFunnel": surfaces.NarrowFunnel,
+        "GaussianMixture": surfaces.GaussianMixture,
+    }
```
jrapin, line 44: can this be created with just: `{name: getattr(surfaces, name) for name in dir(surfaces)}`?
There may be a few additional gotchas (getting only if it's the correct class, skipping some of the names which are not supported) but this way the code will be much more maintainable.

Or otherwise just create `names = ["Michalewicz", ...]` if you prefer to make sure we do not use other surfaces, then, the corresponding object is just `getattr(surfaces, name)`, so we can remove a lot of duplications

nevergrad/functions/olympussurfaces/core.py
```diff
@@ -0,0 +1,121 @@
...
+        "alkox",
+        "benzylation",
+        "photo_pce10",
+        "hplc",
+        "colors_n9",
+    ]
+
+    @classmethod
+    def get_datasets(cls):
+        return cls.datasets
```
jrapin, line 97: Simpler to make it explicitely an unmutable (= a tuple instead of a list) class attribute, and use it directly afterwards

```suggestion
    DATASETS = (
        "suzuki",
        "fullerenes",
        "colors_bob",
        "photo_wf3",
        "snar",
        "alkox",
        "benzylation",
        "photo_pce10",
        "hplc",
        "colors_n9",
    )
```

jrapin (commented):
Hi,
See a few comments here
Please merge the latest master within your branch, this should simplify your PR a bit, and solve the current pytest issues which are unrelated to your PR

nevergrad/benchmark/test_experiments.py
```diff
@@ -36,8 +36,8 @@ def test_experiments_registry(name: str, maker: tp.Callable[[], tp.Iterator[expe
     if "mixsimulator" in name and os.environ.get("CIRCLECI", False):
         raise SkipTest("Sigkill in CircleCI")
 
-    # Our IQAs and our ScikitLearn are not well guaranteed on Windows.
-    if all(x in name for x in ["image", "quality"]) and platform.system() == "Windows":
+    # Our IQAs and our ScikitLearn and Olympus are not well guaranteed on Windows.
+    if all(x in name for x in ["image", "quality", "olympus"]) and platform.system() == "Windows":
```
jrapin, line 40: this should not be necessary anymore now that it raises UnsupportedExperiment, so please try and revert the change

nevergrad/functions/olympussurfaces/core.py
```diff
@@ -0,0 +1,122 @@
...
+        return p.Array(shape=(dimension,), lower=bounds[0], upper=bounds[1])
+
+    def _simulate_emulator(self, x: np.ndarray) -> float:
+        try:
+            from olympus import Emulator  # pylint: disable=import-outside-toplevel
+        except ImportError as e:
+            raise ng.errors.UnsupportedExperiment("Please install olympus for Olympus experiments") from e
+
+        assert self.dataset_kind in OlympusEmulator.DATASETS
+        assert self.model_kind in ["BayesNeuralNet", "NeuralNet"]
```
jrapin, line 116: test this in the `__init__` instead, no need to have it tested at each call of the function

nevergrad/functions/olympussurfaces/core.py
```diff
@@ -0,0 +1,122 @@
...
+
+        emulator = Emulator(dataset=self.dataset_kind, model=self.model_kind)
+        if emulator.get_goal() == "maximize":
+            return -emulator.run(x)[0][0]
+        else:
+            return emulator.run(x)[0][0]
```
jrapin, line 122: I tend to factorize all repeated code asap, since fewer repetition usually means less maintenance :)

```suggestion
        return emulator.run(x)[0][0] * (-1 if emulator.get_goal() == "maximize" else 1)
```

requirements/bench.txt
```diff
@@ -26,3 +26,5 @@ tensorflow>=2.5.0
 image-quality>=1.2.7
 keras>=2.4.3
 compiler_gym>=0.1.8 ; sys_platform == "linux"
+olymp==0.0.1b0 ; sys_platform == "linux"
+silence_tensorflow
```
jrapin, line 30: there seems to be a few more missing imports (cf CI)

jrapin (commented):
looks like we're nearly there ;)

requirements/bench.txt
```diff
@@ -29,3 +29,8 @@ keras>=2.4.3
 compiler_gym>=0.1.8 ; sys_platform == "linux"
 pymoo>=0.4.2.2
 bayes-optim>=0.2.5.5
+olymp==0.0.1b0 ; sys_platform == "linux"
+silence_tensorflow
+tensorflow_probability
+sqlalchemy
+seaborn
```
jrapin, line 36: are those really needed? looks like database and plotting tools so that feels unrelated

jrapin (commented):
I have no idea what is the problem with the test failing in the CI, I have tried updating all the dependency to check if it was due to a dependency update but all worked (#1220), so it must be due to one of the additional imports, no idea which though :s and as long as the CI does not pass we won't be able to merge :s

requirements/bench.txt
ncarraz: Those are used internaly by Olympus. I could try removing them and see if there are no issues.

ncarraz: It was Koncept512 that failed in pytests. Is it possible that it is related ?

ncarraz: It works locally though.

jrapin: > It was Koncept512 that failed in pytests. Is it possible that it is related ?

Initially I would have expected it to be unrelated, but after a clean install of master on the CI, this issue does not appear. So what triggers the problem is necessaritly somewhere here, I would expect one of the packages, and because koncept512 is in tensorflow, probably silence_tensorflow or tensorflow_probability. tensorflow version and keras version are the same in master though so it's not "simply" a problem version :s
Is silence_tensorflow needed? Might be good to try without it I guess :s

nevergrad/benchmark/test_experiments.py
```diff
@@ -55,7 +55,7 @@ def test_experiments_registry(name: str, maker: tp.Callable[[], tp.Iterator[expe
         maker,
         ("mltuning" in name or "anm" in name),
         skip_seed=(name in ["rocket", "images_using_gan"])
-        or any(x in name for x in ["tuning", "image_", "compiler", "anm"]),
+        or any(x in name for x in ["tuning", "image_", "compiler", "anm", "olympus"]),
```
jrapin, line 58: is olympus not seedable in some way? If you had to add this here it means the test cases would not be reproducible :(
I see the shift is using the parametrization random state which is seeded so should be good, so that must come from olympus

ncarraz: Unfortunately, I have not found any ways to seed the emulators but the surfaces are seedable.

requirements/bench.txt
ncarraz: I will have to check all that.

nevergrad/benchmark/test_experiments.py
jrapin: too bad :(

PR #1193, opened 2021-07-16
Adding OptimizationLogger and replacing print for log in nsga2.py

theoajc:
## Types of changes

To be able to log callbacks as well as printing them. Replaced print with log in nsga2.py because more appropriate.

Test added to ensure class works i.e. functions terminate and output checked through running pytest with logging on and looking at output.

## Checklist

### DIFF
--- a/nevergrad/optimization/test_callbacks.py
+++ b/nevergrad/optimization/test_callbacks.py
@@ -116,19 +116,31 @@ def test_early_stopping() -> None:
     assert optimizer.recommend().loss < 12  # type: ignore
 
 
-def test_optimization_logger(tmp_path: Path) -> None:
-    instrum = ng.p.Instrumentation(
-        None, 2.0, blublu="blublu", array=ng.p.Array(shape=(3, 2)), multiobjective=True
-    )
-    logging.basicConfig(level=os.environ.get("LOGLEVEL", "INFO"))
-    logger = logging.getLogger(__name__)
-    optimizer = optimizerlib.OnePlusOne(parametrization=instrum, budget=100)
-    early_stopping = ng.callbacks.EarlyStopping(lambda opt: opt.num_ask > 3)
-    optimizer.register_callback(
-        "tell",
-        callbacks.OptimizationLogger(
-            logger=logger, log_level=logging.INFO, log_interval_tells=10, log_interval_seconds=0.1
-        ),
-    )
-    optimizer.register_callback("ask", early_stopping)
-    optimizer.minimize(_func, verbosity=2)
+def test_optimization_logger(caplog) -> None:
+    with caplog.at_level(logging.INFO):
+        instrum = ng.p.Instrumentation(
+            None, 2.0, blublu="blublu", array=ng.p.Array(shape=(3, 2)), multiobjective=True
+        )
+        logging.basicConfig(level=os.environ.get("LOGLEVEL", "INFO"))
+        logger = logging.getLogger(__name__)
+        optimizer = optimizerlib.OnePlusOne(parametrization=instrum, budget=3)
+        optimizer.register_callback(
+            "tell",
+            callbacks.OptimizationLogger(
+                logger=logger, log_level=logging.INFO, log_interval_tells=10, log_interval_seconds=0.1
+            ),
+        )
+        optimizer.minimize(_func, verbosity=2)
+        assert (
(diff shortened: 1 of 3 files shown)


nevergrad/optimization/test_callbacks.py
```diff
@@ -112,3 +114,21 @@ def test_early_stopping() -> None:
...
+
+
+def test_optimization_logger(tmp_path: Path) -> None:
+    instrum = ng.p.Instrumentation(
+        None, 2.0, blublu="blublu", array=ng.p.Array(shape=(3, 2)), multiobjective=True
+    )
+    logging.basicConfig(level=os.environ.get("LOGLEVEL", "INFO"))
+    logger = logging.getLogger(__name__)
+    optimizer = optimizerlib.OnePlusOne(parametrization=instrum, budget=100)
+    early_stopping = ng.callbacks.EarlyStopping(lambda opt: opt.num_ask > 3)
```
jrapin, line 126: I think you could just remove the early stopping and just use budget=3, this way we avoid risking intereferences with EarlyStopping in this test

nevergrad/optimization/test_callbacks.py
```diff
@@ -112,3 +114,21 @@ def test_early_stopping() -> None:
...
+    optimizer.register_callback(
+        "tell",
+        callbacks.OptimizationLogger(
+            logger=logger, log_level=logging.INFO, log_interval_tells=10, log_interval_seconds=0.1
+        ),
+    )
+    optimizer.register_callback("ask", early_stopping)
+    optimizer.minimize(_func, verbosity=2)
```
jrapin, line 134: can we capture the logs here and check that it actually works?

theoajc: Now checks that beginning of expected log message for 1st, 2nd and 3rd logs are correct. (doesn't check for correct results because they change every time.)

jrapin:
Please merge master within your branch, I've removed a couple of buggy tests for now, so the CI can work correctly

nevergrad/optimization/test_callbacks.py
```diff
@@ -112,3 +114,33 @@ def test_early_stopping() -> None:
     # below functions are inlcuded in the docstring
     assert optimizer.current_bests["minimum"].mean < 12
     assert optimizer.recommend().loss < 12  # type: ignore
+
+
+def test_optimization_logger(caplog) -> None:
+    with caplog.at_level(logging.INFO):
```
jrapin, line 120: can you ut this just before the call to minimize? Let's keep context as small as possible (this way it's clearer what it captures)
I'm actually wondering if the context could be around *only* the minimize method, I think you can check the text afterwards even if you are no more in the context, so that would be even better.

nevergrad/optimization/test_callbacks.py
```diff
@@ -112,3 +114,33 @@ def test_early_stopping() -> None:
...
+        )
+        assert (
+            "After 1, recommendation is Instrumentation(Tuple(None,2.0),Dict(array=Array{(3,2)},blublu=blublu,multiobjective=True))"
+            in caplog.text
+        )
+        assert (
+            "After 2, recommendation is Instrumentation(Tuple(None,2.0),Dict(array=Array{(3,2)},blublu=blublu,multiobjective=True))"
+            in caplog.text
+        )
+        assert "'multiobjective': True" in caplog.text
```
jrapin, line 146: I think one would be enough to make sure it works, no need to check in too much detail. My reason for this is that if ever we change something in the display of the parameters, I'd rather avoid needing to rewrite each of the lines :D long term maintenance is hell :D

nevergrad/optimization/test_callbacks.py
jrapin: you could check the full log if you seed the parametrization. But if you are confident that all the useful information is properly written then it's fine without.

nevergrad/optimization/test_callbacks.py
theoajc: I've implemented these recommendations and you are correct. The context is now only around the minimize method with the assertions outside of the context.

nevergrad/optimization/test_callbacks.py
theoajc: So just to be clear, I'll keep the line looking as is but remove the repetitions?

jrapin: yes

jrapin (approved):
You can "Squash & Merge" when you want ;) (just don't just "Merge" but I think it's disabled on this repo anyway)

PR #1228, opened 2021-09-06
Enable some parallelism in pymoo

theoajc:
## Types of changes

Previously, there was no way to do parallel evaluations using any of the Recast optimizers. However, Pymoo does support parallel evaluations. This new class "BatchRecastOptimizer" that inherits from RecastOptimizer where instead points to evaluate are provided in batches to be asked and told on. This opens the optimizer up to parallelism (within a given batch).

N.B. These implementations are not necessarily robust. You have to complete
a batch before you start a new one so parallelism is only possible
within batches i.e. if a batch size is 100 and you have done 100 asks,
you must do 100 tells before you ask again but you could do those 100
asks and tells in parallel.

New test added to do 100 asks and then 100 tells multiple times and assert that at the end there are no unevaluated points and of course we don't get an error on the way.

## Checklist


nevergrad/optimization/recaster.py
```diff
@@ -206,3 +207,77 @@ class SequentialRecastOptimizer(RecastOptimizer):
...
+            data = self._rng.normal(0, 1, self.dimension)
+            return self.parametrization.spawn_child().set_standardized_data(data)
+        # if no more points left in batch, wait for new batch from fake callable
+        if not self._current_batch:
+            points = self._messaging_thread.messages_ask.get()
+            self.batch_size = len(points)
+            self._current_batch = [self.parametrization.spawn_child(point) for point in points]
+            self._batch_losses = [None] * len(points)  # type: ignore
+            # map each point to an index in preparation to build loss array in tell
+            self.indices = {candidate: i for i, candidate in enumerate(self._current_batch)}
```
jrapin, line 264: better use `candidate.uid`, I'm actually not sure why Python does not complain that candidate is not hashable here :s

nevergrad/optimization/recaster.py
```diff
@@ -206,3 +207,77 @@ class SequentialRecastOptimizer(RecastOptimizer):
...
+                base.errors.FinishedUnderlyingOptimizerWarning,
+            )
+            self._check_error()
+            data = self._rng.normal(0, 1, self.dimension)
+            return self.parametrization.spawn_child().set_standardized_data(data)
+        # if no more points left in batch, wait for new batch from fake callable
+        if not self._current_batch:
+            points = self._messaging_thread.messages_ask.get()
+            self.batch_size = len(points)
+            self._current_batch = [self.parametrization.spawn_child(point) for point in points]
```
jrapin, line 261:
```suggestion
            self._current_batch = [self.parametrization.spawn_child().set_standardized_data(point) for point in points]
```
The value parameter in spawn_child has a different meaning (aka the actual value for the function, which could be a string, a dict whatever, while optimizers work in a standardized space)

nevergrad/optimization/recaster.py
```diff
@@ -206,3 +207,77 @@ class SequentialRecastOptimizer(RecastOptimizer):
...
+            self._batch_losses = [None] * len(points)  # type: ignore
+            # map each point to an index in preparation to build loss array in tell
+            self.indices = {candidate: i for i, candidate in enumerate(self._current_batch)}
+        candidate = self._current_batch.pop()
```
jrapin, line 265: what happens if we pull too many candidates? An explicit error explaining that the batch size is too small could be helpful. Note that some algorithms work in steady mode, i.e if you just did a tell, you can always do a new ask. That won't be the case here so some users could get a bit lost and may need to be explained what happened ;)

nevergrad/optimization/recaster.py
theoajc: Of course! I'm not sure why it didn't seem to mind but will do this.

nevergrad/optimization/recaster.py
theoajc: If you ask before the previous batch has been fully told on you end up in a deadlock potentially (if more than one worker asks) since you have multiple workers blocking to get from a queue that will only have one member. I have added a descriptive runtime error and a new function `can_ask()` that handles the logic of knowing whether an ask can be performed.

theoajc: `
if not self.can_ask(): raise RuntimeError( "You can't get a new batch until the old one has been fully told on. See docstring for more info." )
`
Instead of the batch being too small I have added an error stating that a premature ask has occurred.

nevergrad/optimization/recastlib.py
```diff
@@ -213,33 +212,141 @@ def _post_loss(self, candidate: p.Parameter, loss: float) -> tp.Loss:
...
+            parametrization=self.parametrization,
+            budget=self.budget,
+            num_workers=self.num_workers,
+            algorithm=self.algorithm,
+        )
+        # set num_objectives in sub-instance for Pymoo to use in problem definition
+        if self.num_objectives > 0:
+            subinstance.num_objectives = self.num_objectives
+        else:
+            raise RuntimeError("num_objectives should have been set.")
```
jrapin, line 287: No need of if/else if one of the branch is just an error, I feel it's simpler to read this way, but not really important:
```suggestion
        if self.num_objectives <= 0:
            raise RuntimeError("num_objectives should have been set.")
        subinstance.num_objectives = self.num_objectives
```

jrapin (approved):

nevergrad/optimization/recastlib.py
theoajc: I agree!

PR #1235, opened 2021-09-10
ttp add

yvan-vin:
## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

This change adds another problem that can be used for benchmarking, this problem is the traveling thief problem. This problem is useful as it is the combination of two well understood problems, the travelling salesman and the knapsack problem. 
Also useful as it uses two different datatypes that are optimised, a sequence and bitvector.

The TTP function/class itself has been tested and cross-validated with other implementations to produce consistent results.
Integration has been tested using experiments.

## Checklist

- [ x ] The documentation is up-to-date with the changes I made.
- [ x ] I have read the [**CONTRIBUTING**](https://facebookresearch.github.io/nevergrad/contributing.html) document and completed the CLA (see [**CLA**](https://facebookresearch.github.io/nevergrad/contributing.html#contributor-license-agreement-cla)).
- [ x ] All tests passed, and additional code has been covered with new tests.


nevergrad/functions/ttp/core.py
```diff
@@ -0,0 +1,224 @@
...
+        p2.set_bounds(0, 1)
+        p2.set_name("packingPlan")
+
+        instru = ng.p.Instrumentation(tspTour=p1, packingPlan=p2).set_name("")
+
+        super().__init__(self.evaluate, instru)
+
+    def evaluate(self, tspTour: np.ndarray, packingPlan: np.ndarray) -> float:
+
+        np.reshape(tspTour, (1, -1))
```
jrapin, line 148: instead of reshaping, you should make the parametrization have the right shape from the start, probably with
`init1 = np.zeros(1, self.numberOfNodes - 1)`

nevergrad/functions/ttp/core.py
```diff
@@ -0,0 +1,224 @@
...
+        weightofKnapsack = self.capacityOfKnapsack
+        rentRate = self.rentingRatio
+        vmin = self.minSpeed
+        vmax = self.maxSpeed
+        solution.ftraw = 0
+
+        if tour[0] != tour[-1]:
+            print("ERROR: The last city must be the same as the first city")
+            solution.reset()
+            return float(-1)
```
jrapin, line 171: Is that supposed to be the worst loss? just to make sure we are aligned, nevergrad only minimizes, so a bad loss would be `math.inf` (or just a very large value)

requirements/bench1.txt
```diff
@@ -0,0 +1,27 @@
+requests>=2.21.0
```
jrapin, line 1: please delete this file

requirements/dev1.txt
```diff
@@ -0,0 +1,13 @@
+black==20.8b1
```
jrapin, line 1: please delete this file

requirements/main1.txt
```diff
@@ -0,0 +1,4 @@
+numpy>=1.15.0
```
jrapin, line 1: please delete this file

requirements/t.txt
```diff
@@ -0,0 +1,23 @@
+- certifi==2021.5.30=py38haa95532_0
```
jrapin, line 1: please delete this file

nevergrad/functions/ttp/core.py
```diff
@@ -0,0 +1,224 @@
...
+        self.ftraw = math.inf
+        self.ob = float("-inf")
+
+        self.wend = float("inf")
+        self.wendUsed = float("inf")
+        self.computationTime = math.inf
+
+
+class TTPInstance(base.ExperimentFunction):
+    def __init__(self, file: str = "./instances/a20_n95_01.ttp") -> None:
```
jrapin, line 40: depending on how you want to use this type of files, this may not be the best option as parameter.
You could either give them a number each, it's simpler to use but then you can't use external files directly.
Or you could provide just the name, and then if you want a new file you add it to the folder.

Currently, I think the default path can't work, because the working directory is not the directory which contains the file, and you should not expect it is. This is also what triggers the continuous integration error in the PR.

A first step to check this would actually be to create a unit test and make it work.
That could/should be something as simple as this: https://github.com/facebookresearch/nevergrad/blob/edac4eaa65db5e5e01dc0f1ebcc5134c47b78060/nevergrad/functions/powersystems/test_core.py#L12-L18

nevergrad/functions/ttp/instances/a10_n9_01.ttp
```diff
@@ -0,0 +1,30 @@
+PROBLEM NAME: 	a10-TTP
```
jrapin, line 1: `.ttp` files will not be packaged automatically,
to make sure they are, please add the extension here:
https://github.com/facebookresearch/nevergrad/blob/edac4eaa65db5e5e01dc0f1ebcc5134c47b78060/setup.py#L93

requirements/environment.yml
```diff
@@ -0,0 +1,158 @@
+absl-py==0.12.0
```
jrapin, line 1: please delete this file

jrapin (changes requested):
Hi,
Thank you for the contribution!
Here are a few comments to be addressed before we could merge ;)

nevergrad/functions/ttp/core.py
```diff
@@ -0,0 +1,224 @@
...
+
+            if line.startswith("RENTING RATIO"):
+                line = line[line.index(":") + 1 :]
+                line = line.strip()
+                self.rentingRatio = float(line)
+
+            if line.startswith("EDGE_WEIGHT_TYPE"):
+                line = line[line.index(":") + 1 :]
+                line = line.strip()
+                self.edgeWeightType = line
```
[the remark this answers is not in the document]
yvan-vin: I experimented with a dict but it got complicated with having to link to the names in the file to the correct variable names, will try to find an easy fix though

nevergrad/functions/ttp/core.py
yvan-vin: Couldn't get get this or using choice for the packing plan to immediately work

nevergrad/functions/ttp/core.py
teytaud: pytest nevergrad/functions/powersystems/test_core.py   will launch the test in that file.
You can réplication this file in the directory nevergrad/functions/ttt/test_core.py

PR #1265, opened 2021-10-05
propagate recasted errors

bottler:
## Types of changes

Recaster runs an optimization on a background thread. If it never puts a candidate in the ask queue because it dies, `ask` could wait forever. 

The current pymoo functionality in nevergrad was written before the latest pymoo version, 0.5.0, was released. Without this change, if have pymoo 0.5.0, the (currently skipped) tests which use pymoo hang because the background thread makes a bad import and dies. After this change they fail with a nice ImportError. 

## Checklist

### DIFF
--- a/nevergrad/optimization/recaster.py
+++ b/nevergrad/optimization/recaster.py
@@ -54,9 +54,12 @@ class _MessagingThread(threading.Thread):
         try:
             self.output = self._caller(self._fake_callable, *self._args, **self._kwargs)
         except StopOptimizerThread:  # gracefully stopping the thread
-            pass
+            self.messages_ask.put(ValueError("Optimization told to finish"))
         except Exception as e:  # pylint: disable=broad-except
+            self.messages_ask.put(e)
             self.error = e
+        else:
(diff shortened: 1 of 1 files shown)


nevergrad/optimization/recaster.py
```diff
@@ -54,9 +54,12 @@ def run(self) -> None:
         try:
             self.output = self._caller(self._fake_callable, *self._args, **self._kwargs)
         except StopOptimizerThread:  # gracefully stopping the thread
-            pass
+            self.messages_ask.put(ValueError("Optimization told to finish"))
         except Exception as e:  # pylint: disable=broad-except
+            self.messages_ask.put(e)
             self.error = e
+        else:
+            self.messages_ask.put(ValueError("Optimization finished"))
```
jrapin, line 62: can you explain this? I don't understand the aim here
Wouldnt that interfere with retries (aka starting the optimization again from a different starting point if it finished ahead of budget)?

bottler: But you're right, that in this case we really want to go into the is_alive=False branch rather than pass an error to the user. I'll use a special value.

PR #1269, opened 2021-10-12
Enable pymoo pickling

bottler:
## Types of changes

- [] Breaking change (fix or feature that would cause existing functionality to change)

We add the ability for sequential recast optimizers to be pickled. This works by storing the entire history, which is replayed to the background thread on unpickling.

The implementations of `_internal_ask_candidate` and `_internal_tell_candidate` in RecastOptimizer have been moved to  SequentialRecastOptimizer which is where they are used.

Also added a bit more documentation on how to implement recast optimizers, and demonstrate using a weak reference to avoid a slightly mind-bending copy inside them.

New test.

### DIFF
--- a/nevergrad/optimization/recastlib.py
+++ b/nevergrad/optimization/recastlib.py
@@ -42,40 +42,32 @@ class _ScipyMinimizeBase(recaster.SequentialRecastOptimizer):
         """  # We do not do anything; this just updates the current best.
 
     def get_optimization_function(self) -> tp.Callable[[tp.Callable[[tp.ArrayLike], float]], tp.ArrayLike]:
-        # create a different sub-instance, so that the current instance is not referenced by the thread
-        # (consequence: do not create a thread at initialization, or we get a thread explosion)
-        subinstance = self.__class__(
-            parametrization=self.parametrization,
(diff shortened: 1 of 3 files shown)


nevergrad/optimization/recaster.py
```diff
@@ -222,12 +205,108 @@ def __del__(self) -> None:
...
+    def __init__(
+        self,
+        parametrization: IntOrParameter,
+        budget: tp.Optional[int],
+        num_workers: int = 1,
+        *,
+        enable_pickling: bool = False,
+    ) -> None:
+        super().__init__(parametrization=parametrization, budget=budget, num_workers=num_workers)
+        self._enable_pickling = enable_pickling
```
jrapin, line 251: how about we leave it as a public attribute? So that people can modify it easily on any instance. I'm not sure whether we should default to False (and then fail pickling) or True (and then risk using too much memory) :s

jrapin (approved):

nevergrad/optimization/recaster.py
bottler: I've changed it to a public function `enable_picking`.

PR #1315, opened 2021-12-02
Add architecture search for Gym

teytaud:
## Types of changes

## Checklist

### DIFF
--- a/nevergrad/functions/gym/multigym.py
+++ b/nevergrad/functions/gym/multigym.py
@@ -567,7 +567,6 @@ class GymMulti(ExperimentFunction):
             parametrization1 = parameter.Array(shape=shape)
             repetitions = int(np.prod(shape))
             assert isinstance(repetitions, int), f"{repetitions}"
-            #parametrization2 = parameter.Array(shape=shape)  # 0/1 weights for enabling/disabling
             parametrization2 = ng.p.Choice([0, 1], repetitions=repetitions)  # type: ignore
             parametrization = ng.p.Instrumentation(  # type: ignore
                 weights=parametrization1,
--- a/nevergrad/functions/gym/multigym.py
+++ b/nevergrad/functions/gym/multigym.py
@@ -595,7 +595,10 @@ class GymMulti(ExperimentFunction):
             parametrization.set_name("conformant")
 
         # Now initializing.
-        super().__init__(self.sparse_gym_multi_function if sparse_limit is not None else self.gym_multi_function, parametrization=parametrization)
+        super().__init__(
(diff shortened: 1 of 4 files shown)


nevergrad/functions/gym/multigym.py
```diff
@@ -557,6 +561,18 @@ def __init__(
...
+        if sparse_limit is not None:
+            parametrization1 = parameter.Array(shape=shape)
+            parametrization2 = parameter.Array(shape=shape)  # 0/1 weights for enabling/disabling
+            parametrization2.set_integer_casting()
+            parametrization2.set_bounds(0, 1)
+            parametrization = ng.p.Dict(  # type: ignore
+                weights=parametrization1,
+                enablers=parametrization2,
+            )
+            parametrization.set_name("ng_sparse" + str(sparse_limit))
```
jrapin, line 573: not sure if the name is useful in the end, since sparse_limit will already appears in the report anyway. shouldnt be harmful either, so up to you

nevergrad/functions/gym/multigym.py
```diff
@@ -557,6 +561,18 @@ def __init__(
 
         # Create the parametrization.
         parametrization = parameter.Array(shape=shape).set_name("ng_default")
+        if sparse_limit is not None:
+            parametrization1 = parameter.Array(shape=shape)
+            parametrization2 = parameter.Array(shape=shape)  # 0/1 weights for enabling/disabling
+            parametrization2.set_integer_casting()
+            parametrization2.set_bounds(0, 1)
```
jrapin, line 568: I'd be cautious here:
- setting bounds afterwards is not a very good idea: it uses the default sigma which is 1, while if you pass lower and upper in the init of Array it will use something better scaled and set the middle as initialization
- if everything is discrete, why not use ng.p.Choice([0, 1], repetitions=whatever)? this seems much better

teytaud: Ok. We lose the shape though.

teytaud: Hum there is a problem of shape. "repetitions" is necessarily an int.

nevergrad/functions/gym/multigym.py
teytaud: Ok it was working with the good old Arrays, but I fail to make it work with Instrumentation and Choice or Tuple and Choice.

nevergrad/functions/gym/multigym.py
teytaud: Ok fixed. The fact that the shape is lost is a bit an issue though.

PR #1316, opened 2021-12-03
Add tests for the suggest method

teytaud:
## Types of changes

We want to be sure that "suggest" works.
It turns out that it does work.

## Checklist


nevergrad/optimization/test_optimizerlib.py
```diff
@@ -197,6 +197,86 @@ def test_infnan(name: str) -> None:
...
+        return False
+    if "EMNA" in name or "EDA" in name:
+        return False
+    if "Stupid" in name:
+        return False
+    if "BO" in name:
+        return False
+    if "Pymoo" in name:
+        return False
+    return True
```
jrapin, line 212:
```suggestion
    keywords = ["TBPSA", "BO", "EMNA", "EDA", "BO", "Stupid", "Pymoo"]
    return not any(x in name for x in keywords)
```

I still don't like relying on names though, would be better to use the base classes, or rely on TellNotAskedNotSupported

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -197,6 +197,86 @@ def test_infnan(name: str) -> None:
...
+
+    optimizer_cls = registry[name]
+    instrum = ng.p.Array(shape=(100,)).set_bounds(0.0, 1.0)
+    instrum.set_integer_casting()
+    xs = np.asarray([0] * 17 + [1] * 17 + [0] * 66)
+    optim = optimizer_cls(instrum, budget=7)
+    target = lambda x: 0 if np.all(np.asarray(x, dtype=int) == xs) else 1
+    optim.suggest(xs)
+    optim.minimize(target)
+    assert np.all(optim.recommend().value == xs), "{name} proposes {optim.recommend().value} instead of {xs}"
```
jrapin, line 228:
```suggestion
    assert not target(optim.recommend().value), "{name} proposes {optim.recommend().value} instead of {xs}"
```

jrapin (changes requested):
There is a lot of **very** redundant code, you could easily change the target depending on the case instead

nevergrad/optimization/test_optimizerlib.py
teytaud: I don't get it, your version does not have xs anywhere in the equation ?

teytaud:
> There is a lot of **very** redundant code, you could easily change the target depending on the case instead

The different tests are distinct though. Some cases are continuous and others are discrete. Also, in some cases we just check "if the optimum has been suggested then we find it" and in other cases we check the stronger requirement "if a good suboptimal point has been provided then we find the optimum".

nevergrad/optimization/test_optimizerlib.py
teytaud: Oh I see, you want to use the fact that the optimum has value 0.

teytaud:
> > There is a lot of **very** redundant code, you could easily change the target depending on the case instead
> 
> The different tests are distinct though. Some cases are continuous and others are discrete. Also, in some cases we just check "if the optimum has been suggested then we find it" and in other cases we check the stronger requirement "if a good suboptimal point has been provided then we find the optimum".

I think compacting more would decrease readability.

jrapin:
> > > There is a lot of **very** redundant code, you could easily change the target depending on the case instead
> > 
> > 
> > The different tests are distinct though. Some cases are continuous and others are discrete. Also, in some cases we just check "if the optimum has been suggested then we find it" and in other cases we check the stronger requirement "if a good suboptimal point has been provided then we find the optimum".
> 
> I think compacting more would decrease readability.

But increase cost of maintenance, there is too much shared code here, I am sure it's possible to be both readable and factorized

teytaud:
> > > > There is a lot of **very** redundant code, you could easily change the target depending on the case instead
> > > 
> > > 
> > > The different tests are distinct though. Some cases are continuous and others are discrete. Also, in some cases we just check "if the optimum has been suggested then we find it" and in other cases we check the stronger requirement "if a good suboptimal point has been provided then we find the optimum".
> > 
> > 
> > I think compacting more would decrease readability.
> 
> But increase cost of maintenance, there is too much shared code here, I am sure it's possible to be both readable and factorized

ok for you now ?

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -196,6 +196,90 @@ def test_infnan(name: str) -> None:
...
 
+def suggestable(name: str) -> bool:
+    # Some methods are not good with suggestions.
+    keywords = ["TBPSA", "BO", "EMNA", "EDA", "BO", "Stupid", "Pymoo"]
+    return not any(x in name for x in keywords)
+
+
+def suggestion_testing(
+    name: str,
+    instrumentation: tp.Union[ng.p.Array, ng.p.Instrumentation],
```
jrapin, line 207:
```suggestion
    instrumentation: np.p.Array,
```
you only use Array, and the tests will fail with anything else anyway

nevergrad/optimization/test_optimizerlib.py
```diff
@@ -196,6 +196,90 @@ def test_infnan(name: str) -> None:
...
+    instrum = ng.p.Array(shape=(100,)).set_bounds(0.0, 1.0)
+    instrum.set_integer_casting()
+    optimum = np.asarray([0] * 17 + [1] * 17 + [0] * 66)
+    target = lambda x: min(3, np.sum((np.asarray(x, dtype=int) - optimum) ** 2))
+    suggestion = np.asarray([0] * 17 + [1] * 16 + [0] * 67)
+    suggestion_testing(name, instrum, suggestion, 1500, target, optimum)
+
+
+def good_at_c0_suggest(r: str) -> bool:
+    return "NGOpt" == r  # Let us check on NGOpt only for now.
```
jrapin, line 269: is that really needed as a function?

jrapin (commented):
given it's added 3 very specific tests on one particular topic, it may be better to move to a new file test_suggest.py

jrapin (commented):
the code seems very redundant, but not sure it could be restructured better

nevergrad/optimization/test_optimizerlib.py
teytaud: Removed.

teytaud:
> the code seems very redundant, but not sure it could be restructured better

is it ok for you now ? i refactored cycling.py but not the underlying computations: i assume that original authors would be lost if I changed the deep stuff.

jrapin (approved):
I would still move all suggestions tests to one test_suggest.py file because we start to have quite a few of them, and test_optimizerlib is too big

PR #1323, opened 2021-12-17
The metamodel never uses more than 50% of the total time

teytaud:
## Types of changes

## Checklist

### DIFF
--- a/nevergrad/optimization/optimizerlib.py
+++ b/nevergrad/optimization/optimizerlib.py
@@ -2434,8 +2434,8 @@ class NGOptBase(base.Optimizer):
     def optim(self) -> base.Optimizer:
         if self._optim is None:
             self._optim = self._select_optimizer_cls()(self.parametrization, self.budget, self.num_workers)
-            optim = self._optim if not isinstance(self._optim, NGOptBase) else self._optim.optim
-            logger.debug("%s selected %s optimizer.", *(x.name for x in (self, optim)))
+            self._optim = self._optim if not isinstance(self._optim, NGOptBase) else self._optim.optim
+            logger.debug("%s selected %s optimizer.", *(x.name for x in (self, self._optim)))
         return self._optim
 
     def _select_optimizer_cls(self) -> base.OptCls:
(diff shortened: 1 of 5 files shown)


teytaud:
Can not work because not picklable.
We need something like that though.

teytaud:
> Can not work because not picklable. We need something like that though.

Ok fixed.

jrapin (approved):
I don't quite like relying on multiprocessing but if you are confident with this...

nevergrad/optimization/optimizerlib.py
```diff
@@ -1449,7 +1450,12 @@ class MetaModelFailure(ValueError):
     """Sometimes the optimum of the metamodel is at infinity."""
 
 
-def _learn_on_k_best(archive: utils.Archive[utils.MultiValue], k: int) -> tp.ArrayLike:
+def train_sklearn(model, X2, y, return_dict) -> None:
```
bottler, line 1453:
```suggestion
def _train_sklearn(model, X2, y, return_dict) -> None:
```

teytaud:
> I don't quite like relying on multiprocessing but if you are confident with this...

I would prefer another solution --- if there was one (?)...

nevergrad/optimization/optimizerlib.py

jrapin:
Have you tried the performance ? I'm a bit afraid of the impact it can have as well. Eg: how many time does the metamodel actually finishes? I guess it depends on the application

nevergrad/optimization/optimizerlib.py
```diff
@@ -1449,7 +1451,12 @@ class MetaModelFailure(ValueError):
     """Sometimes the optimum of the metamodel is at infinity."""
 
 
-def _learn_on_k_best(archive: utils.Archive[utils.MultiValue], k: int) -> tp.ArrayLike:
+def _train_sklearn(model, X2, y, return_dict) -> None:
+    model.fit(X2, y)
+    return_dict["model"] = model
+
+
+def _learn_on_k_best(archive: utils.Archive[utils.MultiValue], k: int, timeout: int) -> tp.ArrayLike:
```
jrapin, line 1459: can we move both this functions to a different file? They are begginning to be big, and messy, in a file that is already big, and they should not serve as a model for later development...

jrapin: eg : optimization/utils.py

teytaud:
Do i correctly remember that the use of multiprocesding here would be a problem and that therefore i should close ?

PR #1336, opened 2022-01-19
The user can modify NG_CPUS_PER_TASK

teytaud:
The environment variable NG_CPUS_PER_TASK modifies the number of CPUs to be used per task.

## Types of changes

Possibility to parallelize inside each run in benchmarks.

## Checklist

### DIFF
--- a/nevergrad/benchmark/xpbase.py
+++ b/nevergrad/benchmark/xpbase.py
@@ -7,10 +7,12 @@ import sys
 import time
 import random
 import numbers
+import os
 import warnings
 import traceback
 import typing as tp
 import numpy as np
+from concurrent import futures
 from nevergrad.parametrization import parameter as p
 from nevergrad.common import decorators
 from nevergrad.common import errors
@@ -57,7 +59,11 @@ class OptimizerSettings:
         self.optimizer = optimizer
         self.budget = budget
         self.num_workers = num_workers
-        self.executor = execution.MockedTimedExecutor(batch_mode)
+        ng_cpus_per_task = int(os.getenv("NG_CPUS_PER_TASK", "1"))
+        if ng_cpus_per_task > 0:
+            self.executor = futures.ThreadPoolExecutor(max_workers=ng_cpus_per_task)
+        else:
+            self.executor = execution.MockedTimedExecutor(batch_mode)
 
     @property
     def name(self) -> str:


nevergrad/benchmark/xpbase.py
```diff
@@ -57,7 +59,11 @@ def __init__(
         self.optimizer = optimizer
         self.budget = budget
         self.num_workers = num_workers
-        self.executor = execution.MockedTimedExecutor(batch_mode)
+        ng_cpus_per_task = int(os.getenv("NG_CPUS_PER_TASK", "1"))
+        if ng_cpus_per_task > 0:
```
jrapin, line 63:
```suggestion
        if ng_cpus_per_task > 1:
```

nevergrad/benchmark/xpbase.py
```diff
@@ -57,15 +59,20 @@ def __init__(
         self.optimizer = optimizer
         self.budget = budget
         self.num_workers = num_workers
-        self.executor = execution.MockedTimedExecutor(batch_mode)
+        ng_cpus_per_task = int(os.getenv("NG_CPUS_PER_TASK", "1"))
```
jrapin, line 62: This is only for benchmark so let's make it clear
```suggestion
        ng_cpus_per_task = int(os.getenv("NG_BENCHMARK_CPUS_PER_TASK", "1"))
```

nevergrad/benchmark/xpbase.py
```diff
@@ -57,15 +59,20 @@ def __init__(
...
+        if ng_cpus_per_task > 1:
+            self.executor: tp.Any = futures.ThreadPoolExecutor(max_workers=ng_cpus_per_task)
```
jrapin, line 64: I'm still very much against this since it
- prevents reproducibility
- adds complexity which may create a lot of issues (eg: not sure that it will checkpoint correctly if preempted)
- assumes the benchmarks are threadsafe, which is not always true on our side and possibly on the dependency side

Can you quantify the speed gain?
Python executes one thread at a time for most operations, so it may just waste time for creating threads without going faster (depending on what is happening in the underlying code)
Moving to a ProcessPoolExecutor could help for parallelization, but I would expect a big overhead so that it could be unhelpful most of the time

nevergrad/benchmark/xpbase.py
```diff
@@ -57,15 +59,20 @@ def __init__(
...
+            self.executor = execution.MockedTimedExecutor(batch_mode)
 
     @property
     def name(self) -> str:
         return self.optimizer if isinstance(self.optimizer, str) else repr(self.optimizer)
 
     @property
     def batch_mode(self) -> bool:
-        return self.executor.batch_mode
+        return hasattr(self.executor, "batch_mode") and self.executor.batch_mode
```
jrapin, line 75:
```suggestion
        return isinstance(self.executor, execution.MockedTimedExecutor) and self.executor.batch_mode
```

nevergrad/benchmark/xpbase.py
```diff
@@ -207,7 +214,9 @@ def run(self) -> tp.Dict[str, tp.Any]:
     def _log_results(self, pfunc: fbase.ExperimentFunction, t0: float, num_calls: int) -> None:
         """Internal method for logging results before handling the error"""
         self.result["elapsed_time"] = time.time() - t0
-        self.result["pseudotime"] = self.optimsettings.executor.time
+        self.result["pseudotime"] = (
+            self.optimsettings.executor.time if hasattr(self.optimsettings.executor, "time") else float("nan")
```
jrapin, line 218:
```suggestion
            self.optimsettings.executor.time if  isinstance(self.executor, execution.MockedTimedExecutor) else float("nan")
```

nevergrad/benchmark/xpbase.py
jrapin: checking hasattr is not safe, if we update the code that could break without notice, it's much safer to test against a class

PR #1352, opened 2022-01-26
adding installation instruction for MuJoCo

mathuvu:
## Types of changes

Since MuJoCo is now open source, this test if it is installed and adds installation instructions.

## Checklist

### DIFF
--- a/nevergrad/benchmark/experiments.py
+++ b/nevergrad/benchmark/experiments.py
@@ -1202,10 +1202,21 @@ def mixsimulator(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
                             yield xp
 
 
+def check_mujoco():
+    try:
+        import mujoco_py
+    except:
+        print(
+            "MuJoCo not installed (Linux/OSX support only). If you need it, please follow this installation guide:"
+        )
+        print("https://github.com/openai/mujoco-py#install-mujoco")
+
+
 @registry.register
 def control_problem(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
     """MuJoCo testbed. Learn linear policy for different control problems.
     Budget 500, 1000, 3000, 5000."""
+    check_mujoco()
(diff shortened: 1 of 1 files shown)


nevergrad/benchmark/experiments.py
```diff
@@ -1202,10 +1202,21 @@ def mixsimulator(seed: tp.Optional[int] = None) -> tp.Iterator[Experiment]:
...
 
 
+def check_mujoco():
+    try:
+        import mujoco_py
+    except:
+        print(
+            "MuJoCo not installed (Linux/OSX support only). If you need it, please follow this installation guide:"
+        )
+        print("https://github.com/openai/mujoco-py#install-mujoco")
```
jrapin, line 1212: shouldnt that be raised as an error since these 2 experiments do need it?

mathuvu: Yes, you're right. I'll change that.

jrapin: Better, but I was a bit fast in my comment sorry.
As we can see the tests won't run because of the error, so actually one would need to raise a ng.errors.UnsupportedExperiment which is an error which is skipped in the tests.

Skipping the tests is however always dangereous, because that means we don't make sure the following code works, so it's better to move the mujoco check as deep in the code as possible, which in this case means `control.core.BaseFunction` where we do find the `UnsupportedExperiment` error:
https://github.com/facebookresearch/nevergrad/blob/43d79b51ebc9f3fb091dff1191f7c8ce19a01de5/nevergrad/functions/control/core.py#L113

So can we just update this error with the clearer instructions that you have written? That would be less code, better tested while being clearer for users, would that work?

mathuvu: Yes, of course. I wasn't aware of that. It's much better this way.

jrapin (approved):

jrapin:
Thanks, feel free to "Squash & Merge" ;)

PR #1366, opened 2022-02-28
High-speed differential evolution

teytaud:
## Types of changes

Special recommendation for DE as well.

## Checklist


nevergrad/optimization/differentialevolution.py
```diff
@@ -7,6 +7,8 @@
 import numpy as np
 import nevergrad.common.typing as tp
 from nevergrad.parametrization import parameter as p
+from .metamodel import MetaModelFailure as MetaModelFailure
+from .metamodel import _learn_on_k_best as _learn_on_k_best
```
jrapin, line 11:
```suggestion
from . import metamodel
```
Then just use metamodel._learn_k_best etc...
Then again, that means it should probably be removed to remove the initial _

the syntax import X as X is only necessary for external interfaces, so not here

nevergrad/optimization/differentialevolution.py
```diff
@@ -91,6 +93,7 @@ def __init__(
         super().__init__(parametrization, budget=budget, num_workers=num_workers)
         # config
         self._config = DifferentialEvolution() if config is None else config
+        self.high_speed = self._config.high_speed
```
jrapin, line 96:
```suggestion
```
just use self._config.high_speed... which is already what you do

nevergrad/optimization/differentialevolution.py
```diff
@@ -313,12 +325,15 @@ def __init__(
 
 
 DE = DifferentialEvolution().set_name("DE", register=True)
+HSDE = DifferentialEvolution(high_speed=True).set_name("HSDE", register=True)
```
jrapin, line 328: as usual, is it useful as non-experimental variant? we should avoid cluttering the namespace

nevergrad/optimization/metamodel.py
```diff
@@ -0,0 +1,80 @@
...
+from . import utils
+from .base import registry
+from . import callbacks
+
+
+class MetaModelFailure(ValueError):
+    """Sometimes the optimum of the metamodel is at infinity."""
+
+
+def _learn_on_k_best(archive: utils.Archive[utils.MultiValue], k: int) -> tp.ArrayLike:
```
jrapin, line 17: make it non-private since we use it in other files
```suggestion
def learn_on_k_best(archive: utils.Archive[utils.MultiValue], k: int) -> tp.ArrayLike:
```

nevergrad/optimization/optimizerlib.py
```diff
@@ -17,11 +17,11 @@
 from nevergrad.parametrization import discretization
 from nevergrad.parametrization import _layering
 from nevergrad.parametrization import _datalayers
-from . import callbacks
 from . import oneshot
 from . import base
 from . import mutations
-from . import utils
+from .metamodel import MetaModelFailure as MetaModelFailure
+from .metamodel import _learn_on_k_best as _learn_on_k_best
```
jrapin, line 24:
```suggestion
from . import metamodel
```
then use metamodel.MetaModelFailure etc...
That makes it clearer where things come from.

jrapin (changes requested):

nevergrad/optimization/differentialevolution.py
```diff
@@ -286,6 +296,7 @@ def __init__(
         popsize: tp.Union[str, int] = "standard",
         propagate_heritage: bool = False,  # experimental
         multiobjective_adaptation: bool = True,
+        high_speed: bool = False,
```
jrapin, line 299: please update the docstring + I am not sure "highspeed" is the most descriptive name

teytaud: Done. High-Speed is my best guess for now.
For example FastGA became a famous name whereas it does not say much about what it's doing :-)

nevergrad/optimization/metamodel.py
teytaud: Fixed.

nevergrad/optimization/differentialevolution.py
teytaud: I would rather keep it here if positive results are confirmed a little bit more.
Still I agree that a big cleaning of optimizerlib.py is necessary; based on splitting (a lot) and removing (a bit).

nevergrad/optimization/optimizerlib.py

nevergrad/optimization/differentialevolution.py
```diff
@@ -114,6 +115,13 @@ def __init__(
         self._no_hypervolume = self._config.multiobjective_adaptation
 
     def recommend(self) -> p.Parameter:  # This is NOT the naive version. We deal with noise.
+        sample_size = int((self.dimension * (self.dimension - 1)) / 2 + 2 * self.dimension + 1)
+        if self._config.high_speed and len(self.archive) >= sample_size:
```
jrapin, line 119:
```suggestion
        if self._config.high_speed and len(self.archive) >= sample_size:
            sample_size = int((self.dimension * (self.dimension - 1)) / 2 + 2 * self.dimension + 1)
```

nevergrad/optimization/differentialevolution.py
jrapin: > if positive results are confirmed a little bit mor

My recommendation would be, as long as it's not confirmed, put it into experimentalvariants, and as soon as it is, move it here. Here should not be the defaults, the default should be experimentalvariants because removing stuff from experimentalvariants is not a breaking change

nevergrad/optimization/differentialevolution.py
teytaud: I don't get it --- how can we compare to sample_size before defining sample_size ?

teytaud:
Regarding the experimental variant vs optimizerlib, I'll wait to have more results before merging, so that we don't have to move stuff. Optimizerlib if good results, experimentalvariants otherwise.

jrapin:
>  I'll wait to have more results before merging, so that we don't have to move stuff

It's not a problem **at all** to move stuff from experimentalvariant to optimizerlib. It's only a problem to move stuff out of optimizerlib. So I would recommend merging to experimentalvariants initially, then move in another PRs if you have good results. This way there is no need to keep PRs open waiting for results

nevergrad/optimization/differentialevolution.py
jrapin: my bad, I did not see it was used in the if condition

jrapin (approved):
preferably move to experimentalvariants, merge, and if xps work as expected, move back from experimentalvariants to differentialevolution.py

PR #1410, opened 2022-04-01
randomly skip experiments

teytaud:
## Types of changes

There are complaints against arrays of small jobs in scavenge queues, which are
poorly handled on some clusters and get, for unclear reason, super high priority.

While this is due to the cluster, not to Nevergrad, I propose this simple workaround.

The following bash script:
#!/bin/bash
sbatch --array=0-79%7 --partition=learnlab --time=72:00:00  --job-name=$1 --gpus-per-task=0 --cpus-per-task=67 --wrap="python -m neverg
rad.benchmark $1 --num_workers=67 --plot --random_skip 0.1"

with argument $1 will run the experiment $1 with 90% experiment killed.
That way:
- we launch jobs with 67 cores, not jobs with 1 core.
- the walltime 72 hours is not a problem, because you run only 10% of experiments.
- given the 79 jobs, you will get random replicas and everything should be covered. Anyway you can see what is going on in fight_all.png / fight_all_pure.png

Not much yet, I'm using a different hack for now, I'll check this one from now on.

## Checklist

### DIFF
--- a/nevergrad/benchmark/__main__.py
+++ b/nevergrad/benchmark/__main__.py
@@ -77,6 +77,12 @@ def get_args() -> argparse.Namespace:
         help="Numbers of workers to use for the computation (splits the job in chunks)",
     )
     parser.add_argument(
+        "--random_skip",
+        type=float,
+        default=0.,
+        help="frequency of randomly skipping one of the experiment (for getting fast partial results)",
+    )
+    parser.add_argument(
         "--repetitions",
         type=int,
         default=1,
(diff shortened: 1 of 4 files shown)


jrapin (changes requested):
I expect this will get even more complaints, using 1 cpu jobs has a much lower impact on the cluster

teytaud:
> I expect this will get even more complaints, using 1 cpu jobs has a much lower impact on the cluster

I agree that this should be true if the cluster is ok...

PR #1435, opened 2022-04-22
update ProcessPoolExecutor docs

mathuvu:
## Types of changes

ThreadPoolExecutor use only one thread for all task in async due to GIL lock (one interpreter = one thread). For CPU-bound function ProcessPoolExecutor circumvent this issue.

```
import nevergrad as ng
from concurrent import futures
import time
import multiprocessing as mp

from nevergrad.optimization import optimizerlib

def func(x):
    tmp = 1
    for k in range(4_000_000):#Garbage calculation
        tmp += 1
        tmp -= 1
    return sum((x-1)**2)*tmp
 parametrization = ng.p.Array(shape=(10,))

start = time.time()
budget = 100

optim = optimizerlib.OnePlusOne(parametrization=parametrization, budget=budget, num_workers=10)
# recom = optim.minimize(func, executor=futures.ThreadPoolExecutor(max_workers=optim.num_workers))
recom = optim.minimize(func, executor=futures.ProcessPoolExecutor(max_workers=optim.num_workers, mp_context=mp.get_context('fork')))# Spawn mode raise an error, so it will not work on windows.

end = time.time()
print(end - start)
```

## Checklist

### DIFF
--- a/nevergrad/benchmark/core.py
+++ b/nevergrad/benchmark/core.py
@@ -234,7 +234,7 @@ def _submit_jobs(
     seed: int
         a seed for the experiment plan (if seedable)
     executor: Executor-like object
-        an object such as concurrent.futures.ThreadPoolExecutor for running experiments in parallel
+        an object such as concurrent.futures.ProcessPoolExecutor for running experiments in parallel
     print_function: tp.Callable
         a function to print at the end of each experiment (for custom logging)
     cap_index: int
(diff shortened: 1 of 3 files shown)


teytaud:
I think this PR is super important but complicated for CircleCI because we are not allowed to do multiprocessing....
But we need to change this because of https://stackoverflow.com/questions/58919495/why-is-threadpoolexecutor-slower-than-for-loop)

Issue #1461, opened 2022-07-11
Invalid state for DE unless num_ask - num_tell < popsize


Game4Move78:
Calling `ask` more than `popsize` times will result in an invalid state for DE. In particular it will cause the _uid_queue to wrap around, and produce two asked points with the same parent. Then a tell-not-ask with better loss has the potential to kick the parent from the population, but leave its uid in the _uid_queue, resulting in a KeyError on one of the subsequent calls to `ask`.

If this is for performance reasons, the documentation could stress that calling `ask` no more than `popsize` times should be manually enforced as it can break some optimizers.

bottler:
When the user creates an optimizer object and specifies `num_workers`, they are promising that they will never have more than the given number of untold asks. Nevergrad optimizers are not designed to handle this promise being broken.

Game4Move78:
Perhaps this described in the docs as a promise rather than as typical usage. 

https://github.com/facebookresearch/nevergrad/issues/263 is an example where a user other than myself thought they could ignore impossible-to-evaluate points without a `tell`.

The answer to https://github.com/facebookresearch/nevergrad/issues/1028 had given me the impression that it's not absolutely crucial to set popsize larger than `num_workers`, and that its just a good rule-of-thumb.

Game4Move78:
[This might be a safe way to do it](https://github.com/facebookresearch/nevergrad/compare/main...Game4Move78:nevergrad:patch-1). I don't know if this invariant matters for other optimizers.

PR #1462, opened 2022-07-11
Error checking for  #1461

Game4Move78:
https://github.com/facebookresearch/nevergrad/issues/1461

## Types of changes

https://github.com/facebookresearch/nevergrad/issues/1461

## Checklist

### DIFF
--- a/nevergrad/optimization/differentialevolution.py
+++ b/nevergrad/optimization/differentialevolution.py
@@ -198,7 +198,7 @@ class _DE(base.Optimizer):
         if uid not in self.population:  # parent was removed, revert to tell_not_asked
             self._internal_tell_not_asked(candidate, loss)
             return
-        if uid in self._uid_queue.asked: # if taken from queue in multiple asks, add back only once
+        if uid in self._uid_queue.asked:  # if taken from queue in multiple asks, add back only once
             self._uid_queue.asked.discard(uid)
             self._uid_queue.tell(uid)
         parent = self.population[uid]


teytaud:
A this line: https://github.com/Game4Move78/nevergrad/blob/7152ca66cc5f66c0427579d486075ba18dc003b7/nevergrad/optimization/differentialevolution.py#L104

We might add 
 self.llambda = max(self.llambda, self.num_workers)

This should solve the issue, however it means that the user might specify "I want llambda to be 20 and Nevergrad decides to set llambda to 30".

teytaud:
Your code looks good to me, the problem might be in MixDeterministicRL. I investigate. Thanks for your work.

nevergrad/optimization/differentialevolution.py
```diff
@@ -158,6 +159,8 @@ def _internal_ask_candidate(self) -> p.Parameter:
             self.population[candidate.uid] = candidate
             self._uid_queue.asked.add(candidate.uid)
             return candidate
+        # stop queue wrapping around to lineage waiting for a tell
+        assert self._uid_queue.told, "More untold asks than population size (exceeds num_workers)"
```
teytaud, line 163: @jrapin  you are the expert for self._uid_queue.told (among so many things...), do you validate this assert ?

teytaud: Oh I guess the error is in class Portfolio. Let me propose a fix (fingers crossed :-) ).

Game4Move78: > @jrapin you are the expert for self._uid_queue.told (among so many things...), do you validate this assert ?

If it helps, my thinking was that there should be a tell preceding every ask after the initalization phase keeping the told queue non-empty. Even in the worst case where popsize ==num_workers and all workers are evaluating untold points, the worker that beats the others to the tell can use the same point again on the next ask.

teytaud: I've just sent a message to Jeremy, who knows that code better than anyone else and who might not have been close to github recently. Sorry for the delay; your PR is interesting.

jrapin: I used to be strict with the fact that we should not go beyond `num_workers`, but I changed my mind a couple of years ago because there are many cases you don't master all the details of what is happening (eg: a process dies and you'll never get the result), most times the user won't deal with it and we should be robust to it to simplify use. The code was then  **supposed** to be robust but visibly there are corner cases :s 
I would be therefore rather make it robust to this case (would that just take removing duplicates in UuidQueue.told ? it should be light speed so not a problem)

cc @bottler you seemed to disagree and want the user to strictly conform to the "contract", maybe we can discuss and adapt depending if I change your mind or not ;)

@Game4Move78 as a power user, would you rather it bugged explicitely, or be robust to those corner cases? (why did you happen to ask for more points?)

Game4Move78: I followed the hyper-parameter settings of papers that used DE for HPO and set popsize to 20 explicitly without providing num_workers, and thought it would be robust. I then asked for more points and handed them to my own adaptive resource
allocation + early stopping implementation that evaluated HPO choices multiple budgets and only provided a tell to the NG optimiser when points were either stopped early or allocated maximum budget.

This would work fine for hundreds of points until it hit that corner case with a point in the told queue that has been deleted from population. My current workaround is to provide feedback immediately on the minimum budget and then treats all evaluations on higher budgets as untold points, which works fine for DE.

If you want less strict, how about we allow duplicates in tell but at [L162](https://github.com/facebookresearch/nevergrad/blob/main/nevergrad/optimization/differentialevolution.py#L162) we add
```
while lineage not in self.population:
      lineage = self._uid_queue.ask()
```

Which I believe would toss away those deleted points.

Game4Move78: My personal preference to help users to master those details is to [copy Ax's client interface](https://github.com/facebook/Ax/blob/3a8f8dfd8061a78f71ec7ba697c268b4a5196520/ax/service/ax_client.py#L577) with an `abandon_tell`. For most optimisers this would just provide the optimiser with a large value, and the BO optimisers might do something different to avoid damaging the model.

Game4Move78:
@jrapin Any chance of getting this merged :smiley:? Line https://github.com/facebookresearch/nevergrad/blob/8403d6c9659f40fec2a3cf7f474b3d8610f0f2e4/nevergrad/optimization/utils.py#L340 also uses absence in `UidQueue.asked` to check for presence in `told`, and `self._uid_queue.asked` is configured directly on many lines in `_DE` already. 

I believe this code enforces that for asked points with the same parent, the lineage will be added to the told queue only once:
```
if uid in self._uid_queue.asked:  # if taken from queue in multiple asks, add back only once
            self._uid_queue.asked.discard(uid)
            self._uid_queue.tell(uid)
```

PR #1596, opened 2024-02-16
Update sphere.py -- add pooling and max_pooling

mzameshina:
## Types of changes

## Checklist


nevergrad/common/sphere.py
```diff
@@ -826,6 +855,8 @@ def metric_pack_big_conv(x, budget=default_budget):
     "Riesz_blursum_lowconv_loworder",
     "Riesz_blursum_lowconv_midorder",
     "Riesz_blursum_lowconv_highorder",
+    "max_pooling",
+    "pooling"
```
teytaud, line 859: All methods in the list have the same signature, so that they can be used by quasi-randomize.
Please use the same signature.

teytaud (commented):
Please use the same signature as other codes so that it can work in the quasi-randomize function.

Please add your code as default in quasi-randomize, when len(shape) > 1, instead of dispersion_with_big_conv.

teytaud:
Does it run ? This does not work as Torch is not in the dependencies.
I recommend not using torch, as it would imply a big dependency.

Also you need black.

PR #1673, opened 2025-02-12
Ceviche mode converter FDFD

Milloupe:
Created a loss_fn_ceviche file that contains the importable function computing the mode converter performance.
Added mode_converter.py and functions.py which are necessary imports.


teytaud:
buggy pr, fix in another branch

PR #1712, opened 2026-03-16
Fix loss_function_sm for recent numpy compatibility

bottler:
fixes #1707


teytaud:
> fixes #1707

Thanks a lot @bottler :-) I was precisely going to do a PR for this.

